PR #14, opened 2015-09-22
test/unit: adding micro_unit_shmem.c

kseager:
micro unit test addition for initial fabric interface development and debugging


sayantansur:
:+1:

PR #15, opened 2015-10-05
Ofi transport

kseager:
generalization of datatypes for multi-transport usage, and OFI transport insert

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -2,6 +2,10 @@ Copyright 2011 Sandia Corporation. Under the terms of Contract
 DE-AC04-94AL85000 with Sandia Corporation, the U.S.  Government
 retains certain rights in this software.
 
+Copyright (c) 2015 Intel Corporation. All rights reserved.
+This software is available to you under the BSD license. For
+license information, see the LICENSE file in the top level directory.
+
 
 ======================================================================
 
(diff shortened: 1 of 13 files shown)


jdinan:
What is the license of config/ompi_check_ofi.m4?  Is it compatible with the Sandia SHMEM licensing?

jdinan:
Sorry, should have also asked -- does this file come from Open MPI or some other source?  If you wrote it or derived it from existing Sandia SHMEM code, then please ignore my previous question.

regrant:
In shmem_comm.h why do you undef the DTYPE_* defines. These are done through a config script, so it seems like a cleaner way to do this would be to set the values you want at configure time instead of having to undef the portals-based definitions.

jdinan:
The USE_PORTALS4 and USE_OFI logic is probably fine, but it isn't consistent.  In some places the code checks whether the macros are defined and in others, it checks for truth.  Also, sometimes it's two if checks and others it's if followed by elif.  Also, it might be nice to define a transport interface and let the linker handle this rather than doing all of these preprocessor checks.

jdinan:
"ret" checks don't seem to be performed consistently.  For the simple ret checks, it might be helpful to have a macro like:

#define OFI_RET_CHECK(ret) do { if (ret) { RAISE_ERROR(ret); } } while (0)

You can use a configure flag to disable error checks and define this to an empty macro for high performance builds.

jdinan:
If you click on the filename in "jdinan commented on src/transport_ofi.h in 80599da an hour ago", it will take you to the line of code where I made the comment.  This was not obvious to me.  :)  Sorry for the number of comments.

sayantansur:
Kayla, can you use Signed-Off-By line for the contributions. git commit -s.

PR #30, opened 2015-10-27
NEW feature: shmem_pmi

kseager:
++built-in pmi support
Note on usage:
        build and install shmem_pmi before building shmem
        to use during shmem configuration add: --with-pmi=${PWD} --with-pmi-libdir=shmem_pmi
        ->the above will find your libshmem_pmi.so regardless of shmem_pmi's config prefix
	choosing shmem_pmi will assume mpiexec.hydra for make check (set env variables accordingly)
++if "with-ofi" isn't on configure line ofi will not be built

### DIFF
--- a/shmem_pmi/include/pmi.h
+++ b/shmem_pmi/include/pmi.h
@@ -1,0 +1,473 @@
+/* -*- Mode: C; c-basic-offset:4 ; indent-tabs-mode:nil ; -*- */
+/*
+ *  (C) 2001 by Argonne National Laboratory.
+ *      See COPYRIGHT in top-level directory.
+ */
+
+#ifndef PMI_H_INCLUDED
+#define PMI_H_INCLUDED
+
+#ifdef USE_PMI2_API
+#error This header file defines the PMI v1 API, but PMI2 was selected
+#endif
+
+/* prototypes for the PMI interface in MPICH */
+
+#if defined(__cplusplus)
+extern "C" {
+#endif
+
+/*D
+PMI_CONSTANTS - PMI definitions
+
+Error Codes:
++ PMI_SUCCESS - operation completed successfully
+. PMI_FAIL - operation failed
+. PMI_ERR_NOMEM - input buffer not large enough
+. PMI_ERR_INIT - PMI not initialized
+. PMI_ERR_INVALID_ARG - invalid argument
+. PMI_ERR_INVALID_KEY - invalid key argument
+. PMI_ERR_INVALID_KEY_LENGTH - invalid key length argument
+. PMI_ERR_INVALID_VAL - invalid val argument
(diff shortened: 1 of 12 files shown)


sayantansur:
One alternative we discussed was to simply add a dependency on MPICH that provides the PMI client, vs. ripping one out and putting it in SHMEM. If we copy/paste the PMI code here, we are also on the hook to keep it updated blah blah.

We should check the COPYRIGHT stuff to ensure MPICH copyright/license terms are obeyed.

kseager:
this insert is more so for ease of use. You can build once and use multi-PMI launchers. Currently I assume mpiexec.hydra in the build system for make check but this won't limit someone in their own test harness. Ideally I think it would be nice to have a full bodied PMI built-in option to ease portability and limit dependencies....

otherwise I can hack at the build system a little more to enable the user to specify which .so the PMI symbols are coming from (they aren't all called libpmi.so which is the current build assumption) and where this is installed. Otherwise it will assume you have one of the following: a libpmi.so in your default path,  slurm, or portals4 yod. 

We found shmem_pmi to be the easiest option to use in practice.

Ryan, give your experience, do you have a comment for which is the preferred PMI usage flavor?

sayantansur:
At the very least the COPYRIGHT stuff needs to be fixed. The PMI header files refer to top-level COPYRIGHT file, that doesn't exist. Also, the Argonne copyright is added just after a line that describes how the config macros are taken from Open MPI ;-) At the very least, we should create a separate section describing the simple PMI client and its origins.

shmem_pmi/include/pmi.h
```diff
@@ -0,0 +1,473 @@
...
+- PMI_FAIL - unable to return the kvsname
+
+Notes:
+This function returns the name of the keyval space that this process and all
+other processes in the process group have access to.  The output parameter,
+kvsname, must be at least as long as the value returned by
+'PMI_KVS_Get_name_length_max()'.
+
+@*/
+int PMI_KVS_Get_my_name( char kvsname[], int length );
```
regrant: Documentation for this function is wrong.

shmem_pmi/include/pmi.h
```diff
@@ -0,0 +1,473 @@
...
+Notes:
+This function puts the key/value pair in the specified keyval space.  The
+value is not visible to other processes until 'PMI_KVS_Commit()' is called.
+The function may complete locally.  After 'PMI_KVS_Commit()' is called, the
+value may be retrieved by calling 'PMI_KVS_Get()'.  All keys put to a keyval
+space must be unique to the keyval space.  You may not put more than once
+with the same key.
+
+@*/
+int PMI_KVS_Put( const char kvsname[], const char key[], const char value[]);
```
regrant: Documentation for this function is incorrect.

PR #38, opened 2015-11-06
Refactoring of transport code and general cleanups

jdinan:


sayantansur:
:100:

kseager:
bravo on this refactor; it really cleaned up the code base! 

One small comment comes to mind, for the CT test build option, maybe we should make the shmem_ct building option transport independent. For example, we could have a flag that transports set if they have implemented extensions.

PR #45, opened 2015-11-18
shmem_pmi implementation update: --enable-pmi-simple

kseager:
New use case model for using shmem_pmi
-thanks Jim for the idea of adding the --enable-pmi-simple feature
-reinvisioned build implementation (removed old implementation)
        -if you use configure option shmem_pmi will be built in by default
        else you must use the old usage pattern for pmi support
note: added autogen.sh for ease of use
This closes #23
Signed-off-by: kseager <kayla.seager@intel.com>

PR #48, opened 2015-11-19
add autogen script

jeffhammond:
How does one go through life without `autogen.sh`? :-)


jeffhammond:
rebase did some weird stuff so maybe just apply this change manually...

jdinan:
Thanks, but this isn't needed.  PR #45 already includes autogen.sh.

PR #53, opened 2015-11-23
Collectives update: new reduction algorithm for long double

kseager:
some providers can't support long double; as a result,
enable SHMEM-OFI to detect the limitation and change algorithms accordingly.
By default, long double support is assumed.
new file op.h: long double software atomic support

### DIFF
--- a/src/collectives_c.c
+++ b/src/collectives_c.c
@@ -392,9 +392,14 @@ shmem_longdouble_min_to_all(long double *target, long double *source, int nreduc
 {
     SHMEM_ERR_CHECK_INITIALIZED();
 
-    shmem_internal_op_to_all(target, source, nreduce, sizeof(long double),
+    if(!shmem_long_dub_supported) {
+	shmem_internal_op_to_all_recdbl(target, source, nreduce, sizeof(long double),
+			    PE_start, logPE_stride, PE_size, pWrk, pSync, SHM_INTERNAL_MIN, SHM_INTERNAL_LONG_DOUBLE);
+    } else {
+	shmem_internal_op_to_all(target, source, nreduce, sizeof(long double),
                     PE_start, logPE_stride, PE_size,
                     pWrk, pSync, SHM_INTERNAL_MIN, SHM_INTERNAL_LONG_DOUBLE);
+    }
 }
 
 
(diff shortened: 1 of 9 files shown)


jdinan:
IIUC, the intent of this PR is to extend support for the reduction operations to OFI providers that don't support one-sided atomics on long double.  I'm a little concerned about the approach, which seems to involve adding a new reduction algorithm that is used only for long double.  Can the existing reduction algorithms be extended to support this, by adding a software reduction mode?  Or can the OFI transport absorb this by supporting long double atomics in software?

kseager:
It would change the Mechanics of the existing algorithm to use software atomics in the other reduction algorithms. Currently, they rely on atomic_nb to do the operation which has the added benefit of not having to collect all the data points to do the operation on: you can assume each child "PE" will atomically perform the operation on the parents PE's current data. The recursive doubling method avoids a collect given the communication structure between the teams, however the other algorithms (linear/tree), given their current structure, adding in the software atomics would reduce performance (as far as I can tell) since it would require collecting all the data then doing the operation in software. I just don't feel like there is much to gain from extending the other algorithms. 

I think given the proposed options, putting long double into the transport layer would be cleaner.... You could keep the other algorithms as is, but you'd have to do some work on the recursive doubling routine. The recursive doubling scheme would have to be re-written to include atomic_nb, the software atomics would have to include extra communication to fake the atomic_nb operation, and a branch would have to be added in the atomic_nb pathway to detect the long double case.  

I really think for now we should keep this addition as is, at least for the first release, and revisit at a later point if it starts to become an issue. I just don't see a pressing enough need at this time.

kseager:
the bzero that showed up in the initial commit has since been removed

src/collectives_c.c
```diff
@@ -392,9 +392,14 @@ shmem_longdouble_min_to_all(long double *target, long double *source, int nreduc
 {
     SHMEM_ERR_CHECK_INITIALIZED();
 
-    shmem_internal_op_to_all(target, source, nreduce, sizeof(long double),
+    if(!shmem_long_dub_supported) {
```
jdinan: Push this conditional into shmem_internal_op_to_all instead of in the language bindings.  We anticipate needing a SW solution for float/double complex and potentially other type/op combinations.

src/transport_ofi.c
```diff
@@ -72,7 +72,7 @@ static int SHM_DT_INT[]=
 static int SHM_DT_CMP[]=
 {
   DTYPE_SHORT, DTYPE_INT, DTYPE_LONG,
```
jdinan: Is it safe to mix DTYPE_ and SHM_INTERNAL_?

src/collectives_c.c
kseager: in this case I'd have to undo some op.h work I've done. I've specialized everything for only the long double case. What are the combinations we are are anticipating? Should I just re-generalize for any atomic combination case?

jdinan: Don't worry for now.  I will generate a patch for you to review.

jdinan:
Superseded by #73.

PR #56, opened 2015-12-03
CI Updates, Bugfixes, and Cleanups

### LINKED ISSUE #28: Update deprecated API usage in tests
Many tests call deprecated API routines.  Update to use new API or mark tests as deprecated if they are testing deprecated functionality (e.g. cache API).

jdinan:
This will close #28.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,37 +1,110 @@
 sudo: true
 language: c
 compiler:
+    #- clang
     - gcc
 os:
     - linux
+
 before_install:
-        - sudo apt-get update -qq
-        - sudo apt-get install -y dc
-        - wget http://dist.schmorp.de/libev/Attic/libev-4.15.tar.gz
-        - tar -xzvf libev-4.15.tar.gz
-        - cd libev-4.15 && ./configure --prefix=/usr/ && make && sudo make install
-        - cd ..
-        - git clone https://github.com/regrant/portals4.git portals4-code
-        - cd portals4-code
-        - ./autogen.sh
-        - ./configure --prefix=$HOME/portals4/ --with-libev=/usr/lib/ --enable-zero-mrs --enable-reliable-udp --enable-transport-shmem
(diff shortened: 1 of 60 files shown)


jdinan:
Huzzah!  All checks pass -- compile with Portals, compile with OFI, unit tests with OFI, UH C/Fortran feature tests with OFI.  :+1:

jdinan:
Opened #57, fix global_exit.

regrant:
:+1: Looks fine to me. I'll take a look at the P4 udp code to see if I can track down why it's hanging as it only seems to be happening on the travis CI testbeds.

jdinan:
It's likely that the hang was caused by the global_exit issue.  But, P4 UDP still errors out on multiple tests, which causes Travis to report a failed build for SOS even though P4 is the source of the error.

regrant:
P4 UDP is passing on the main portals repo for all ~1200 tests, but I still want to follow up on this to confirm that it might not be a reference implementation issue, as it might just be behaviour that's not getting captured in the P4 test suite.

jdinan:
That's good to know.  There are several tests that have always failed for me when I use P4 UDP on my Mac.  I typically pass configure '--enable-transport-udp --disable-transport-ib --enable-zero-mrs'.

regrant:
try using --enable-reliable-udp that should help. The problem with the macs is the limited ethernet buffering that exists and the tests overload things causing some issues. I'm not all that surprised on the mac, I've had some issues in the past as well. It should be reasonable on Linux boxes though, from my testing. Admittedly the UDP code doesn't get tested as often as the IB stuff though, so bugs do sneak in...

jdinan:
I just tried it, without success.  Some tests fail, others hang and have to be killed.

regrant:
Ok, can we take a look next week? I'd like to understand what you're seeing error wise so I can track this down...

Issue #57, opened 2015-12-04
Global Exit Busted


jdinan:
The current shmem_global_exit() implementation is not up to scratch.  PEs sitting in shmem_barrier when shmem_global_exit is called end up hanging or erroring out without the desired exit code.

jdinan:
Resolved in 2f8c4beb2ab5bbf133d6c36d9def41032050102b

PR #59, opened 2015-12-04
Fix for global_exit

jdinan:
This closes issue #57.


jdinan:
Patch 2f8c4be should also be applied to Portals 4, and the version of hydra included with Portals 4 needs to also be updated to include support for the new PMI abort command.  This looks like it could be tricky, because it looks like there are local changes to Hydra in the Portals 4 repository.  In the meantime, we can build Portals 4 without PMI support and use the PMI support from Sandia-SHMEM with an external launcher.

PR #60, opened 2015-12-05
Fix for global_exit

jdinan:
This closes issue #57.

Patch 2f8c4be should also be applied to Portals 4, and the version of hydra included with Portals 4 needs to also be updated to include support for the new PMI abort command. This looks like it could be tricky, because it looks like there are local changes to Hydra in the Portals 4 repository. In the meantime, we can build Portals 4 without PMI support and use the PMI support from Sandia-SHMEM with an external launcher.

### DIFF
--- a/shmem_pmi/simple_pmi.c
+++ b/shmem_pmi/simple_pmi.c
@@ -314,14 +314,21 @@ int PMI_Finalize( void )
 
 int PMI_Abort(int exit_code, const char error_msg[])
 {
-    PMIU_printf(1, "aborting job:\n%s\n", error_msg);
-    MPIU_Exit(exit_code);
+    char buf[PMIU_MAXLINE];
+
+    /* include exit_code in the abort command */
+    MPIU_Snprintf( buf, PMIU_MAXLINE, "cmd=abort exitcode=%d\n", exit_code);
+
+    PMIU_printf(PMI_debug, "aborting job:\n%s\n", error_msg);
+    GetResponse( buf, "", 0 );
+
+    /* the above command should not return */
     return -1;
 }
 
 /************************************* Keymap functions **********************/
 
-/*FIXME: need to return an error if the value of the kvs name returned is
+/*FIXME: need to return an error if the value of the kvs name returned is 
   truncated because it is larger than length */
 /* FIXME: My name should be cached rather than re-acquired, as it is
    unchanging (after singleton init) */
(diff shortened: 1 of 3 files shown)


jdinan:
Ugh, Travis is hanging and not running any of the tests, but I haven't changed anything since the previous PR #59 when this passed on 32e2ea9...

jdinan:
It looks like our build was just migrated to the new legacy infrastructure, https://blog.travis-ci.com/2015-11-27-moving-to-a-more-elastic-future.  Either these are growing pains that will go away on their own, or we're going to have to debug the testing scripts all over again...

regrant:
Crud. The new infrastructure doesn't allow root level access so installing some of the dependencies could be tough... I'll have to take a close look at the changes.

Have you tried restarting the ci runs from the travis interface?

jdinan:
We're still on the legacy infrastructure, which allows root access.  They
just finished rolling out new legacy infrastructure and migrated us
yesterday.  The container based infrastructure should be doable for us, but
I'm not sure it will solve the issue with tests hanging.

regrant:
The container infrastructure was what I was talking about. There's some limitations to what you can do as root there, so I'll have to double check to see if we should just upgrade to dockers or not.

jdinan:
Looking over the build history, these failures are definitely correlated with our testing being moved from the old OpenCV infrastructure over to Google Compute Engine instances.  The build process is still working fine, testing just hangs when we try to run any SHMEM programs.

As far as moving to containers goes, it shouldn't be hard, I had just punted for now because I wanted to get to a working state before making that switch.  We can easily move libev to a local install (I already do this on other machines).  We'll just have to figure out what to go about the packages we're installing from apt.

jdinan:
I tried a build with timeouts on the unit tests and roughly half of the tests succeeded.  I'm not convinced that this is an error in the environment; it could be a race/deadlock in our initialization code that we just haven't seen before.  The old environment gave us 1.5 cores, I'm not sure what we get from GCE, but it's seems likely that PEs are context switching frequently.

jdinan:
I found the bug.  I don't have a great explanation for why -- apart from it just makes sense -- but we needed a runtime barrier before finalizing the transport.  A barrier semantic is expected from shmem_finalize, and I think that a barrier during shutdown should also be consistent with the legacy atexit shutdown method.

PR #62, opened 2015-12-06
Global exit fix

jdinan:
Cleanup of PR #60.  Adds 2c7270f, which fixes deadlock issue.


jdinan:
@regrant, this resolves the deadlock issue that we have been tracking since testing was moved over to GCE instances.

jdinan:
@kseager, could you take a look at 2c7270f and check whether this is the correct way to quiesce locally before taking down the OFI transport?  Also, is there anything we should do to ensure that OFI sends pending ack responses before finalizing?

PR #63, opened 2015-12-07
OFI: error handling clean-up

kseager:
ofi_init: use inits expected error handling (return the error vs. abort on spot)
	-unify usage of OFI_ABORT and check if address exists before fi_close
ofi_comm: fixed some indentation, add OFI error translation to error handling output

### DIFF
--- a/src/transport_ofi.c
+++ b/src/transport_ofi.c
@@ -132,7 +132,7 @@ static inline int allocate_endpoints(struct fi_info * p_info)
     ret = fi_endpoint(shmem_transport_ofi_domainfd,
 		    p_info, &shmem_transport_ofi_epfd, NULL);
     if(ret!=0){
-	OFI_ABORT("epfd creation failed\n");
+	OFI_ERRMSG("epfd creation failed\n");
 	return ret;
     }
 
@@ -147,7 +147,7 @@ static inline int allocate_endpoints(struct fi_info * p_info)
     ret = fi_endpoint(shmem_transport_ofi_domainfd,
 		    p_info, &shmem_transport_ofi_cntr_epfd, NULL);
     if(ret!=0){
-	OFI_ABORT("cntr_epfd creation failed\n");
+	OFI_ERRMSG("cntr_epfd creation failed\n");
 	return ret;
     }
 
@@ -166,14 +166,14 @@ static inline int bind_resources_to_and_enable_ep(void)
     ret = fi_ep_bind(shmem_transport_ofi_epfd,
 		    &shmem_transport_ofi_stx->fid, 0);
     if(ret!=0){
-	OFI_ABORT("ep_bind epfd2stx failed\n");
+	OFI_ERRMSG("ep_bind epfd2stx failed\n");
 	return ret;
     }
 
     ret = fi_ep_bind(shmem_transport_ofi_cntr_epfd,
 		    &shmem_transport_ofi_stx->fid, 0);
     if(ret!=0){
-	OFI_ABORT("ep_bind cntr_epfd2stx failed\n");
+	OFI_ERRMSG("ep_bind cntr_epfd2stx failed\n");
 	return ret;
     }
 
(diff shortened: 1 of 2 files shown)


src/transport_ofi.c
```diff
@@ -132,7 +132,8 @@ static inline void allocate_endpoints(struct fi_info * p_info)
     ret = fi_endpoint(shmem_transport_ofi_domainfd,
 		    p_info, &shmem_transport_ofi_epfd, NULL);
     if(ret!=0){
-	RAISE_ERROR(ret);
+	OFI_ABORT("epfd creation failed\n");
```
jdinan: Could you rename this macro to OFI_ERRMSG()?

src/transport_ofi.c
```diff
@@ -387,13 +415,17 @@ static inline void exchange_and_av_insert(const int npes)
...
-	RAISE_ERROR(ret);
-    }
+	OFI_ABORT("av insert failed\n");
+	return ret;
+    } else
+	ret = 0;
 
     free(alladdrs);
+
+    return ret;
```
jdinan: Would it be cleaner to return 0 here?  Or is there an instance where you can reach this statement and have nonzero ret?  This would allow you to delete "else ret = 0" above.

src/transport_ofi.c
```diff
@@ -479,31 +515,35 @@ static inline struct fi_info * query_for_fabric(char *provname)
...
+		      NULL, NULL, 0, &hints, p_info);
     if(ret!=0){
-	RAISE_ERROR(ret);
+	OFI_ABORT("getinfo didn't find any providers\n");
+	return ret;
     }
 
-    if (!p_info) {
-	    OFI_ABORT("fi_getinfo didn't find any providers\n");
+    if(*p_info && (*p_info)->tx_attr->inject_size >= sizeof(long double)) {
```
jdinan: Is this assuming that long double is the longest type used by SHMEM?  I'm not sure if the C standard allows double complex to be longer than long double?

src/transport_ofi.c
```diff
@@ -518,26 +558,40 @@ int shmem_transport_init(long eager_size)
...
+    if(ret!=0)
+	return ret;
 
     fi_freeinfo(p_info);
 
     if(provname)
       p_info->fabric_attr->prov_name = NULL;
 
-    return 0;
+    return ret;
```
jdinan: Similar comment as above -- I think I prefer return 0 here, unless there's a way to reach this statement after an error has occurred.

src/transport_ofi.h
```diff
@@ -120,7 +120,9 @@ extern shmem_free_list_t *shmem_transport_ofi_frag_buffers;
 #define OFI_ABORT(...) { fprintf(stderr, __FILE__ ":%d: \n", __LINE__); \
                             fprintf(stderr, __VA_ARGS__);  }
 
-#define OFI_RET_CHECK(ret) do { if (ret) { RAISE_ERROR(ret); } } while (0)
+#define OFI_RET_CHECK(ret) do { if (ret) { \
+	fprintf(stderr,"OFI error #%d: %s \n", (int)ret, fi_strerror(ret)); \
```
jdinan: Is ret ever not an int?  I would prefer to drop this type cast if we can.  Type casts can mask bugs.

src/transport_ofi.h
```diff
@@ -150,37 +152,37 @@ void shmem_transport_ofi_drain_cq(void)
...
-		}
+				OFI_RET_CHECK(ret);
+			}
 		}
 
 		if(ret == 1) {
-	shmem_transport_ofi_frag_t *frag =
-			container_of(buf.op_context,
-			struct shmem_transport_ofi_frag_t,
-		     context);
```
jdinan: It's a good practice to isolate whitespace changes to separate commits.  Please don't bother making a change for this PR.

jdinan:
Looks good -- thanks!  Can you review the comments when you get a chance?

jdinan:
It looks like there is an assumption that 'long double' is the largest size type.  This is not actually guaranteed by the C language standard.  For determining things like the inject size, we should be checking to find the largest size, like ```largest_type_size = max(sizeof(long double), sizeof(double complex))```.  There may also be a need to consider the Fortran types (I don't recall if the user can do things like set the size of real to be 128-bit).  Those sizes are captured at configure time in src/config.h.

In any case, this patch looks good.  Could you create an issue or patch for the above item?

PR #73, opened 2015-12-14
Collectives update: new recursive doubling reduction algorithm 

kseager:
The recursive double algorithm relies on software atomics to enable fabrics with atomic data type limitations to still be able to do a reduction.


jdinan:
psync_tail1 is no longer used.  Could you delete this variable and give psync_tail2 a more meaningful name?  Maybe something like "extra_peer_notify"?

jdinan:
Great work!  :smile:

PR #76, opened 2015-12-17
Clean up Travis build and add PRK to CI testing

jdinan:


jdinan:
Some of the PRK kernels are disabled because of bugs in the kernel.  I expect to re-enable them later this week when the PRK team pushes out a new release.

PR #79, opened 2015-12-18
Remove dependence on GNU dc and update Travis

jdinan:


jdinan:
After we merge this change, we will be one dependence away from being sudo free!  :grinning:

jdinan:
@regrant or @kseager, do you have time to review this change today?

This was made at the request of a user who is having trouble building because of the dc dependence.

PR #84, opened 2015-12-22
add rpath option

jeffhammond:
This adds `-Wl,-rpath -Wl,$libdir` to the wrapper LDFLAGS so that one need not set `LD_LIBRARY_PATH` for SOS libraries in order to use them.

This is convenient if one intends to have multiple installs of SOS and rely upon the uniqueness of their respective wrappers to get the right shared library.

This does not rpath libfabric.  I think I can add that later, as an option to `--enable-wrapper-rpath`.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -107,6 +107,12 @@ AC_ARG_ENABLE([libtool-wrapper],
 AS_IF([test "$enable_libtool_wrapper" = "no"], [LIBTOOL_WRAPPER_LDFLAGS="-no-install"], [LIBTOOL_WRAPPER_LDFLAGS=])
 AC_SUBST([LIBTOOL_WRAPPER_LDFLAGS])
 
+AC_ARG_ENABLE([wrapper-rpath],
+  [AS_HELP_STRING([--enable-wrapper-rpath],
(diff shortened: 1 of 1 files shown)


jdinan:
Thanks for looking into this.  I'd like to look at the problem a little more before we settle on a solution.  I'm not sure whether the right place to fix it is in configure or in the wrappers.  If we do it in configure, I'd like it to be automatic.  I think the wrappers are also cribbed from Open MPI; I'd like to see how they solve this and whether it's a solution we can adopt.

hppritcha:
:+1:

This configury change makes me much happier when trying to use sandia shmem on cray.  Now I don't need to fiddle with LD_LIBRARY_PATH to avoid picking up the cray sma library.  It also avoids the having libfabric appear linked in twice (first by the sma dependency on libfabric), and then the extralibs having libfabric in it.

jdinan:
Is there a good reason for this to be optional?

jeffhammond:
@jdinan https://github.com/rust-lang/rust/issues/5219 appears to contain some arguments...

jdinan:
Superseded by #101

PR #85, opened 2015-12-23
OFI support for MR_BASIC

jdinan:
Add support for MR_BASIC.

This patch should be reviewed by our OFI expert, @kseager, before merging.


kseager:
this looks really good, great job! thanks for putting this patch together!

jdinan:
Thanks for the review!  The next order of business is to decide where we want to integrate this change.  We could mark the current master as v1.2.0rc2, merge this, and mark it as v1.2.0rc3.  Alternatively, we could also put this into a v1.2.1 branch.  However, v1.3.0 coming shortly after the new OpenSHMEM spec passes and I don't want us to have to do too much release juggling.  And, given the timing of v1.3.0, we might not end up making a v1.2.1 release.

@kseager @regrant -- What do you think about the first approach?

Issue #86, opened 2015-12-30
libtoolize: command not found


jeffhammond:
From https://travis-ci.org/jeffhammond/PRK/jobs/99387697:
```sh
+cd /Users/travis/PRK-deps
+git clone --depth 10 https://github.com/regrant/sandia-shmem.git
Cloning into 'sandia-shmem'...
remote: Counting objects: 252, done.
remote: Compressing objects: 100% (193/193), done.
remote: Total 252 (delta 99), reused 116 (delta 58), pack-reused 0
Receiving objects: 100% (252/252), 267.67 KiB | 0 bytes/s, done.
Resolving deltas: 100% (99/99), done.
Checking connectivity... done.
+cd sandia-shmem
+./autogen.sh
+test -d ./config
+aclocal -I config
+libtoolize --force --copy
./autogen.sh: line 6: libtoolize: command not found
+autoheader
+automake --foreign --add-missing --copy
```

This is addressed on [StackOverflow](http://stackoverflow.com/a/15448876/2189128).

A patch is in-progress.

jdinan:
I think this was fixed.  Please re-open if I'm missing something.

PR #87, opened 2015-12-30
use glibtoolize on Mac

jeffhammond:
see
http://stackoverflow.com/questions/15448582/installed-libtool-but-libtoolize-not-found
for a description of the problem.

fixes issue #86.

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -3,7 +3,10 @@
 set -x
 test -d ./config || mkdir ./config
 aclocal -I config
-libtoolize --force --copy
+case `uname` in
+    Darwin*) glibtoolize --force --copy ;;
+    *)        libtoolize --force --copy ;;
+esac
 autoheader
 automake --foreign --add-missing --copy
 autoconf
--- a/shmem_pmi/autogen.sh
+++ b/shmem_pmi/autogen.sh
@@ -3,7 +3,10 @@
 set -x
 test -d ./config || mkdir ./config
 aclocal -I config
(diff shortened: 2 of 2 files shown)


jdinan:
Why predicate on os?  I think it would be better to look at the env var,
then search for libtoolize, then search for glibtoolize, then error if not
found.

jeffhammond:
That's fine with me too.  I just want PRK SHMEM to pass CI with OSX.

jdinan:
I symlink $HOME/bin/libtoolize to glibtoolize, because this issue breaks
many builds, not just SOS.  What do libfabric and/or Portals do?  Would
they need a workaround for this, too?

jeffhammond:
Neither [libfabric](https://github.com/ofiwg/libfabric/blob/master/autogen.sh) nor [Portals4](https://github.com/regrant/portals4/blob/master/autogen.sh) require libtoolize.

jdinan:
Surely they do.  It looks like they use autoreconf instead of calling the
tools directly.  That is probably the right way to fix this.

jeffhammond:
Sorry, all I was saying is that their `autogen.sh` do not call `libtoolize`.  I do not know what happens behind the curtain.

jdinan:
My understanding is that autoreconf was added to the autotools to address
these sort of issues.  It will take a little work, but we should add an
issue to get compliant with that approach.  In the meantime, I have less
concern about merging a band aid patch if a proper fix is in the making.

PR #92, opened 2016-01-12
small memory management clean-up init/ofi

kseager:
+included added error checking to environment variable conversion

### DIFF
--- a/src/init.c
+++ b/src/init.c
@@ -92,9 +92,8 @@ atol_scaled(char *s)
     errno = 0;
 
     val = strtol(s,&e,0);
-    if(errno != 0) {
-        perror("env var conversion");
-        exit(1);
+    if(errno != 0 || e == s) {
+        shmem_runtime_abort(1, "env var conversion");
     }
     if (e == NULL || *e =='\0')
         return val;
--- a/src/init.c
+++ b/src/init.c
@@ -67,6 +67,21 @@ static char shmem_internal_my_hostname[MAXHOSTNAMELEN];
 static char shmem_internal_my_hostname[HOST_NAME_MAX];
 #endif
 
+
+static inline long errchk_atol(char *s)
+{
+    long val;
+    char *e;
+    errno = 0;
+
+    val = strtol(s,&e,0);
+    if(errno != 0) {
+        perror("env var conversion");
+        exit(1);
+    }
+
+    return val;
+}
  
 /* atol() + optional scaled suffix recognition: 1K, 2M, 3G, 1T */
 static long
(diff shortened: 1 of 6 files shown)


kseager:
OFI has an atomic max size we grab from atomicvalid.

in terms of the init assert, sure I can change it to an if statement, personally I think it would be nice to compile in two modes, one where you're developing a provider and need additionally safeguards, and one where you aren't, thus I liked the idea of an assert

jdinan:
The checks that we typically compile out (in SOS they are actually opt-in with --enable-error-checking) are checks for whether or not the program uses the API correctly.  In this case, a correct program could run into something that we don't support properly.  I think we need to always have this error check, until it's fixed.

src/init.c
```diff
@@ -67,6 +67,21 @@ static char shmem_internal_my_hostname[MAXHOSTNAMELEN];
...
+static inline long errchk_atol(char *s)
+{
+    long val;
+    char *e;
+    errno = 0;
+
+    val = strtol(s,&e,0);
+    if(errno != 0) {
+        perror("env var conversion");
+        exit(1);
```
jdinan: If the runtime is already initialized, we should use shmem_runtime_abort(...) here.

src/init.c
```diff
@@ -67,6 +67,21 @@ static char shmem_internal_my_hostname[MAXHOSTNAMELEN];
...
+    }
```
jdinan: Unfortunately, the strtol error check doesn't do what you are hoping for.  See: http://stackoverflow.com/questions/9445171/strtol-using-errno

The better way to do this is to use sscanf:
```
int nread = sscanf(s, "%ld", &e);
if (nread != 1) {
  shmem_runtime_abort(1, "Error parsing environment variable");
}

File: src/init.c [hunk 3]
```diff
@@ -74,8 +89,13 @@ atol_scaled(char *s)
 {
     long val;
     char *e;
+    errno = 0;
 
     val = strtol(s,&e,0);
```
jdinan: Same comment as above, use sscanf instead of strtol.

src/init.c
```diff
@@ -67,6 +67,21 @@ static char shmem_internal_my_hostname[MAXHOSTNAMELEN];
...
 
+
+static inline long errchk_atol(char *s)
+{
+    long val;
+    char *e;
+    errno = 0;
+
+    val = strtol(s,&e,0);
+    if(errno != 0) {
```
jdinan: Need to also check for "e == s" which indicates a parsing error.

jdinan:
Thanks!

Issue #77, opened 2015-12-17
Potential shmem_test_lock race


jdinan:
The UH test ```feature_tests/C/test_shmem_lock.c``` occasionally hangs after printing:

```Test for set, and clear lock: Passed```

There could be a race in the ensuing test of ```shmem_test_lock()```.  So far, I've only observed this hang in the Travis environment.

jdinan:
This was cause by a bug in the sockets provider that has since been fixed.

PR #96, opened 2016-01-14
Fix race in shmem_lock_clear

jdinan:
Correct a race in clear_lock that occurred in the sequence (1) check if
next PE pointer (lock->data) was updated (2) else wait for update.  In
the second step, lock->data was re-read rather than using the value from
the first step.  As a result, if an update arrived between steps,
shmem_lock_clear could deadlock.


jdinan:
This should close #77 and resolve the transient failures we have been seeing in Travis testing.

PR #97, opened 2016-01-14
Fix races in shmem_lock_set/clear

jdinan:
Correct races in set_lock and clear_lock that occurred in the sequence
(1) check if lock object was updated (2) else wait for update.  In the
second step, the lock object was re-read rather than using the value
read in the first step.  As a result, if an update arrived between
steps, shmem_lock_clear could wait using the new value as the wait
condition, resulting in a deadlock.

This should close issue #77.


jdinan:
Argh, something is still busted...

PR #105, opened 2016-01-23
Change type of datatype size table to size_t

jdinan:


jdinan:
CI testing passed.  We have transient failures at the moment because of a deadlock issue with atomic compare-and-swap in the OFI transport.

PR #106, opened 2016-01-23
Add support for put/get larger than provider's max msg size

jdinan:


jdinan:
CI testing failed.  I'll look into it:

```
FAIL: bigget (exit: 1)
======================
OFI error #-11: Unknown error -11 
[001] ERROR: transport_ofi.h:372 return code -11
```

jdinan:
The error that is being reported is ```-FI_EAGAIN```, which I think indicates that ```try_again()``` is exceeding ```shmem_transport_ofi_max_poll``` (currently ```1<<20```).  @kseager, any suggestions apart from increasing the maximum allowed number of polling cycles?  Should ```try_again()``` yield the core or throttle the issue rate after a certain number of polling cycles?

jdinan:
Follow-up thought: The problem with yielding in the SOS code is that we don't know whether it will help or not, since SOS is not multithreaded.  Perhaps, the sockets provider should yield to the progress thread after some amount of EAGAIN to try and free up resources?

jdinan:
Increasing the polling limit seems to have resolved the problem.  We may want to ping Jithin to see if there's potentially a solution internal to the sockets provider that pokes progress occasionally.

src/transport_ofi.h
```diff
@@ -461,23 +507,45 @@ shmem_transport_get(void *target, const void *source, size_t len, int pe)
...
+                              NULL,
+                              GET_DEST(dst),
+                              (uint64_t) addr,
+                              key,
+                              NULL);
+            } while (try_again(ret,&polled));
+
+            shmem_transport_ofi_pending_get_counter++;
+        }
+        else {
```
kseager: why not use a counter instead of the queue

test/unit/bigput.c
```diff
@@ -196,6 +196,14 @@ main(int argc, char **argv)
...
     shmem_barrier_all();
 
+    for (i = 0; i < elements; i++) {
+      if (Target[i] != i + 1) {
+          printf("%d: Error Target[%d] = %d, expected %d\n",
+                 me, i, Target[i], i + 1);
+          shmem_global_exit(1);
+      }
+    }
+
```
kseager: this is a great catch! thanks for the add

src/transport_ofi.h
```diff
@@ -417,26 +416,42 @@ shmem_transport_put_nb(void *target, const void *source, size_t len,
...
-		long_frag->reference++;
-		shmem_transport_ofi_pending_cq_count++;
+                    (*(long_frag->completion))++;
+                    long_frag->reference++;
+                    shmem_transport_ofi_pending_cq_count++;
+
+                    frag_source += frag_len;
+                    frag_target += frag_len;
+                }
 	}
```
kseager: put_nb last two cases: I don't think the counter usage is going to quite work as is...you can't assume buffer reuse after this returns since put_wait doesn't force a counter wait as it does for get, however do you even need to separate these two cases? maybe you could just keep the else and get rid of the else if, the else won't loop unless its an overflow so it should work for either case

jdinan: I don't think I follow your comment about counter usage.  Put_wait should wait for the operation's completion flag to be set, which only happens after all of the frags for the given put have completed.  Or were you commenting on a design where we remove the frags and use a counter instead?  In that model, we would remove the completion flag and wait for all puts to complete in put_wait.

Re: two cases -- The cases are separated because we don't want to allocate the frag object for puts below the max message size.  We could merge them, but it would add instructions to the critical path for non-fragmented puts.

kseager: thats true about the instruction add when you combine the two cases

what I was referring to earlier with the counter case was that nothing in put_wait forces the counter completion and as long as we are blocking we need to enforce that.
So I guess we can either add a counter wait in put_wait or combine the two cases....

jdinan: Oh, I'm an idiot, I see your point now.  I think you're right; the right fix for now is to use frags for any puts that need to be completed via put_wait.  I'll go ahead and collapse these cases.

It looks like the PRK build broke; not an SOS issue.  I'll restart the Travis tests.  I think this issue has been fixed on their end.

PR #110, opened 2016-01-25
look for mpirun before mpiexec

jeffhammond:
...because Intel MPI provides MPD-based mpiexec.

Intel MPI uses gawdawful MPD by default for mpiexec.  On machines where mpiexec is in the path, this leads to very sad pandas, because all the tests fail.

### DIFF
--- a/src/shmem_launcher_script.in
+++ b/src/shmem_launcher_script.in
@@ -11,7 +11,7 @@
 # information, see the LICENSE file in the top level directory of the
 # distribution.
 
-SEARCH_LAUNCHERS="yod mpiexec mpirun srun"
+SEARCH_LAUNCHERS="yod mpirun mpiexec srun"
 LAUNCHER=""
 
 # Users can set the OSHRUN_LAUNCHER environment variable to specify the command


jdinan:
Does mpirun work for you?  If so, which mpirun is it?  Unfortunately, I don't know of a good way to determine if a launcher supports PMI-1.  OpenMPI, for example, doesn't.  We could be more conservative and set the search list to "mpiexec.hydra yod srun" and then ask the user to set OSHRUN_LAUNCHER if none of those are found.

jeffhammond:
`mpirun` from Intel MPI works, as does any from Open-MPI.

Can't `oshrun` parse `mpirun --version` to see if it is a compatible MPICH?  It's a shell script, no?
```
$ mpirun --version
HYDRA build details:
    Version:                                 3.2
    Release Date:                            unreleased development copy
    CC:                              gcc-5    
    CXX:                             g++-5    
    F77:                             gfortran-5   
    F90:                             gfortran-5   
    Configure options:                       '--disable-option-checking' '--prefix=/opt/mpich/dev/gcc/default' 'CC=gcc-5' 'CXX=g++-5' 'FC=gfortran-5' 'F77=gfortran-5' '--enable-cxx' '--enable-fortran' '--enable-threads=runtime' '--enable-g=dbg' '--with-pm=hydra' '--enable-wrapper-rpath' '--enable-static' '--enable-shared' '--cache-file=/dev/null' '--srcdir=../../../../src/pm/hydra' 'CFLAGS= -g -O2' 'LDFLAGS=' 'LIBS=-lpthread ' 'CPPFLAGS= -I/Users/jrhammon/Work/MPI/MPICH/git/build/src/mpl/include -I/Users/jrhammon/Work/MPI/MPICH/git/src/mpl/include -I/Users/jrhammon/Work/MPI/MPICH/git/src/openpa/src -I/Users/jrhammon/Work/MPI/MPICH/git/build/src/openpa/src -D_REENTRANT -I/Users/jrhammon/Work/MPI/MPICH/git/build/src/mpi/romio/include'
    Process Manager:                         pmi
    Launchers available:                     ssh rsh fork slurm ll lsf sge manual persist
    Topology libraries available:            hwloc
    Resource management kernels available:   user slurm ll lsf sge pbs cobalt
    Checkpointing libraries available:       
    Demux engines available:                 poll select
```

jeffhammond:
I'd really like to see Sandia OpenSHMEM have a nice `oshrun`, because it is my intent to borrow it for OSHMPI ;-)

PR #117, opened 2016-02-05
test/performance: new shmem put performance test

kseager:
+small put ping-pong bw/latency
+varying message size ping bw/latency

### DIFF
--- a/test/performance/Makefile.am
+++ b/test/performance/Makefile.am
@@ -17,12 +17,14 @@ if ENABLE_LENGTHY_TESTS
 TESTS += \
 	shmemlatency \
 	msgrate \
-	shmem_put_perf
+	shmem_put_perf \
+	shmem_bw_put_perf
 else
 check_PROGRAMS += \
         shmemlatency \
 	msgrate \
-	shmem_put_perf
+	shmem_put_perf \
+	shmem_bw_put_perf
 endif
 
 NPROCS ?= 2
--- a/test/performance/Makefile.am
+++ b/test/performance/Makefile.am
@@ -15,12 +15,14 @@ TESTS =
 
 if ENABLE_LENGTHY_TESTS
 TESTS += \
-        shmemlatency \
-	msgrate
+	shmemlatency \
+	msgrate \
+	shmem_put_perf
 else
 check_PROGRAMS += \
         shmemlatency \
-	msgrate
+	msgrate \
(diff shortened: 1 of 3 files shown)


test/performance/Makefile.am
```diff
@@ -11,7 +11,8 @@
 
 check_PROGRAMS = $(TESTS)
 
-TESTS =
+TESTS = \
+	shmem_put_perf
```
jdinan: This should be below under ENABLE_LENGHTY_TESTS, rather than unconditionally including it in TESTS.  The test takes too long to run in the Travis environment and is causing Travis to timeout.

kseager: just did an amend for this!

jdinan: It also need to be added to ```check_PROGRAMS``` so that we always compile the test case, even when we don't run it as part of ```make check```.  Part of the bit rot prevention program.  :)

jhrobichaux:
Did a quick glance at the code, have a couple of initial comments. 

1.  Buffer 'aligned_buf' is allocated in the global .data section of the executable. This could be problematiice for a couple of reasons. 
      A. If in future want to run test with HBM or other memory the .bss & .data sections will typically be allocated to the "default" memory type for the system which may or may not be HBM. 
     B. Because it is allocated in the the global memory section the data affinity for the memory "may" not be in the same numa domain as the one the process actually runs on. 
      C. The osu_oshm_put.c does provides two mechanisms for allocating the symmetric variables, one using global data and the other using the symmetric heap allocation. Runs Ulf and I did on the Cray inidicated that the dynamic allocation of the key variable used for the data transfer always resulted in more consistent, and often better performance. I would like to preproduce this on the Cray again or with a version of SoS but our hypothesis is the performance difference was likely attributed to 1B above. 

2. However you decide to allocate the data buffer would always suggest initiating the memory before it is used. For small bufferens this should ensure that the buffer is in cache and paged in avoiding cache & TLB misses when actually doing the operations being measured. Secondly, if it is dynamically allocated then the initilizatin will ensure that the memory is also affinitized to the local core the task is running on if the standard first touch page allocation policy is enabled. 

That's a mouth full.

kseager:
thanks Joe! I'll change the buffer usage scheme

jdinan:
Could you please add sign-off to your commits?

PR #118, opened 2016-02-05
Code cleanups

jdinan:
Clean up handling of environment variables and resolve several build warnings.


kseager:
I like that you pulled out the getenv work into its own file

PR #122, opened 2016-02-08
OFI: inject usage modification

### LINKED ISSUE #72: correct size for inject
It looks like there is an assumption that 'long double' is the largest size type. This is not actually guaranteed by the C language standard. For determining things like the inject size, we should be checking to find the largest size, like largest_type_size = max(sizeof(long double), sizeof(double complex)). There may also be a need to consider the Fortran types (I don't recall if the user can do things like set the size of real to be 128-bit). Those sizes are captured at configure time in src/config.h.

kseager:
+if provider enables a larger inject threshold use it
+assert that put_small is staying within inject boundaries

closes #72

### DIFF
--- a/src/transport_ofi.c
+++ b/src/transport_ofi.c
@@ -718,7 +718,7 @@ static inline int query_for_fabric(struct fi_info ** p_info, char *provname)
     ep_attr.type              = FI_EP_RDM; /* reliable connectionless */
     hints.fabric_attr	      = &fabric_attr;
     tx_attr.op_flags          = FI_DELIVERY_COMPLETE;
-    tx_attr.inject_size       = shmem_transport_ofi_max_buffered_send;
+    tx_attr.inject_size       = shmem_transport_ofi_max_buffered_send; /*require provider to support this as a min*/
     hints.tx_attr	      = &tx_attr; /* TODO: fill tx_attr */
     hints.rx_attr	      = NULL;
     hints.ep_attr             = &ep_attr;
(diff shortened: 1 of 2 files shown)


src/transport_ofi.c
```diff
@@ -718,7 +718,7 @@ static inline int query_for_fabric(struct fi_info ** p_info, char *provname)
     ep_attr.type              = FI_EP_RDM; /* reliable connectionless */
     hints.fabric_attr	      = &fabric_attr;
     tx_attr.op_flags          = FI_DELIVERY_COMPLETE;
-    tx_attr.inject_size       = shmem_transport_ofi_max_buffered_send;
+    tx_attr.inject_size       = shmem_transport_ofi_max_buffered_send; /*require provider to support this as a min*/
```
jdinan: How does this part of the API work?  If we request something less than the max possible in the provider will it return its max or the value we asked for?  Should we be asking for more?

kseager: essentially this is telling the provider you *must* support at least this inject size, the provider will check if their supported inject is at least this size, if it is less than the requested inject size it will abort saying that it cannot support your requirements. Thus we should treat it as a minimal allowable requirement

sayantansur:
Did Github automatically close this issue?

I thought if you put "closes #72" in the commit message, then Github should automatically close the issue. Wonder if that works if you mention it in the PR?

jdinan:
Github closed #72 when I merged this PR.

PR #128, opened 2016-02-20
V1.3.0 dev

parthapkundu:


jdinan:
This PR looks like it may have been opened in error.  If you want to download the code, you can clone the repository (```$ git clone https://github.com/Sandia-OpenSHMEM/SOS.git```).  You can also download a zip file from the releases page (https://github.com/Sandia-OpenSHMEM/SOS/releases).

PR #132, opened 2016-02-25
Improvements to MR_SCALABLE support

jdinan:


jdinan:
@sayantansur, this should address the MR_SCALABLE changes that you suggested.

PR #137, opened 2016-02-26
test/performance added shmem_perf test suite

kseager:
includes:
+bi-directional and uni-directional tests for put/get
+latency put/get tests
+README for usage info


kseager:
@jdinan please review

configure.ac
```diff
@@ -369,7 +369,9 @@ AC_CONFIG_FILES([Makefile
   src/Makefile
   test/Makefile
   test/unit/Makefile
-  test/performance/Makefile])
+  test/performance/Makefile
+  test/performance/shmem_perf_suite/Makefile
+  test/performance/tests/Makefile])
```
jdinan: Could you check the RPM scripts to see if the added directories impact the RPM build?

test/performance/shmem_perf_suite/bw_common.h
```diff
@@ -0,0 +1,358 @@
...
+#define WINDOW_SIZE 64
+#define WARMUP 50
+
+#define TRIALS_LARGE  100
+#define WINDOW_SIZE_LARGE 64
+#define WARMUP_LARGE  10
+#define LARGE_MESSAGE_SIZE  8192
+
+#define TRUE  (1)
+#define FALSE (0)
```
jdinan: You could use stdbool.h for true/false, or if you'd rather define them here, this approach: http://stackoverflow.com/questions/17010041/why-define-true-1-1-in-a-c-boolean-macro-instead-of-simply-as-1

test/performance/shmem_perf_suite/common.h
```diff
@@ -0,0 +1,83 @@
...
+    buf = (char *) (((unsigned long) buf + (page_align - 1)) /
+            page_align * page_align);
+
+    return buf;
+}
+
+int static inline is_divisible_by_4(int num)
+{
+    assert(sizeof(int) == 4);
+    return (!(num & 0x00000003));
```
jdinan: I think this is unreliable when ```num``` is negative.  I think bitwise arithmetic on signed values is not portable/reliable since C does not define the representation.  Add ```assert(num >= 0)```?

test/performance/shmem_perf_suite/common.h
```diff
@@ -0,0 +1,83 @@
...
+
+/*to be a power of 2 must only have 1 set bit*/
+int static inline is_pow_of_2(unsigned int num)
+{
+    int i = 0;
+    unsigned int count = 0;
+
+    do {
+        if(num & 1)
+            count++;
```
jdinan: I think that you're gathering the population count of ```num``` into ```count```, which is more than you need.  You should be able to replace ```count++``` with ```return num == 1``` and drop the next ```if``` block.

I realize that this code section is likely not on a performance sensitive path, but I can't help myself when it comes to bit twiddling.  :)

test/performance/shmem_perf_suite/shmem_bibw_get_perf.c
```diff
@@ -0,0 +1,75 @@
...
+                start = shmemx_wtime();
+
+            for(j = 0; j < metric_info->window_size; j++)
+                shmem_getmem(metric_info->buf2, metric_info->buf2, len, dest);
+        }
+        end = shmemx_wtime();
+
+        calc_and_print_results((end - start), bw, len, *metric_info, EVEN_SET);
+
+    } else {
```
jdinan: This looks like the same code as above, except for the timing.  Can you just move the ```if (...my_node % 2 == 0)``` in to cover the call to ```calc_and_print_stats```?

test/performance/shmem_perf_suite/shmem_bibw_put_perf.c
```diff
@@ -0,0 +1,80 @@
...
+            for(j = 0; j < metric_info->window_size; j++)
+                shmem_putmem(metric_info->buf, metric_info->buf, len, dest);
+
+            shmem_quiet();
+        }
+        end = shmemx_wtime();
+
+        calc_and_print_results((end - start), bw, len, *metric_info, EVEN_SET);
+
+    } else {
```
jdinan: Same comment as above; I would have all PEs use the same series of loops and gather timings, but only have the even PEs call ```calc_and_print_stats```.

test/performance/shmem_perf_suite/common.h
kseager: good point, this could be condensed, I'll refactor =)

test/performance/shmem_perf_suite/shmem_bibw_get_perf.c
kseager: they use different buffers...I mean I guess I could I might leave this as is

test/performance/shmem_perf_suite/bw_common.h
kseager: agreed; I thought about this at one point, but I think the thought got lost somewhere along the line, thanks for pointing it out

test/performance/shmem_perf_suite/shmem_bibw_get_perf.c
jdinan: Oh, I missed that...

jdinan: Why did you choose to use symmetric buffers for both the source and destination?

test/performance/shmem_perf_suite/common.h
kseager: will do

configure.ac
kseager: we don't include test directory so we're good!

jdinan: What about sandia-openshmem.spec.in:50?
```
install -d %{testdir}/performance
install test/performance/.libs/* %{testdir}/performance
```

test/performance/shmem_perf_suite/shmem_bibw_get_perf.c
kseager: I the two PE's to be able to put to buffers simultaneously so that required two symmetric buffers...and I'm just overloading them I guess..

configure.ac
kseager: good catch =) I'll fix this as well and roll it all up in an amend

Issue #91, opened 2016-01-07
cq usage in OFI transport


kseager:
cq is no longer needed with openSHMEM 1.3 spec which exposes non-blocking API. The options are to either use the cq for counter error handling or remove the queue all together (this isn't yet exposed in OFI).

kseager:
fixed by merged PR #160

kseager:
fixed by merged PR #160

PR #140, opened 2016-03-03
1.3 transport cleanups

kseager:
1) took out long_frag usage from OFI transport
2) refactored nbi pathway in transport
@jdinan review please!
resolves issue #91


jdinan:
Can we drop the last argument from *_put_nbi?  IIUC, this function should be for used for implicitly completed puts, which will never need to supply a completion handle.

src/shmem_comm.h
```diff
@@ -95,6 +95,30 @@ shmem_internal_put_nb(void *target, const void *source, size_t len, int pe,
...
+                      long *completion)
+{
+    int node_rank;
+
+    if (-1 != (node_rank = SHMEM_GET_RANK_SAME_NODE(pe))) {
+#if USE_XPMEM
+        shmem_transport_xpmem_put(target, source, len, pe, node_rank);
+#elif USE_CMA
+        if (len > shmem_transport_cma_put_max) {
+            shmem_transport_put_nb(target, source, len, pe, completion);
```
jdinan: Should this be ```put_nbi``` instead of ```put_nb```?

src/transport_ofi.h
```diff
@@ -354,6 +354,42 @@ shmem_transport_put_small(void *target, const void *source, size_t len, int pe)
 
 static inline
 void
+shmem_transport_put_large(void *target, const void *source, size_t len, int pe)
```
jdinan: Could you change the name to something like ```shmem_transport_ofi_put_large``` so that it's clear this isn't part of the transport layer interface?

kseager: agreed

kseager:
moving to differnent PR

PR #142, opened 2016-03-04
Merge v1.3.0 development branch 

jdinan:


jdinan:
Establishing a PR for code review.

src/atomic_f.c
```diff
@@ -304,3 +305,139 @@ FC_SHMEM_INT8_INC(int64_t *target,
...
+void FC_SHMEM_REAL8_SET(double *dest,
+                        double *value,
+                        fortran_integer_t *pe);
+void
+FC_SHMEM_REAL8_SET(double *dest,
+                   double *value,
+                   fortran_integer_t *pe)
+{
+    SHMEM_ERR_CHECK_INITIALIZED();
+
```
jdinan: add: shmem_internal_assert(sizeof(double) == 8);

src/collectives.c
```diff
@@ -1047,3 +1048,116 @@ shmem_internal_fcollect_recdbl(void *target, const void *source, size_t len,
...
+                        int PE_start, int logPE_stride, int PE_size, long *pSync)
+{
+    const long one = 1;
+    const int stride = 1 << logPE_stride;
+    const int my_as_rank = (shmem_internal_my_pe - PE_start) / stride;
+    const void *dest_ptr = (uint8_t *) dest + my_as_rank * len;
+    int peer;
+
+    if (0 == len)
+        return;
```
jdinan: Log an issue to check if length == 0 is handled correctly for the other collectives?

src/collectives.c
```diff
@@ -1047,3 +1048,116 @@ shmem_internal_fcollect_recdbl(void *target, const void *source, size_t len,
...
+    do {
+        shmem_internal_atomic_small(pSync, &one, sizeof(long), peer,
+                                    SHM_INTERNAL_SUM, DTYPE_LONG);
+        peer = shmem_internal_circular_iter_next(peer, PE_start, logPE_stride,
+                                                 PE_size);
+    } while (peer != shmem_internal_my_pe);
+
+    SHMEM_WAIT_UNTIL(pSync, SHMEM_CMP_EQ, PE_size);
+    *pSync = 0;
+}
```
jdinan: OpenSHMEM spec question: is there a requirement that a given PE should be able to read the data from all other PEs when the collective returns?

src/collectives.c
```diff
@@ -1047,3 +1048,116 @@ shmem_internal_fcollect_recdbl(void *target, const void *source, size_t len,
...
+{
+    const long one = 1;
+    const int stride = 1 << logPE_stride;
+    const int my_as_rank = (shmem_internal_my_pe - PE_start) / stride;
+    const void *dest_base = (uint8_t *) dest + my_as_rank * nelems * dst * elem_size;
+    int peer;
+
+    if (0 == nelems)
+        return;
+
```
jdinan: Need to assert on elem_size to make sure call to put_small is valid.

src/collectives_f.c
```diff
@@ -1222,3 +1262,131 @@ FC_SHMEM_BROADCAST64(void *target,
...
+    SHMEM_ERR_CHECK_ARG_POSITIVE(*sst);
+    SHMEM_ERR_CHECK_ARG_POSITIVE(*dst);
+    SHMEM_ERR_CHECK_ACTIVE_SET(*PE_start, *logPE_stride, *PE_size);
+
+    /* SHMEM_ALLTOALL_SYNC_SIZE is defined to allow this cast */
+    pSync_c = (long*) pSync;
+
+    shmem_internal_alltoalls(target, source, *dst, *sst, 8, *nelems,
+                             *PE_start, *logPE_stride, *PE_size, pSync_c);
+}
```
jdinan: CR ended here

src/data_c.c
```diff
@@ -642,6 +939,132 @@ shmem_getmem(void *target, const void *source, size_t len, int pe)
...
+}
+
+
+void
+shmem_get128_nbi(void *dest, const void *source, size_t nelems, int pe)
+{
+  SHMEM_ERR_CHECK_INITIALIZED();
+
+  shmem_internal_get(dest, source, 16*nelems, pe);
+}
```
jdinan: @stjohnt, is there another unit test in the works that will exercise all of the NBI functions?

src/data_f.c
```diff
@@ -209,6 +210,138 @@ FC_SHMEM_PUTMEM(void *target, void *source, fortran_integer_t *len, fortran_inte
...
 
+#define FC_SHMEM_CHARACTER_PUT_NBI FC_FUNC_(shmem_character_put_nbi, SHMEM_CHARACTER_PUT_NBI)
+void FC_SHMEM_CHARACTER_PUT_NBI(void *dest, void *source, fortran_integer_t *nelems, fortran_integer_t *pe);
+void
+FC_SHMEM_CHARACTER_PUT_NBI(void *dest, void *source, fortran_integer_t *nelems, fortran_integer_t *pe)
+{
+    SHMEM_ERR_CHECK_INITIALIZED();
+
+    shmem_internal_put_nb(dest, source, SIZEOF_FORTRAN_CHARACTER * *nelems, *pe, NULL);
+}
```
jdinan: Is there another argument to the C function, because Fortran strings are awful?

src/malloc.c
```diff
@@ -592,7 +592,7 @@ MAX_RELEASE_CHECK_RATE   default: 4095 unless not HAVE_MMAP
 #define FOOTERS 0
 #endif  /* FOOTERS */
 #ifndef ABORT
-#define ABORT  abort()
+#define ABORT  shmem_runtime_abort(1, "OpenSHMEM exited in error")
```
jdinan: Update string to "Sandia OpenSHMEM memory allocation error".  Use name from config.h.

src/shmem_internal.h
```diff
@@ -54,12 +54,74 @@ extern int shmem_internal_thread_level;
...
+#define SHMEM_ERR_CHECK_ACTIVE_SET(PE_start, logPE_stride, PE_size)                                     \
+    do {                                                                                                \
+        int shmem_err_check_active_stride = 1 << logPE_stride;                                          \
+        if (PE_start < 0 || logPE_stride < 0 || PE_size < 0 ||                                          \
+            PE_start + (PE_size - 1) * shmem_err_check_active_stride > shmem_internal_num_pes) {        \
+            fprintf(stderr, "ERROR: %s(): Invalid active set (PE_start = %d, logPE_stride = %d, PE_size = %d)\n", \
+                    __func__, PE_start, logPE_stride, PE_size);                                         \
+            shmem_runtime_abort(100, "OpenSHMEM exited in error");                                      \
+        }                                                                                               \
+        if (! (shmem_internal_my_pe >= PE_start &&                                                      \
```
jdinan: else if?

src/transport_portals4.h
```diff
@@ -451,29 +458,42 @@ shmem_transport_portals4_put_nb_internal(void *target, const void *source, size_
...
+             * exposed to other threads until the PtlPut. */
+            (*(long_frag->completion))++;
+            long_frag->reference++;
+
+        } else {
+            md = shmem_transport_portals4_put_cntr_md_h;
+            long_frag = NULL;
+        }
+
+        ret = PtlPut(md,
```
jdinan: AR: Jim, review fragmentation code and put counter

test/unit/alltoall.c
```diff
@@ -0,0 +1,115 @@
...
+    for (i = 0; i < npes; i++) {
+        int expected;
+
+        if (is_active(me, pe_start, pe_stride, pe_size))
+            expected = pe_group_to_world(i, pe_start, pe_stride, pe_size);
+        else
+            expected = -1;
+
+        if (out[i] != expected)
+            printf("[%d] out[%d] = %d, expected %d\n", me, i, out[i], expected);
```
jdinan: FIXME: Return error code if the test case fails.

test/unit/alltoall.c
```diff
@@ -0,0 +1,115 @@
...
+        /* Remove PE n-1 */
+        alltoall_test(out, in, 0, 0, npes-1);
+        /* Remove PE 0 */
+        alltoall_test(out, in, 1, 0, npes-1);
+        /* Only odd PEs */
+        alltoall_test(out, in, 1, 1, npes / 2);
+    }
+
+    shmem_finalize();
+    return 0;
```
jdinan: Return nonzero when there are errors.

test/unit/alltoalls.c
```diff
@@ -0,0 +1,148 @@
...
+        for (j = 0; j < nelem; j++) {
+            for (k = 0; k < dst; k++) {
+                int idx = i*dst*nelem + j*dst + k;
+                if (is_active(me, pe_start, pe_stride, pe_size))
+                    expected = (k % dst == 0) ? pe_group_to_world(i, pe_start, pe_stride, pe_size) : -1;
+                else
+                    expected = -1;
+
+                if (out[idx] != expected)
+                    printf("[%d] out[%d] = %d, expected %d\n", me, idx, out[idx], expected);
```
jdinan: Fixme, track errors.

test/unit/alltoalls.c
```diff
@@ -0,0 +1,148 @@
...
+        /* Only odd PEs */
+        alltoalls_test(out, in, 1, 1, 1, 1, 1, npes / 2); /* Same as alltoall */
+        alltoalls_test(out, in, 1, 1, 0, 1, 1, npes / 2); /* No op */
+        alltoalls_test(out, in, 2, 2, NELEM/2, 1, 1, npes / 2); /* Alternate elements */
+        alltoalls_test(out, in, 2, 1, NELEM/2, 1, 1, npes / 2); /* dst != sst */
+        alltoalls_test(out, in, 1, 2, NELEM/2 - 1, 1, 1, npes / 2); /* dst != sst */
+    }
+
+    shmem_finalize();
+    return 0;
```
jdinan: Return nonzero when errors.

test/unit/get_nbi.c
```diff
@@ -0,0 +1,74 @@
...
+{
+    int i, j, num_pes;;
+
+    shmem_init();
+
+    if (shmem_my_pe() == 0) {
+        num_pes=shmem_n_pes();
+
+        for(j = 0;j < num_pes; j++) {
+            memset(target, 0, sizeof(target));
```
jdinan: Is sizeof(target) wrong?  I think sizeof just gets us the length of the array.

test/unit/get_nbi.c
```diff
@@ -0,0 +1,74 @@
...
+            /* get 10 elements from source on remote PEs */
+            shmem_long_get_nbi(target, source, 10, j);
+            shmem_quiet();
+
+            if (0 != memcmp(source, target, sizeof(long) * 10)) {
+                fprintf(stderr,"[%d] Src & Target mismatch?\n",shmem_my_pe());
```
jdinan: Clean up error message.

test/unit/get_nbi.c
```diff
@@ -0,0 +1,74 @@
...
+    if (shmem_my_pe() == 0) {
+        num_pes=shmem_n_pes();
+
+        for(j = 0;j < num_pes; j++) {
+            memset(target, 0, sizeof(target));
+            /* get 10 elements from source on remote PEs */
+            shmem_long_get_nbi(target, source, 10, j);
+            shmem_quiet();
+
+            if (0 != memcmp(source, target, sizeof(long) * 10)) {
```
jdinan: Prefer not to use memcmp.

test/unit/put_nbi.c
```diff
@@ -0,0 +1,85 @@
...
+    memset(target, 0, sizeof(*target)*10);
+
+    shmem_barrier_all();
+
+    if (shmem_my_pe() == 0) {
+        /* put 10 elements into target on remote PEs */
+        for(i = 0; i < num_pes; i++) {
+            shmem_long_put_nbi(target, source, 10, i);
+            shmem_fence();
+            shmem_int_inc(&flag[0], i);
```
jdinan: SRSLY

test/unit/put_nbi.c
```diff
@@ -0,0 +1,85 @@
...
+
+    shmem_init();
+
+    target = (long*) shmem_malloc(sizeof(long) * 10);
+    flag = (int*) shmem_malloc(sizeof(int));
+    flag[0] = 0;
+
+    num_pes=shmem_n_pes();
+
+    memset(target, 0, sizeof(*target)*10);
```
jdinan: (*target)*10 :puke:

test/unit/put_nbi.c
```diff
@@ -0,0 +1,85 @@
...
+        for(i = 0; i < num_pes; i++) {
+            shmem_long_put_nbi(target, source, 10, i);
+            shmem_fence();
+            shmem_int_inc(&flag[0], i);
+        }
+    }
+
+    shmem_int_wait_until(&flag[0], SHMEM_CMP_EQ, 1);
+
+    if (0 != memcmp(source, target, sizeof(long) * 10)) {
```
jdinan: Clean up as in other test case.

src/collectives.c
jdinan: I think this is covered in the spec, although somewhat obliquely.  (Collectives return after local completion)

src/collectives.c
jdinan: The size restriction for put_small currently comes from the transport layer (which is kind of awful).  The length of the put_small is already checked for overrun at the transport layer.

src/data_f.c
jdinan: Ugh, this may be the case.  I'm going to log an issue for it, since the issue was propagated, but not introduced in 1.3.

src/shmem_internal.h
jdinan: I prefer it as-is, since the first check validates that the active set args are valid and the second validates that the calling PE is a member of the active set.

test/unit/get_nbi.c
jdinan: Nope, I was wrong.  sizeof(target) gives us the full size of the array in bytes.

src/transport_portals4.h
jdinan: The commit that added the put_cntr is 79662e3.

jdinan: @kseager, I reviewed this commit again and it looks ok.  In Portals 4, volatile is a property of the MD.  Before this change, we had two MDs available for issuing puts: (1) volatile and (2) full event.  This is because OpenSHMEM 1.2 only had blocking puts, so we either issued volatile/inject puts (that are locally completed immediately) or we asked for a full event to set the user's completion handle.  I had to add another MD to capture counting events for nonblocking puts for scenarios where we don't want a full event back (e.g. user didn't give a completion handle).  The ```put_cntr_md_h``` MD uses the same counter as the volatile MD, so there were no changes needed to quiet.

jdinan:
@kseager, I ran the test suite using the Portals 4 IB build on our IB cluster and all tests pass.  I think we're ready to merge this branch.  What do you think?

src/transport_portals4.h
kseager: ah, that makes sense, thanks

PR #146, opened 2016-03-14
Interface Selection

jdinan:
Add automatic interface selection using Hydra hostname variable and manual interface selection using SMA_OFI_IFACE environment variable.

### DIFF
--- a/src/transport_ofi.c
+++ b/src/transport_ofi.c
@@ -6,8 +6,9 @@
  *
  */
 
-#include <errno.h>
 #include "config.h"
+
+#include <errno.h>
 #include <stdlib.h>
 #include <sys/time.h>
 #include <sys/param.h>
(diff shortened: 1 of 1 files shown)


jdinan:
@kseager, @sayantansur, @jithinjosepkl -- Review would be much appreciated.

This is intended to fix issues with interface selection when we use the sockets provider with IPoIB.  The NULL address query to fi_getinfo has been causing the sockets provider to select the ethernet network rather than IPoIB, regardless of which hostnames are used during launch.

Questions:
 * Is the interface selection method I used reasonably portable (there's a different route which uses ioctl on a socket)?
 * Will the automatic hostname selection using the MPIR_... var mess up other providers (i.e. if we pass the hostname used to launch instead of NULL)?  I think it will, and am leaning toward removing it.  An alternative would be to push this hack (Hydra sets the env var in question) down into the sockets provider similarly to how it's consumed by the netmod MPICH.

jithinjosepkl:
:+1:  You may also want to set FI_SOURCE as flags for getinfo(), since iface_addr/svc_name specify local source address.

FI_SOURCE
Indicates that the node and service parameters specify the local source address to associate with an endpoint. If specified, either the node and/or service parameter must be non-NULL.

sayantansur:
I'm somewhat concerned that the usage of getifaddrs() builds in an assumption that interface name for every provider will be the same one as the "native" provider. There is no such requirement from OFI actually, each provider gets to define its own address structure.

jdinan:
Closing.  This will be fixed in a future PR by searching the list of fabrics returned by fi_getinfo.

PR #150, opened 2016-03-21
pmi: add support for pmi2

hppritcha:
This commit adds support for using pmi2 and hence
the ability to use the Cray PMI.  Cray PMI is weird in
several ways:

1) the PMI_KVX functions only work when craypich is launched
   using hydra (only supported on single node)
2) Cray has PMI2 but doesn't install a libpmi2
3) Cray has PMI2 but doesn't define PMI2_SUCCESS.

Verified using a modified ofi-transport.c that some of the
perf test work using socket provider on cray.

More work is need to the ofi layer to get the GNI provider
to work with sandia shmem.

@kseager 
@jdinan 

### DIFF
--- a/src/runtime-pmi2.c
+++ b/src/runtime-pmi2.c
@@ -19,9 +19,14 @@
 #include <stdio.h>
 #include <string.h>
 #if defined(PMI_PORTALS4)
-#include <portals4/pmi.h>
+#error "PMI_PORTALS4 doesn't support PMI2"
 #else
-#include <pmi.h>
+#include <pmi2.h>
+#endif
+
+/* crazy cray pmi defined PMI2 functions but uses PMI_SUCCESS (0) */
+#ifndef PMI2_SUCCESS
+#define PMI2_SUCCESS 0
 #endif
 
 #include "runtime.h"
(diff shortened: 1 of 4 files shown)


jdinan:
Could you look into the build breakages that Travis is reporting?

Also, can we move the USE_PMI1/2 guards into the Makefile rather than in the runtime-pmi1/2.c files?

hppritcha:
I need to do a configury fix for the case where one is using the built-in simple pmi. Working on it.

hppritcha:
@jdinan check now

jdinan:
Thanks for the updates.  One more question -- I'm not too familiar with the OMPI build system, what was the reason for replacling orte_check_pmi.m4 with opal_...?  Was the old file just out dated?  I assume that they are functionally equivalent?

hppritcha:
it was out of date.  we switched to using opal_check_pmi.m4 as part of the pmix work.

hppritcha:
I didn't want to remove the old orte_check_pmi.m4 yet.

src/runtime-pmi2.c
```diff
@@ -0,0 +1,210 @@
...
+ * Run-time support for the built-in runtime that is part of the
+ * shared memory implementation Portals
+ */
+
+#include "config.h"
+
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+#if defined(PMI_SLURM)
```
kseager: you aren't defining this in your m4 like we were in orte...m4 and this causes build issues with the slurm+pmi2 combo

hppritcha: okay, fixing.

hppritcha:
@kseager now try on a slurm/pmi2 installation.  Thanks.

hppritcha:
Fixed a problem with setting the max allowable value len for pmi2.

hppritcha:
added @kseager 's commits to allow for explicit pmi1 configury selection as well as to fix support
for portals4 pmi.

PR #152, opened 2016-03-25
request FI_RMA_EVENT in caps

hppritcha:
Since sandia-shmem over ofi by defalt tries to
use remote counters, it needs to request FI_RMA_EVENT
caps.  According to the libfabric man pages, this secondary
capability is required in order for the app to be able to
rely on remote counters being supported by the provider.

there was also some white space cleanup here.
unfortunately it looks like some folks use tabs, others use spaces.

@jdinan 


sayantansur:
Yes, this needs to be added in to the caps. :+1:

PR #159, opened 2016-04-13
OFI: Add support for manual interface selection

jdinan:
Add support for manual interface selection via SMA_OFI_IFACE environment
variable.  OFI transport init code was refactored (query_for_fabric was
merged into transport_init) to preserve list of fabrics, so that
fi_freeinfo can be called.

### DIFF
--- a/src/transport_ofi.c
+++ b/src/transport_ofi.c
@@ -6,8 +6,9 @@
  *
  */
 
-#include <errno.h>
 #include "config.h"
+
+#include <errno.h>
 #include <stdlib.h>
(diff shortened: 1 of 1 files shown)


jdinan:
This won't do anything useful until Shantonu's sockets provider fix is merged into libfabric.

@kseager, sorry for the churn.  I had to pull the ```query_for_fabric()``` function in to save the ```fabrics``` list so that ```fi_freeinfo``` could be called safely.  As long as I was moving the code around, I cleaned up indenting and updated the error messages.

PR #161, opened 2016-04-14
Clean up and extend fabric selection code

jdinan:
* Adds support for SMA_OFI_FABRIC env var to allow users to specify the fabric name (```fabric_attr->name```).  Shell-style wildcards can be used (e.g. SMA_OFI_FABRIC="en*").
* Removes support for the SMA_OFI_SERVICE env var, which is not anticipated to be useful for configuring the fabric with SHMEM.

This change is leading libfabric by a bit, but I think it's stable enough to merge.  In the future, I anticipate also adding a similar SMA_OFI_DOMAIN option to allow users to also specify the desired fabric domain.

### DIFF
--- a/README
+++ b/README
@@ -154,10 +154,10 @@ options.
         doubling (recdbl) will fall back to ring if the PE set is not a
         power of two in size.
 
-    CMA_PUT_MAX (default: 8192)
+    SMA_CMA_PUT_MAX (default: 8192)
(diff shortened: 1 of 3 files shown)


jdinan:
@kseager, please review the new patches.  We could also add src_addr selection, but I think provider->fabric->domain should cover what users need.

kseager:
I agree, fabric->domain has enough coverage

PR #165, opened 2016-04-21
OFI: add configury for using FI_RMA_EVENT

hppritcha:
The SOS doesn't currently do the right thing
when trying to use libfabric counter objects
bound to memory registrations.  This is actually
only available if the provider supports the
FI_RMA_EVENT secondary capability.

Add a configurery option to optinally disable
the FI_RMA_EVENT request.  When configuring with

./configure --disable-fi-rma-event

the code that depends on FI_RMA_EVENT support in the
provider is ifdef'd out.

There are a few minor cleanups and yes, a bit of whitespace change.
Normally I try to avoid that.

@kseager 

### DIFF
--- a/src/shmem_synchronization.h
+++ b/src/shmem_synchronization.h
@@ -141,17 +141,37 @@ shmem_internal_fence(void)
     } while(0)
 #elif defined(USE_OFI)
 
+#if defined(ENABLE_HARD_POLLING) || defined(USE_ON_NODE_COMMS)
+
+#define SHMEM_WAIT(var, value)                                          \
+    do {                                                                \
+        while (*(var) == value) {                                       \
+            COMPILER_FENCE();                                           \
+            if (*(var) != value) break;                                 \
+        }                                                               \
+    }   while(0)
+
+#define SHMEM_WAIT_UNTIL(var, cond, value)                              \
+    do {                                                                \
+        int cmpret;                                                     \
+        do {                                                            \
+            COMP(cond, *(var), value, cmpret);                          \
+        } while (!cmpret);                                              \
+    } while(0)
+
+#else
+
 #define SHMEM_WAIT(var, value)                                          \
     do {                                                                \
         int ret;                                                        \
         uint64_t count;                                                 \
                                                                         \
         while (*(var) == value) {                                       \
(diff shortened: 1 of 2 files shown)


src/transport_ofi.c
```diff
@@ -707,10 +718,13 @@ static inline int query_for_fabric(struct fabric_info *info)
 
     fabric_attr.prov_name = info->prov_name;
 
-    hints.caps	  = FI_RMA |     /* request rma capability
+    hints.caps   = FI_RMA |     /* request rma capability
                                     implies FI_READ/WRITE FI_REMOTE_READ/WRITE */
                    FI_ATOMICS |  /* request atomics capability */
                    FI_RMA_EVENT; /* want to use remote counters */
+#ifdef EABLE_FI_RMA_EVENT
```
kseager: typo #ifdef ENABLE_FI_RMA_EVENT

hppritcha:
reworked this to reuse the enable-hard-poll configury option.  
@kseager  check now

src/shmem_synchronization.h
```diff
@@ -141,17 +141,37 @@ shmem_internal_fence(void)
     } while(0)
 #elif defined(USE_OFI)
 
+#if defined(ENABLE_HARD_POLLING) || defined(USE_ON_NODE_COMMS)
```
kseager: isn't this case already accounted for on line 82?

jdinan: +1, this passage is redundant.  The macro re-definition should have generated a compiler warning.

src/transport_ofi.c
```diff
@@ -313,12 +313,14 @@ static inline int allocate_recv_cntr_mr(void)
     cntr_attr.events   = FI_CNTR_EVENTS_COMP;
     cntr_attr.flags    = 0;
 
+#ifndef ENABLE_HARD_POLLING
```
jdinan: This should be ```!defined(ENABLE_HARD_POLLING) && !defined(USE_ON_NODE_COMMS)```

jdinan:
I added a few helpful things in #168.  That will simplify some of the preprocessor checks.

jdinan:
I pulled these changes into #168.

PR #168, opened 2016-04-22
Add received counter to transport interface

jdinan:
Add received counter get/wait API to the transport interface and push
Portal 4 and OFI calls in the wait implementation down into the
transport layer.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -190,7 +190,9 @@ CHECK_CMA(
 AM_CONDITIONAL([USE_CMA], [test "$transport_cma" = "yes"])
 
 AS_IF([test "transport_xpmem" = "yes" -o "$transport_cma" = "yes"],
-      [AC_DEFINE([USE_ON_NODE_COMMS], [1], [Define if any on-node comm transport is available])])
+      [AC_DEFINE([USE_ON_NODE_COMMS], [1], [Define if any on-node comm transport is available])
+       AC_DEFINE([ENABLE_HARD_POLLING], [1], [Enable hard polling])
+      ])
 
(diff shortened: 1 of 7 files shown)


jdinan:
I have been meaning to fix this for a long time.  :)

configure.ac
```diff
@@ -190,7 +190,9 @@ CHECK_CMA(
 AM_CONDITIONAL([USE_CMA], [test "$transport_cma" = "yes"])
 
 AS_IF([test "transport_xpmem" = "yes" -o "$transport_cma" = "yes"],
-      [AC_DEFINE([USE_ON_NODE_COMMS], [1], [Define if any on-node comm transport is available])])
+      [AC_DEFINE([USE_ON_NODE_COMMS], [1], [Define if any on-node comm transport is available])
+       AC_DEFINE([ENABLE_HARD_POLLING], [1], [Enable hard polling])
```
kseager: currently, this will then force all the wait_until calls to use the variable polling option, but don't we only want this when its a shared memory case?

hppritcha:
you didn't include all my changes.
this branch fails to get past initialization using GNI provider.

hppritcha:
I'm going to to fix things in another PR anyway.  The current SOS misses stuff that gets caught in the stricter GNI provider EP capability checking.

jdinan:
@hppritcha - There was only one patch in #165, which I integrated into this branch, so all of those changes should be here.  We needed to expand the scope a little to cover Portals, and do some adjacent refactoring.  There's also #169 that @kseager is working to get through the atomics hangups.  And we just merged #167 which eliminates the byte atomics in barrier and fcollect.  If you pull in the changes from this branch and #169, where are the blockers?

PR #169, opened 2016-04-22
OFI: added SMA_OFI_ATOMIC_CHECKS_WARN

kseager:
+environment variable option for enabling atomic limitation warnings vs. errors
+small clean-up of atomic valid output message
+initial atomic size check uses FI_SUM instead of FI_MAX since sum is more portable

### DIFF
--- a/README
+++ b/README
@@ -180,5 +180,5 @@ options.
         identifying the desired fabric domain.
 
     SMA_OFI_ATOMIC_CHECKS_WARN (default: off)
-        OFI will not abort if fabric provider doesn't support every data type x op
-        combination, instead it will print a warning.
+        If defined, OFI will not abort if fabric provider doesn't support every data
+        type x op combination, instead it will print a warning.
--- a/README
+++ b/README
@@ -178,3 +178,7 @@ options.
         Shell-style wildcards, including * and ?, are allowed.  The fi_info
         utility included with libfabric can be used for assistance with
         identifying the desired fabric domain.
+
+    SMA_OFI_ATOMIC_CHECKS_WARN (default: off)
+        OFI will not abort if fabric provider doesn't support every data type x op
+        combination, instead it will print a warning.
--- a/src/collectives_c.c
+++ b/src/collectives_c.c
@@ -480,7 +480,7 @@ shmem_longlong_min_to_all(long long *target, const long long *source,
 
     shmem_internal_op_to_all(target, source, nreduce, sizeof(long long),
                     PE_start, logPE_stride, PE_size,
-                    pWrk, pSync, SHM_INTERNAL_MIN, SHM_INTERNAL_LONG);
+                    pWrk, pSync, SHM_INTERNAL_MIN, SHM_INTERNAL_LONG_LONG);
 }
 
 
--- a/src/init.c
+++ b/src/init.c
@@ -298,9 +298,6 @@ shmem_internal_init(int tl_requested, int *tl_provided)
             printf("SMA_INFO                %s\n",
                    (NULL != shmem_util_getenv_str("INFO")) ? "Set" : "Not set");
             printf("\tIf set, print this help message at startup\n");
-            printf("SMA_OFI_ATOMIC_CHECKS_WARN  %s\n",
-                   (NULL != shmem_util_getenv_str("OFI_ATOMIC_CHECKS_WARN")) ? "Set" : "Not set");
-            printf("\tIf set, change aborting atomic checks to warnings\n");
             printf("SMA_SYMMETRIC_SIZE      %ld\n", heap_size);
             printf("\tSymmentric heap size\n");
             printf("SMA_COLL_CROSSOVER      %d\n", crossover);
--- a/src/init.c
+++ b/src/init.c
@@ -298,6 +298,9 @@ shmem_internal_init(int tl_requested, int *tl_provided)
             printf("SMA_INFO                %s\n",
                    (NULL != shmem_util_getenv_str("INFO")) ? "Set" : "Not set");
             printf("\tIf set, print this help message at startup\n");
+            printf("SMA_OFI_ATOMIC_CHECKS_WARN  %s\n",
+                   (NULL != shmem_util_getenv_str("OFI_ATOMIC_CHECKS_WARN")) ? "Set" : "Not set");
+            printf("\tIf set, change aborting atomic checks to warnings\n");
             printf("SMA_SYMMETRIC_SIZE      %ld\n", heap_size);
             printf("\tSymmentric heap size\n");
             printf("SMA_COLL_CROSSOVER      %d\n", crossover);
--- a/src/transport_ofi.c
+++ b/src/transport_ofi.c
@@ -115,10 +115,20 @@ static int SHM_BOPS[]=
   SHM_INTERNAL_BAND, SHM_INTERNAL_BOR, SHM_INTERNAL_BXOR,
 };
 
+static char * SHM_NAMED_BOPS[]=
+{
+  "Bitwise AND", "Bitwise OR", "Bitwise XOR",
+};
+
 static int SHM_OPS[]=
 {
   SHM_INTERNAL_MIN,  SHM_INTERNAL_MAX, SHM_INTERNAL_SUM,
   SHM_INTERNAL_PROD,
+};
+
+static char * SHM_NAMED_OPS[]=
+{
+  "MIN", "MAX", "SUM", "PROD",
 };
 
 int shmem_transport_have_long_double = 1;
@@ -544,15 +554,18 @@ static inline int atomic_limitations_check(void)
     /* ----------------------------------------*/
 
     int i, j, ret = 0;
+    long atomicwarn = 0;
+
+    atomicwarn = shmem_util_getenv_long("OFI_ATOMIC_CHECKS_WARN", 0, 0);
 
     init_dt_size();
 
     /*RETRIEVE atomic max size for ATOMIC_NB case */
     size_t atomic_size;
-    ret = fi_atomicvalid(shmem_transport_ofi_epfd, FI_INT64, FI_MAX,
+    ret = fi_atomicvalid(shmem_transport_ofi_epfd, FI_INT64, FI_SUM,
 			&atomic_size);
     if(ret!=0 || (atomic_size == 0)){ //not supported
-	OFI_ERRMSG("atomicvalid failed\n");
+	    OFI_ERRMSG("atomicvalid failed: cannot determine max atomic size for transport\n");
 	return ret;
     }
     shmem_transport_ofi_max_atomic_size = atomic_size * (sizeof(long));
@@ -568,8 +581,16 @@ static inline int atomic_limitations_check(void)
         ret = fi_atomicvalid(shmem_transport_ofi_epfd, SHM_DT_INT[i], SHM_BOPS[j],
                         &atomic_size);
         if(ret!=0 || atomic_size == 0) {
-           OFI_ERRMSG("ret=%d atomic_size=%d %d %d\n", ret, (int)atomic_size, i, j);
-	   return ret;
+            fprintf(stderr, "Atomic Support: data size %d not "\
+                    "supported with %s ", SHMEM_Dtsize[SHM_DT_INT[i]],
+                    SHM_NAMED_BOPS[j]);
+            if(atomicwarn) {
+                fprintf(stderr, "WARNING\n");
+            } else {
+                OFI_ERRMSG("Error: atomicvalid ret=%d atomic_size=%d \n",
+                            ret, (int)atomic_size);
+	            return ret;
+            }
         }
       }
     }
@@ -580,8 +601,16 @@ static inline int atomic_limitations_check(void)
         ret = fi_atomicvalid(shmem_transport_ofi_epfd, SHM_DT_CMP[i], SHM_OPS[j],
                         &atomic_size);
         if(ret!=0 || atomic_size == 0) {
-           OFI_ERRMSG("ret=%d atomic_size=%d %d %d\n", ret, (int)atomic_size, i, j);
-	   return ret;
+            fprintf(stderr, "Atomic Support: data size %d not "\
+                    "supported with %s \n", SHMEM_Dtsize[SHM_DT_CMP[i]],
+                    SHM_NAMED_OPS[j]);
+            if(atomicwarn) {
+                fprintf(stderr, "WARNING\n");
+            } else {
+                OFI_ERRMSG("Error: atomicvalid ret=%d atomic_size=%d \n",
+                            ret, (int)atomic_size);
+	            return ret;
+            }
         }
       }
     }
--- a/src/transport_ofi.c
+++ b/src/transport_ofi.c
@@ -99,6 +99,8 @@ static inline void init_dt_size(void)
   SHMEM_Dtsize[FI_LONG_DOUBLE_COMPLEX] = sizeof(long double complex);
 }
 
+#define SIZEOF_DT_INT 4
+
 static int SHM_DT_INT[]=
 {
   SHM_INTERNAL_INT, SHM_INTERNAL_LONG, SHM_INTERNAL_LONG_LONG, SHM_INTERNAL_SHORT
@@ -108,6 +110,8 @@ static char * SHM_NAMED_DT_INT[]=
 {
   "Int", "Long", "Long Long", "Short"
 };
+
+#define SIZEOF_DT_CMP 6
 
 static int SHM_DT_CMP[]=
 {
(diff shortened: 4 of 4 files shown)


src/transport_ofi.c
```diff
@@ -534,15 +544,18 @@ static inline int atomic_limitations_check(void)
     /* ----------------------------------------*/
 
     int i, ret = 0;
+    long atomicwarn = 0;
+
+    atomicwarn = shmem_util_getenv_long("OFI_ATOMIC_CHECKS_WARN", 0, 0);
```
jdinan: ```shmem_util_getenv_long``` isn't quite right here.  You could either say that this is enabled when the env var is defined, or we could add a ```shmem_util_getenv_bool``` function.

README
```diff
@@ -178,3 +178,7 @@ options.
         Shell-style wildcards, including * and ?, are allowed.  The fi_info
         utility included with libfabric can be used for assistance with
         identifying the desired fabric domain.
+
+    SMA_OFI_ATOMIC_CHECKS_WARN (default: off)
+        OFI will not abort if fabric provider doesn't support every data type x op
+        combination, instead it will print a warning.
```
jdinan: Need to indicate how this option is enabled -- e.g.

```SMA_INFO: if defined, print (stdout) SHMEM environment variables.```

File: src/transport_ofi.c [hunk 3]
```diff
@@ -559,8 +572,16 @@ static inline int atomic_limitations_check(void)
         ret = fi_atomicvalid(shmem_transport_ofi_epfd, SHM_DT_INT[i], SHM_BOPS[j],
                         &atomic_size);
         if(ret!=0 || atomic_size == 0) {
-           OFI_ERRMSG("ret=%d atomic_size=%d %d %d\n", ret, (int)atomic_size, i, j);
-	   return ret;
+            fprintf(stderr, "Atomic Support: data size %d not "\
+                    "supported with %s ", SHMEM_Dtsize[SHM_DT_INT[i]],
```
jdinan: Why Dtsize and not a string with the name of the datatype?  With only the size, one can't distinguish between long and double..

src/transport_ofi.c
```diff
@@ -559,8 +572,16 @@ static inline int atomic_limitations_check(void)
         ret = fi_atomicvalid(shmem_transport_ofi_epfd, SHM_DT_INT[i], SHM_BOPS[j],
                         &atomic_size);
         if(ret!=0 || atomic_size == 0) {
-           OFI_ERRMSG("ret=%d atomic_size=%d %d %d\n", ret, (int)atomic_size, i, j);
-	   return ret;
+            fprintf(stderr, "Atomic Support: data size %d not "\
```
jdinan: The ```\``` at the end of this line is not necessary.

src/transport_ofi.c
```diff
@@ -559,8 +572,16 @@ static inline int atomic_limitations_check(void)
...
+                    "supported with %s ", SHMEM_Dtsize[SHM_DT_INT[i]],
+                    SHM_NAMED_BOPS[j]);
+            if(atomicwarn) {
+                fprintf(stderr, "WARNING\n");
```
jdinan: The output needs a little work:

>  Atomic Support: data size 8 not supported with SUM
>  WARNING

Should be more like:

> Warning: OFI detected no support for atomic 'MAX' on type 'double'

You can create a format string to handle both cases like this:

```"%s: OFI detected no support...", atomicward ? "Warning" : "Error", ...```

And then abort when ```atomicwarn``` is not set.

File: src/transport_ofi.c (hunk 3 above)
Comment (kseager) [inline src/transport_ofi.c]: true, I'll change it

File: src/init.c [hunk 6]
```diff
@@ -298,6 +298,9 @@ shmem_internal_init(int tl_requested, int *tl_provided)
             printf("SMA_INFO                %s\n",
                    (NULL != shmem_util_getenv_str("INFO")) ? "Set" : "Not set");
             printf("\tIf set, print this help message at startup\n");
+            printf("SMA_OFI_ATOMIC_CHECKS_WARN  %s\n",
```
jdinan: I don't think we want this at the top level...  The way it's written right now, the OFI transport info would also be printed out in the Portals build, even though the env var has no effect.  I think it would be better to have a transport layer function that gets called during initialization to print out an info message.

jdinan: In the interest of closing out this PR, why don't we file an issue to add an info function to the transport interface and drop this for now?

src/transport_ofi.c
```diff
@@ -546,6 +548,24 @@ static int populate_mr_tables(void)
     return 0;
 }
 
+static inline int atomicvalid_rtncheck(int ret, int atomic_size, long atomicwarn,
+                                    char strOP[], char strDT[]) {
+
+    if(ret != 0 || atomic_size == 0) {
+        fprintf(stderr, "%s OFI detected no support for atomic '%s'"
+               "on type '%s'\n", (atomicwarn ? "Warning" : "Error"),
+                strOP, strDT);
```
jdinan: Is it right to return ret in this case?  Should we return success?

src/transport_ofi.c
```diff
@@ -99,24 +99,46 @@ static inline void init_dt_size(void)
...
-  SHM_INTERNAL_MIN,  SHM_INTERNAL_MAX, SHM_INTERNAL_SUM,
-  SHM_INTERNAL_PROD,
+  SHM_INTERNAL_MIN,  SHM_INTERNAL_MAX, SHM_INTERNAL_PROD,
+  SHM_INTERNAL_SUM, FI_ATOMIC_WRITE, FI_ATOMIC_READ, FI_CSWAP,
+  FI_MSWAP
+};
+
+static char * SHM_NAMED_OPS[]=
+{
+  "MIN", "MAX", "PROD", "SUM", "SUM", "WRITE", "READ", "CSWAP",
```
jdinan: "SUM" appears twice.

src/transport_ofi.c
kseager: this function just passes forward the predetermined ret value, if it comes in success it will return success

src/transport_ofi.c
```diff
@@ -578,48 +598,65 @@ static inline int atomic_limitations_check(void)
 
     int j;
     /* Binary OPS check */
-    for(i=0; i<3; i++) {//DT
+    for(i=0; i<4; i++) {//DT
```
jdinan: Could you pull out this constant and #define it right next to the array declaration?

src/transport_ofi.c
```diff
@@ -578,48 +598,65 @@ static inline int atomic_limitations_check(void)
...
-        }
+         if(atomicvalid_rtncheck(ret, atomic_size, atomicwarn,
+                            SHM_NAMED_BOPS[j], SHM_NAMED_DT_INT[i]))
+           return ret;
       }
     }
 
     /* OTHER OPS check */
-    for(i=0; i<5; i++) {//DT
+    for(i=0; i<6; i++) {//DT
```
jdinan: Pull out this constant, too.

src/transport_ofi.c
```diff
@@ -578,48 +598,65 @@ static inline int atomic_limitations_check(void)
...
     }
 
     /* LONG DOUBLE limitation is common */
     for(j=0; j<4; j++) { //OPS
-        ret = fi_atomicvalid(shmem_transport_ofi_epfd, SHM_INTERNAL_LONG_DOUBLE, SHM_OPS[j], &atomic_size);
-        if(ret!=0 || atomic_size == 0) {
-		shmem_transport_have_long_double = 0;
+      ret = fi_atomicvalid(shmem_transport_ofi_epfd, SHM_INTERNAL_LONG_DOUBLE, SHM_OPS[j],
+                &atomic_size);
+      if(ret!=0 || atomic_size == 0) {
```
jdinan: Does OFI require the ```||``` here?  It seems like we should have an else case that asserts ```atomic_size > 0```, or better yet ```atomic_size == sizeof(long double)```?

src/transport_ofi.c
kseager: sure, I sort of prefer a sizeof calculation in the limitations check but either is fine

src/transport_ofi.c
kseager: so count = provider atomic size, but since I've got that value at this point I can do the assert, and agreed that should be added

kseager:
@jdinan I did the rebase onto master, resolved conflicts, and added in edits from your PR review, please review and merge when ready

src/collectives_c.c
```diff
@@ -480,7 +480,7 @@ shmem_longlong_min_to_all(long long *target, const long long *source,
 
     shmem_internal_op_to_all(target, source, nreduce, sizeof(long long),
                     PE_start, logPE_stride, PE_size,
-                    pWrk, pSync, SHM_INTERNAL_MIN, SHM_INTERNAL_LONG);
+                    pWrk, pSync, SHM_INTERNAL_MIN, SHM_INTERNAL_LONG_LONG);
```
jdinan: Good catch -- we must have a gap here in testing coverage...

src/transport_ofi.c
```diff
@@ -544,16 +592,20 @@ static inline int atomic_limitations_check(void)
...
-    ret = fi_atomicvalid(shmem_transport_ofi_epfd, FI_INT64, FI_MAX,
+    ret = fi_atomicvalid(shmem_transport_ofi_epfd, FI_INT64, FI_SUM,
 			&atomic_size);
     if(ret!=0 || (atomic_size == 0)){ //not supported
-	OFI_ERRMSG("atomicvalid failed\n");
-	return ret;
+	    OFI_ERRMSG("atomicvalid failed: cannot determine max atomic size for transport\n");
+	    return ret;
     }
     shmem_transport_ofi_max_atomic_size = atomic_size * (sizeof(long));
```
jdinan: Should this be ```sizeof(int64_t)``` since that's what is checked in the ```fi_atomicvalid``` call above?

src/transport_ofi.c
```diff
@@ -563,36 +615,71 @@ static inline int atomic_limitations_check(void)
...
-           OFI_ERRMSG("ret=%d atomic_size=%d %d %d\n", ret, (int)atomic_size, i, j);
-	   return ret;
-        }
+         if(atomicvalid_rtncheck(ret, atomic_size, atomicwarn,
+                            SHM_NAMED_OPS[j], SHM_NAMED_DT_CMP[i]))
+           return ret;
+      }
+    }
+
+    for(i=0; i<3; i++) {//DT does not include short
```
jdinan: Pull out this constant, too, please.  I would prefer a separate constant (rather than arithmetic on the existing constant) that is well documented to capture the convention in how the array is ordered.

src/transport_ofi.c
```diff
@@ -563,36 +615,71 @@ static inline int atomic_limitations_check(void)
     }
 
     /* Binary OPS check */
-    for(i=0; i<3; i++) {//DT
-      for(j=0; j<3; j++) { //OPS
+    for(i=0; i<SIZEOF_DT_INT; i++) {//DT
+      for(j=0; j<SIZEOF_BOPS; j++) { //OPS
```
jdinan: I know this is obnoxious feedback, but these comments are not very helpful.  I think this is a helpful spot for comments, these are just too brief.  Could you expand to something like ```/* Loop over bitwise ops and supported datatypes */```?

src/transport_ofi.c
```diff
@@ -563,36 +615,71 @@ static inline int atomic_limitations_check(void)
...
+    }
+
+    ret = fi_compare_atomicvalid(shmem_transport_ofi_epfd, SHM_INTERNAL_INT, FI_MSWAP,
+                        &atomic_size);
+     if(atomicvalid_rtncheck(ret, atomic_size, atomicwarn,
+                        "FI_MSWAP", "Int"))
+       return ret;
+
+    for(i=0; i<5; i++) {//DT
+      for(j=4; j<6; j++) { //OPS
```
jdinan: Constants...

src/transport_ofi.c
```diff
@@ -563,36 +615,71 @@ static inline int atomic_limitations_check(void)
...
+        ret = fi_fetch_atomicvalid(shmem_transport_ofi_epfd, SHM_DT_CMP[i], SHM_OPS[j],
+                        &atomic_size);
+        if(atomicvalid_rtncheck(ret, atomic_size, atomicwarn,
+                            SHM_NAMED_OPS[j], SHM_NAMED_DT_CMP[i]))
+          return ret;
       }
     }
 
     /* LONG DOUBLE limitation is common */
     for(j=0; j<4; j++) { //OPS
```
jdinan: And, another constant..

src/transport_ofi.c
jdinan: Was there an assert added?

jdinan:
Could we create separate tables for each set of checks?  There would be some repetition across the tables, but it would simplify the checking code -- a simple loop over the tables without any fiddling.  Someting like:
```
/* Cover OpenSHMEM atomics API */
DT_AMO_STANDARD = { int, long, long long, int32_t, int64_t } /* int32/64_t for Fortran */
AMO_STANDARD_OPS = { SUM }
FETCH_AMO_STANDARD_OPS = { CSWAP, SUM }

DT_AMO_EXTENDED = { float, double, int, long, long long, int32_t, int64_t, FORTRAN_INTEGER }
AMO_EXTENDED_OPS = { WRITE }
FETCH_AMO_EXTENDED_OPS = { SWAP, READ }

/* Cover one-sided implementation of reduction */
DT_REDUCE_BITWISE = { short, int, long, long long, int32_t, int64_t }
REDUCE_BITWISE_OPS = { AND, OR, XOR }

DT_REDUCE_COMPARE = { float, double, long double, short, int, long, long long, int32_t, int64_t }
REDUCE_COMPARE_OPS = { MAX, MIN }

DT_REDUCE_ARITH = { float, double, long double, float complex, double complex, short, int, long, long long, int32_t, int64_t } /* FIXME: Fortran COMP4 and COMP8 are not yet implemented */
REDUCE_ARITH_OPS = { SUM, PROD }
```

kseager:
I like this suggestion and it works except for the following tweaks:

1)long double case...I'm going to have to pull that out..just as an FYI 
2) I'll add in MSWAP and separate out 
3) standards ops must have SUM and CSWAP separate because they require separate FI atomicvalid checks (fetch vs. compare)

src/transport_ofi.c
```diff
@@ -99,28 +99,124 @@ static inline void init_dt_size(void)
   SHMEM_Dtsize[FI_LONG_DOUBLE_COMPLEX] = sizeof(long double complex);
 }
 
-static int SHM_DT_INT[]=
+static char * SHMEM_DtName[FI_DATATYPE_LAST]=
```
jdinan: Are ```FI_DATATYPE_LAST``` and ```FI_ATOMIC_OP_LAST``` part of the public libfabric API?  I don't see them in the manpages.

src/transport_ofi.c
```diff
@@ -99,28 +99,124 @@ static inline void init_dt_size(void)
...
 {
-  SHM_INTERNAL_SHORT, SHM_INTERNAL_INT, SHM_INTERNAL_LONG,
+  "int8", "uint8", "int16", "uint16", "int32", "uint32", "int64", "uint64"
+  "float", "double", "float complex", "double complex", "long double",
+  "long double complex"
```
jdinan: Need to be sure that we can rely on the values of the FI_* type and op constants landing on these array indices.

src/transport_ofi.c
```diff
@@ -536,63 +632,192 @@ static int populate_mr_tables(void)
...
+
+    /* LONG DOUBLE limitation is common */
+    for(j=0; j<SIZEOF_REDC_OPS; j++) { //OPS
+      ret = fi_atomicvalid(shmem_transport_ofi_epfd, SHM_INTERNAL_LONG_DOUBLE,
+                REDUCE_COMPARE_OPS[j], &atomic_size);
+      if(ret!=0 || atomic_size == 0) {
+	    shmem_transport_have_long_double = 0;
+		break;
+	  } else if((atomic_size*sizeof(long double)) !=
+                                    shmem_transport_ofi_max_atomic_size) {
```
jdinan: Is this check new?  I'm not sure I understand why the long double support would have to be equal to (e.g. not less than) ```max_atomic_size```?

src/transport_ofi.c
kseager: so I checked up on this, these assumptions are encoded in the version number you get when you call getinfo and its part of the ABI compatibility.

src/transport_ofi.c
kseager: so max atomic size is a pre-calculated size we just got from the fabric a few atomicvalid calls before. Sizeof(long double) should be the same between SHMEM and OFI provider...and thus that is really all this is checking as a sanity check that you had asked for earlier..

src/transport_ofi.c
jdinan: This is true for how we use the constants in building Dtsize, but with
Dtname you are assuming a specific value.

PR #170, opened 2016-04-22
Add SHM_INTERNAL types for standard integer types

jdinan:
Replace direct usage of DTYPE_ configure macros with SHM_INTERNAL_ for
the standard integer types (int, short, long, etc.).


jdinan:
Another thing I have been wanting to fix for a long time.  :)

PR #175, opened 2016-04-26
OFI: enable USE_ON_NODE_COMMS

kseager:
+added optional hostname exchange

### DIFF
--- a/src/transport_ofi.c
+++ b/src/transport_ofi.c
@@ -591,9 +591,16 @@ static inline int atomic_limitations_check(void)
 
 static inline int exchange_and_av_insert(struct fabric_info *info)
 {
-    int                 i, rank = 0, ret = 0;
-    char   epname[128], *alladdrs = NULL;
-    size_t epnamelen = sizeof(epname);
+    int    i, rank = 0, ret = 0;
+    size_t epnamelen = 128, full_ep_len = epnamelen + HOST_NAME_MAX, adrlen = 0;
+    char   epname[full_ep_len], *alladdrs = NULL, *epnamebuff = NULL;
+#ifdef USE_ON_NODE_COMMS
+    int num_on_node = 0
+    size_t pidlen = HOST_NAME_MAX;
+    full_ep_len = pidlen;
+#else
+    full_ep_len = 0; /* full_ep_len = epnamelen */
+#endif
 
     rank = shmem_runtime_get_rank();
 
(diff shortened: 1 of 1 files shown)


src/transport_ofi.c
```diff
@@ -591,9 +591,16 @@ static inline int atomic_limitations_check(void)
...
-    int                 i, rank = 0, ret = 0;
-    char   epname[128], *alladdrs = NULL;
-    size_t epnamelen = sizeof(epname);
+    int    i, rank = 0, ret = 0;
+    size_t epnamelen = 128, full_ep_len = epnamelen + HOST_NAME_MAX, adrlen = 0;
+    char   epname[full_ep_len], *alladdrs = NULL, *epnamebuff = NULL;
+#ifdef USE_ON_NODE_COMMS
+    int num_on_node = 0
+    size_t pidlen = HOST_NAME_MAX;
+    full_ep_len = pidlen;
```
jdinan: ```full_ep_len``` is used above to set the size of the ```epname``` array and then the value is changed here.  This seems weird...

src/transport_ofi.c
```diff
@@ -603,24 +610,44 @@ static inline int exchange_and_av_insert(struct fabric_info *info)
 	return ret;
     }
 
+    full_ep_len = full_ep_len + epnamelen;
+    epnamebuff = malloc(full_ep_len);
```
jdinan: Check for NULL.

src/transport_ofi.c
```diff
@@ -591,9 +591,16 @@ static inline int atomic_limitations_check(void)
 
 static inline int exchange_and_av_insert(struct fabric_info *info)
 {
-    int                 i, rank = 0, ret = 0;
-    char   epname[128], *alladdrs = NULL;
-    size_t epnamelen = sizeof(epname);
+    int    i, rank = 0, ret = 0;
+    size_t epnamelen = 128, full_ep_len = epnamelen + HOST_NAME_MAX, adrlen = 0;
```
jdinan: Is ```HOST_NAME_MAX``` portable?  It looks like you may need to add 1 for the NULL byte.

MacOS doesn't define ```HOST_NAME_MAX```.  The manpage says you are supposed to use ```_POSIX_HOST_NAME_MAX``` +1 for NULL instead.

src/transport_ofi.c
```diff
@@ -603,24 +610,44 @@ static inline int exchange_and_av_insert(struct fabric_info *info)
...
 	OFI_ERRMSG("alladdrs is NULL\n");
 	return ret;
     }
 
-    ret = shmem_runtime_put("OFI", epname, epnamelen);
+#ifdef USE_ON_NODE_COMMS
+    gethostname(&epname[epnamelen], pidlen);
+#endif
+
+    ret = shmem_runtime_put("OFI", epname, full_ep_len);
```
jdinan: Can we use a more specific string here, to avoid a potential future clash?  For example, "fi_hostname", following the convention of other OFI entries in the PMI KVS.

src/transport_ofi.c
```diff
@@ -603,24 +610,44 @@ static inline int exchange_and_av_insert(struct fabric_info *info)
...
     shmem_runtime_barrier();
 
     for(i=0; i<info->npes; i++)
     {
-	    void *tgt = alladdrs + i*epnamelen;
+	    adrlen = i*epnamelen;
+
 	    if (i == rank) {
-		    memcpy(tgt, epname, epnamelen);
+		    memcpy(epnamebuff, epname, full_ep_len);
```
jdinan: Can this overflow or walk off end end of epname?  On line 613, full_ep_len is increased beyond the size of epname, and this is memcpy, not strncpy, so it will wander past the NULL in epname.

src/transport_ofi.c
```diff
@@ -603,24 +610,44 @@ static inline int exchange_and_av_insert(struct fabric_info *info)
 	return ret;
     }
 
+    full_ep_len = full_ep_len + epnamelen;
+    epnamebuff = malloc(full_ep_len);
+
     alladdrs = malloc(info->npes * epnamelen);
```
jdinan: epnamelen and note full_ep_len?

src/transport_ofi.c
kseager: actually epname is either as large or larger than full_ep_len. At first full_ep_len is of a larger than needed size including the predicted size of hostname. After fi_getname, this readjusts the epnamelen to a likely much smaller value. For instances in sockets its around ~16. At this point epnamebuff is created with the adjusted epnamelen size + hostname_max size. epname is still at its original size so by the time you hit this memcpy you have full_ep_len = sizeof(epnamebuff) which is >=  sizeof(epname)

src/transport_ofi.c
kseager: yes, because the alladdrs array is just holding the first half of the epname value, which is the provider address used for avinsert. Essentially my method here is to piggy back on the av_insert PMI exchange so I've tagged on the hostname to the end of that array. I readjust the epname array based on whether or not you are using on_node_comms

I'm passing around epname which looks like this:

|___provider_address _from_getname__|____hostname____|
|--------epnamelen, updated by getname-|-HOSTNAME_MAX-|

src/transport_ofi.c
kseager: will do!

src/transport_ofi.c
kseager: yeah I agree, I'll fix this so that the initial epname size is saved in a different variable

src/transport_ofi.c
kseager: agreed

src/transport_ofi.c
kseager: sure

kseager:
closing: its easier to branch off master and put on changes than rebase given edits needed from feedback; will submit new PR

PR #177, opened 2016-04-27
Split AV setup into publish/populate

jdinan:
Please review -- @kseager and @sayantansur

### DIFF
--- a/src/transport_ofi.c
+++ b/src/transport_ofi.c
@@ -598,13 +598,11 @@ static inline int atomic_limitations_check(void)
     return 0;
 }
 
-static inline int exchange_and_av_insert(struct fabric_info *info)
+static inline int publish_av_info(struct fabric_info *info)
 {
-    int                 i, rank = 0, ret = 0;
-    char   epname[128], *alladdrs = NULL;
+    int    ret = 0;
+    char   epname[128];
     size_t epnamelen = sizeof(epname);
-
-    rank = shmem_runtime_get_rank();
 
     ret = fi_getname((fid_t)shmem_transport_ofi_epfd, epname, &epnamelen);
     if(ret!=0 || (epnamelen > sizeof(epname))){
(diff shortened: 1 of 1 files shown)


src/transport_ofi.c
```diff
@@ -598,54 +598,61 @@ static inline int atomic_limitations_check(void)
...
+    alladdrs = malloc(shmem_internal_num_pes * max_epnamelen);
+    if (alladdrs == NULL) {
+        OFI_ERRMSG("Out of memory allocating 'alladdrs'\n");
+        return ret;
+    }
+
+    for (i = 0, addr_ptr = alladdrs; i < shmem_internal_num_pes; i++) {
+            size_t epnamelen;
+            shmem_runtime_get(i, "fi_epnamelen", &epnamelen, sizeof(size_t));
+            shmem_runtime_get(i, "fi_epname", addr_ptr, epnamelen);
```
kseager: why exchange the epnamelen? it should be standardized per fabric selection and thus homogeneous across PE's

src/transport_ofi.c
```diff
@@ -598,52 +598,59 @@ static inline int atomic_limitations_check(void)
...
```
sayantansur: As we discussed in the chat, you only need the epnamelen from PMI if you're going to check that you're not in a heterogeneous fabric. Otherwise, if you assume that the user has chosen the same fabric, same provider on all PEs, you don't need to get this from every PE. You can just assume that all PE name length == local PE name length.

If you are getting the name length from every PE, then at least we can check if it is the same as local name length as a caution - although the best thing to exchange and check would be info->addr_format, since you could have two PEs (in an heterogenous case) with same sized address, but different addr_format.

My suggestion is to be lazy assume epnamelen is same across the job, and its the users job to ensure that they don't break that assumption.

src/transport_ofi.c
jdinan: Looking at address formats like FI_SOCKADDR, it looks like you could also have heterogeneous addresses with both the same addr_format and the same addrlen...

jdinan:
Added 8c23195, which drops ```epnamelen``` exchange.

Issue #164, opened 2016-04-20
shmem_ptr with OFI transport


kseager:
### Issue

shmem_ptr is currently not supported with OFI: enable supporting PE's that use the same node

(currently supported by Portals4 transport)

### Solution Ideas

1) enable XPMEM, including filing out internal on-node PE tracking array (USE_ON_NODE_COMMS)
2) look into extending malloc heap setup to expose heap (require no-remote-va)
3) look into the capability exposure through OFI's shared memory implementation in the utility library

jdinan:
@kseager, do you want to keep this one open?

kseager:
sure we can let it hang for a bit, but label it enhancement since we have a functionality with XPMEM now

PR #180, opened 2016-05-01
OFI: enable USE_ON_NODE_COMMS

kseager:
+added optional hostname exchange
+can be utilized by XPMEM for shared memory topology detection

resolves issue #164

### DIFF
--- a/src/transport_ofi.c
+++ b/src/transport_ofi.c
@@ -79,6 +79,10 @@ size_t    			shmem_transport_ofi_max_msg_size;
 size_t    			shmem_transport_ofi_bounce_buffer_size;
 size_t    			shmem_transport_ofi_addrlen;
 fi_addr_t			*addr_table;
+#ifdef USE_ON_NODE_COMMS
+#define EPHOSTNAMELEN  _POSIX_HOST_NAME_MAX + 1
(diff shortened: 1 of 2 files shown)


src/transport_ofi.c
```diff
@@ -605,6 +609,17 @@ static inline int publish_av_info(struct fabric_info *info)
     char   epname[128];
     size_t epnamelen = sizeof(epname);
 
+#ifdef USE_ON_NODE_COMMS
+    gethostname(myephostname, (EPHOSTNAMELEN - 1));
```
jdinan: Should we check the return value from gethostname?  The manpage says:

> Upon successful completion, the value 0 is returned; otherwise the value -1 is returned and the global variable errno is set to indicate the error.

jdinan:
This is perfect -- thanks for the PR!  I had one minor question about whether you want to add another error check.

src/transport_ofi.c
kseager: agreed!

kseager:
I tested this on CORI with ppn 2 for N=2/4 and did a print to verify it was using the XPMEM pathway. Seems to be working!

jdinan:
:shipit:  😃

PR #181, opened 2016-05-02
OFI: init fix for FI_RMA_EVENT

kseager:


jdinan:
Good catch -- thanks!

PR #183, opened 2016-05-02
OFI: add error cq to cntr_ep

kseager:
+OFI requirement
+additionally using cq for cntr error handling

@sayantansur : we've already discussed, but if you could review and merge in, thanks! 

@hppritcha this looks like its fixes the unit test failures we were seeing

### DIFF
--- a/src/transport_ofi.c
+++ b/src/transport_ofi.c
@@ -151,12 +151,12 @@ static int AMO_STANDARD_OPS[]=
 {
   SHM_INTERNAL_SUM
 };
-#define SIZEOF_AMO_FOPS 2
+#define SIZEOF_AMO_FOPS 1
 static int FETCH_AMO_STANDARD_OPS[]=
 {
   SHM_INTERNAL_SUM
 };
-#define SIZEOF_AMO_COPS 2
+#define SIZEOF_AMO_COPS 1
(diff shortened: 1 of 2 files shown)


src/transport_ofi.c
```diff
@@ -348,6 +348,14 @@ static inline int bind_resources_to_and_enable_ep(void)
 	return ret;
     }
 
+    /* attach CQ for error handling on cntr EP */
+    ret = fi_ep_bind(shmem_transport_ofi_cntr_epfd,
```
sayantansur: Super tiny nit. Is this still OK to be called cntr_epfd :-)

jdinan:
👍

jdinan:
The other way to solve this is to check the error count on event counter.  This is what's done in the Portals transport, but it looks like we missed it in the OFI implementation of quiet.  Given that we have the CQ anyway, it's always nice to print an error message (all we know from the counter is that an error occurred, not what caused it).  You might want to add a comment about the event counter approach, for future performance tuning in case binding the the CQ to the EP has a performance impact for some provider.

sayantansur:
@jdinan is right. Should we be calling cq_readerr() or cntr_readerr()? I suppose that's a question for Sean.

@kseager can you follow up with Sean to see what he thinks the right behavior should be?

PR #191, opened 2016-05-10
OFI: improve nb bw

kseager:
+only force completion of outstanding puts after a nb put
+only force completion in fragmented put case

@jdinan please review


jdinan:
Great work!  :+1:

PR #193, opened 2016-05-12
Update autogen to use autoreconf

### LINKED ISSUE #99: Update autogen.sh to use autoreconf
We added a workaround in #87 to make autogen.sh work on Mac.  The proper fix is to update autogen.sh to use autoreconf instead of calling the autotools directly.

jdinan:
Update the autogen.sh script to call autoconf instead of calling the individual autotools directly.

Closes #99.


kseager:
this worked great on my system, and Travis seems happy about it, approved!

PR #194, opened 2016-05-12
Nb perf tests

kseager:
+enables stand-alone build for perf tests
+adds non-blocking tests
+fixes shmem_free bug that showed up in Cray SHMEM runs

@jdinan or @sayantansur for review

### DIFF
--- a/test/performance/shmem_perf_suite/common.h
+++ b/test/performance/shmem_perf_suite/common.h
@@ -130,8 +130,3 @@ void static inline validate_recv(char * buf, int len, int partner_pe)
                     partner_pe);
     }
 }
-
-int static inline partner_node(int my_node)
-{
-    return ((my_node % 2 == 0) ? (my_node + 1) : (my_node - 1));
-}
--- a/test/performance/shmem_perf_suite/common.h
+++ b/test/performance/shmem_perf_suite/common.h
@@ -15,6 +15,8 @@
 #include <assert.h>
 #include <stdbool.h>
 #include <sys/time.h>
+#include <stdint.h>
+
 
 #ifndef HAVE_SHMEMX_WTIME
 double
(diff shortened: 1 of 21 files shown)


test/performance/shmem_perf_suite/shmem_bibw_nb_get_perf.c
```diff
@@ -0,0 +1,74 @@
...
+
+    shmem_barrier_all();
+
+    if (metric_info->my_node % 2 == 0) {
+        for (i = 0; i < metric_info->trials + metric_info->warmup; i++) {
+            if(i == metric_info->warmup)
+                start = shmemx_wtime();
+
+            for(j = 0; j < metric_info->window_size; j++)
+                shmem_getmem_nbi(metric_info->dest, metric_info->src, len, dest);
```
sayantansur: Should we put a synchronization point after every window operations? My concern is that one PE can 'race' ahead of the other and this can lead to the test reporting a bibw that can be greater than the network max (since transactions don't necessarily hit the wire at the same time).

test/performance/shmem_perf_suite/shmem_bibw_nb_put_perf.c
```diff
@@ -0,0 +1,79 @@
...
+
+    shmem_barrier_all();
+
+    if (metric_info->my_node % 2 == 0) {
+        for (i = 0; i < metric_info->trials + metric_info->warmup; i++) {
+            if(i == metric_info->warmup)
+                start = shmemx_wtime();
+
+            for(j = 0; j < metric_info->window_size; j++)
+                shmem_putmem_nbi(metric_info->dest, metric_info->src, len, dest);
```
sayantansur: same sync comment applies here.

test/performance/shmem_perf_suite/common.h
```diff
@@ -14,6 +14,21 @@
...
+{
+    double wtime;
+    struct timeval tv;
+
+    gettimeofday(&tv, NULL);
+    wtime = tv.tv_sec;
+    wtime += (double)tv.tv_usec / 1000000.0;
+    return wtime;
+}
+#endif
```
sayantansur: I don't know what shmemx_wtime() is supposed to return. Seconds? This will lose the timer resolution. Should we be reporting ns instead?

jdinan: This returns the current time, in seconds, just like ```MPI_Wtime``.  There is a corresponding implementation of ```shmemx_wtime()``` in the SOS implementation that should be updated if we make any changes here (the code you commented on here is not actually exercised when you build this benchmark using SOS).

There is a secondary issue to the loss of precision you noted, which is that ```gettimeofday``` is a poor way to measure the passage of time (see, https://blog.habets.se/2010/09/gettimeofday-should-never-be-used-to-measure-time).  We can use the POSIX timer, ```clock_gettime``` with ```CLOCK_REALTIME``` or ```CLOCK_MONOTONIC``` when it's available and fall back to ```gettimeofday``` otherwise.

```shmemx_time``` was a quick hack from a while back.  Good timers will take some configury and system-specific hacking.  We should be conscious of scope creep..

test/performance/shmem_perf_suite/shmem_bibw_nb_get_perf.c
jdinan: Should the barrier go before the beginning of each trial?  If we are not doing it already, we could also look at the time from both PEs in the bidirectional test so we are reporting on end-to-end completion.

test/performance/shmem_perf_suite/common.h
jdinan: #195

sayantansur: OK sounds good.

test/performance/shmem_perf_suite/shmem_bibw_nb_get_perf.c
sayantansur: Don't the shmem semantics force you to look at end-to-end completion? Another option might be to do a barrier at the end of the warmups?

We can proceed for now if you're kind enough to create an issue so we don't forget to reinvestigate this on some fast network :-)

jdinan: I think we are missing a ```quiet``` here to complete those nonblocking get operations.

The end-to-end completion that I was thinking of is waiting for the gets to complete at both PEs in the bidirectional test.

sayantansur: Duh. Yea, you're right.

kseager: added in the quiet, thanks for the catch

PR #196, opened 2016-05-27
Add pi unit test

jpdoyle:
Adds a simple unit test to compute pi via a monte-carlo method. Calls srand() with 1+pe_id, so it is deterministic for any fixed PE count.


test/unit/pi.c
```diff
@@ -0,0 +1,94 @@
...
+        }
+    }
+
+    shmem_barrier_all();
+
+    if(me == 0) {
+        for(int i = 1; i < myshmem_n_pes; ++i) {
+            unsigned long long remoteInside,remoteTotal;
+            shmem_longlong_get(&remoteInside,&inside,1,1);
+            shmem_longlong_get(&remoteTotal,&total,1,1);
```
jdinan: There's a bug in one of the arguments to both get routines.

test/unit/pi.c
```diff
@@ -0,0 +1,94 @@
...
+            total += remoteTotal;
+            inside += remoteInside;
+        }
+
+        double approx_pi = 4.0*inside/(double)total;
+
+        assert(fabs(M_PI-approx_pi) < 0.1);
```
jdinan: Could you change this from assert to ```if (...) ++errors;``` and return errors below instead of 0?

jpdoyle:
Woops, I'll fix those on Monday.

jdinan:
Great work!  :+1: :shipit:

PR #207, opened 2016-06-16
atomic limitation support change

kseager:
now we detect on the fly if we need to provide reduction software support, for all other atomic limitation cases there will be a detection in init and this will cause an abort unless the atomic warn environment variable is used

@jdinan I tweaked the support from what we discussed on the phone so please review...

-I tried testing this via faking a limitation in sockets (since CORI is down)

### DIFF
--- a/src/transport_ofi.c
+++ b/src/transport_ofi.c
@@ -246,7 +246,11 @@ static int INTERNAL_REQ_OPS[]=
   FI_MSWAP
 };
 
-
+typedef enum{
+    NONE,
+    LIMITATION_WARNING,
+    SOFT_SUPPORT,
+}atomic_support_lv;
 
 
 int shmem_transport_have_long_double = 1;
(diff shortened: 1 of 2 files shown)


sayantansur:
Does this require the user to still pass in the env var to enable software reduction support? It would be desirable to not have to pass in any variable if the user program is compatible with the SHMEM standard.

src/transport_ofi.c
```diff
@@ -672,14 +674,17 @@ static int populate_mr_tables(void)
...
     if(ret != 0 || atomic_size == 0) {
-        fprintf(stderr, "%s OFI detected no support for atomic '%s' "
-               "on type '%s'\n", (atomicwarn ? "Warning" : "Error"),
-                strOP, strDT);
-        if(!atomicwarn) {
+        if(atomic_sup == LIMITATION_WARNING) {
+            fprintf(stderr, "Warning OFI detected no support for atomic '%s' "
+               "on type '%s'\n", strOP, strDT);
+        }
+        if(atomic_sup == NONE) {
```
jdinan: Defensive coding -- ```else if``` instead of ```if```?

src/transport_ofi.h
```diff
@@ -871,11 +869,15 @@ shmem_transport_fetch_atomic(void *target, const void *source, void *dest,
...
+/* detect atomic limitation on the fly and provide software reduction support
+    if needed */
 static inline
 int shmem_transport_atomic_supported(shm_internal_op_t op,
                                      shm_internal_datatype_t datatype)
 {
-    return datatype != SHM_INTERNAL_LONG_DOUBLE || shmem_transport_have_long_double;
+   size_t size = 0;
+   int ret = fi_atomicvalid(shmem_transport_ofi_epfd, datatype, op, &size);
+   return ((ret != 0 || size == 0) ? 0 : 1 );
```
jdinan: The conditional is redundant.  You can return the condition:
```return !(ret != 0 || size == 0);```

jdinan:
Quick review -- looks ok to me.  @jpdoyle, could you take a look?  The behavior we want is to get an error if any of the atomics needed by the SHMEM atomics API are missing and to silently fall back to software for the SHMEM reductions.

kseager:
@sayantansur with this model we do automatically use the software reduction support; I agree with your sentiment. @jdinan thanks for pointing those out, I was rushing and it came out a little sloppy

jdinan:
This looks ok to me now.  @kseager, if you're happy with it, please go ahead and merge.

PR #208, opened 2016-06-17
Latency perf addition

kseager:
small put/get streaming latency addition

@jdinan last thing we need on my end for 1.3


jdinan:
@sayantansur, Could you review?

sayantansur:
I was waiting to talk to @kseager on Monday to understand the changes and the context behind them.

I'm a bit confused by the name of a test (nb_put), but that doesn't use a non blocking put. I'm also a bit confused by the name 'stream' when we are using a blocking SHMEM call? I'm sure Kayla has the answers:-)

kseager:
@sayantansur I added a clean-up commit to address your concerns: please review and pull in if latest version is sufficient

sayantansur:
thanks for the name changes!

PR #209, opened 2016-06-17
Add shmem_char_{iput,iget}

jpdoyle:
This commit adds the missing functions the simple way so they can be included in the upcoming release, and includes the rma_coverage test.


jdinan:
The SOS code looks good.  @kseager, could you review Joe's test case (I need to go work on our paper)?

PR #236, opened 2016-09-03
Update Portals to use operation complete (OC) acks

jdinan:


jdinan:
@regrant, could you please review?  This cleared testing on our IB cluster.

regrant:
This is fine with me, it will work correctly. Just be aware that you're asking for a heavier weight ACK in this case.

jdinan:
The OC ACK should be lighter weight than the CT ACK.  From Sec. 3.15.1:

> The PTL_OC_ACK_REQ acknowledgment type is useful when only operation counting is required and it is known that there is a list entry at the target that will accept the message. The PTL_OC_ACK_REQ acknowledgment type may be more efficient in some implementations because the PTL_OC_ACK_REQ acknowledgment type communicates no information about the state of the target when the message arrived.

PR #237, opened 2016-09-09
Remove unused externs from OFI xport header

jdinan:


jdinan:
@kseager please review/merge.  I noticed that these externs were unused.  Removing them to trim down the set of things exported from ```transport_ofi.c```.

kseager:
thanks for the clean-up work

PR #239, opened 2016-09-13
Add support for SMA_DEBUG env var

### LINKED ISSUE #41: Add support for SMA_DEBUG env var
If SMA_DEBUG is set, the implementation should print out additional debugging information.
Presently, this is not supported.

richardnorth3:
Added support for SMA_DEBUG environment variable & shmem_internal_debug
variable.  Also added change to README.

Closes #41

Signed-off-by: Richard North III <richard.north.iii@intel.com>

### DIFF
--- a/src/init.c
+++ b/src/init.c
@@ -54,6 +54,7 @@ int shmem_internal_initialized_with_start_pes = 0;
 int shmem_internal_global_exit_called = 0;
 
 int shmem_internal_thread_level;
+int shmem_internal_debug = 0;
 
 #ifdef ENABLE_THREADS
 shmem_internal_mutex_t shmem_internal_mutex_alloc;
@@ -161,6 +162,7 @@ shmem_internal_init(int tl_requested, int *tl_provided)
     heap_size = shmem_util_getenv_long("SYMMETRIC_SIZE", 1, 512 * 1024 * 1024);
     eager_size = shmem_util_getenv_long("BOUNCE_SIZE", 1, 2048);
     heap_use_malloc = shmem_util_getenv_long("SYMMETRIC_HEAP_USE_MALLOC", 0, 0);
+
     /* huge page support only on Linux for now, default is to use 2MB large pages */
 #ifdef __linux__
     if (heap_use_malloc == 0) {
@@ -172,6 +174,11 @@ shmem_internal_init(int tl_requested, int *tl_provided)
     }
 #endif
 
+    shmem_internal_debug = (NULL != shmem_util_getenv_str("DEBUG")) ? 1 : 0;
+    if (shmem_internal_debug > 0){
+
+      DEBUG_STR("Debug info test\n");
+    }
 
     /* Find symmetric data */
 #ifdef __APPLE__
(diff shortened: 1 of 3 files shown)


jdinan:
Could you use ```shmem_util_getenv_str``` to get on/off behavior similar to info/version env vars, instead of ```shmem_util_getenv_long```?  The spec says we have to enable debug output for any value, including "foobar" and the empty string.  We would have to write a bunch of code to handle all of these cases and that doesn't seem like a good use of time at the moment.

jdinan:
Could you also add a ```DEBUG_STR``` macro to ```shmem_internal.h```, similar to the ```RAISE_WARN_STR``` macro?

jdinan:
It looks like there are change markers in your latest patch, which is breaking the build.  Let me know if you want any help sorting this out.

src/init.c
```diff
@@ -172,6 +174,11 @@ shmem_internal_init(int tl_requested, int *tl_provided)
     }
 #endif
 
+    shmem_internal_debug = (NULL != shmem_util_getenv_str("DEBUG")) ? 1 : 0;
+    if (shmem_internal_debug > 0){
+
+      DEBUG_STR("Debug info test\n");
```
jdinan: Did you mean to commit this test code?

src/shmem_internal.h
```diff
@@ -57,6 +58,12 @@ extern int shmem_internal_thread_level;
                 shmem_internal_my_pe, __FILE__, __LINE__, str);         \
     } while (0)
 
+#define DEBUG_STR(str)                                                  \
+    do {                                                                \
+        fprintf(stderr, "[%03d] DEBUG: %s:%d: %s\n",                    \
+                shmem_internal_debug, __FILE__, __LINE__, str);         \
```
jdinan: This should be ```shmem_internal_my_pe```, not ```shmem_internal_debug```.

src/shmem_internal.h
```diff
@@ -57,6 +58,12 @@ extern int shmem_internal_thread_level;
                 shmem_internal_my_pe, __FILE__, __LINE__, str);         \
     } while (0)
 
+#define DEBUG_STR(str)                                                  \
+    do {                                                                \
+        fprintf(stderr, "[%03d] DEBUG: %s:%d: %s\n",                    \
```
jdinan: Need ```if (shmem_internal_debug) { ...``` here.

src/init.c
```diff
@@ -345,6 +352,8 @@ shmem_internal_init(int tl_requested, int *tl_provided)
             printf("SMA_CMA_PUT_MAX         %s\n", shmem_transport_cma_put_max);
             printf("SMA_CMA_GET_MAX         %s\n", shmem_transport_cma_get_max);
 #endif /* USE_CMA */
+            printf("SMA_DEBUG               %s\n", (NULL != shmem_util_getenv_str("DEBUG")) ? "On" : "Off");
+            printf("\tValue for runtime debug information.\n");
```
jdinan: This description is a little odd.  Could you make it something like "Enable debugging messages".

PR #247, opened 2016-09-28
OFI: Drop attr changes in cntr EP creation

jdinan:
These attribute changes have no effect, thus they should be dropped.

### DIFF
--- a/src/transport_ofi.c
+++ b/src/transport_ofi.c
@@ -294,10 +294,7 @@ static inline int allocate_endpoints(struct fabric_info *info)
      * support shmem_wait that blocks on an update arriving from the network */
 #ifndef ENABLE_HARD_POLLING
     info->p_info->caps |= FI_REMOTE_WRITE | FI_REMOTE_READ;
-#else
-    info->p_info->caps &= (!FI_REMOTE_WRITE & !FI_REMOTE_READ);
 #endif
-    info->p_info->tx_attr->op_flags = FI_DELIVERY_COMPLETE;
     /* Remove FI_CONTEXT flag from this endpoint */
     info->p_info->mode = 0;
     info->p_info->tx_attr->mode = 0;
--- a/src/transport_ofi.c
+++ b/src/transport_ofi.c
@@ -292,13 +292,6 @@ static inline int allocate_endpoints(struct fabric_info *info)
 
     /* this endpoint is used only to get read and write
      * counter updates */
-    info->p_info->caps = FI_RMA | FI_WRITE | FI_READ | /*SEND ONLY */
-                         FI_ATOMICS; /* request atomics capability */
-    info->p_info->caps |= FI_REMOTE_WRITE | FI_REMOTE_READ;
-    info->p_info->tx_attr->op_flags = FI_DELIVERY_COMPLETE | FI_INJECT_COMPLETE;
-    info->p_info->mode = 0;
-    info->p_info->tx_attr->mode = 0;
-    info->p_info->rx_attr->mode = 0;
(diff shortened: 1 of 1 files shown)


jdinan:
@sayantansur can you review while @kseager is OOO?

src/transport_ofi.c
```diff
@@ -292,13 +292,6 @@ static inline int allocate_endpoints(struct fabric_info *info)
 
     /* this endpoint is used only to get read and write
      * counter updates */
-    info->p_info->caps = FI_RMA | FI_WRITE | FI_READ | /*SEND ONLY */
-                         FI_ATOMICS; /* request atomics capability */
```
sayantansur: The comment indicates that the intent here to have a TX only endpoint. However, the caps indicate RX capabilities as well (also follow on bits that are OR'd in).

I'd like to understand better what the intent is, and maybe what we need here is to downselect some RX attributes from the endpoint that is being created.

src/transport_ofi.c
```diff
@@ -292,13 +292,6 @@ static inline int allocate_endpoints(struct fabric_info *info)
...
-    info->p_info->caps |= FI_REMOTE_WRITE | FI_REMOTE_READ;
-    info->p_info->tx_attr->op_flags = FI_DELIVERY_COMPLETE | FI_INJECT_COMPLETE;
```
sayantansur: This line is OK to be removed, and also FI_DELIVERY_COMPLETE | FI_INJECT_COMPLETE is unnecessary (it will get bumped to FI_DELIVERY_COMPLETE).

src/transport_ofi.c
```diff
@@ -292,13 +292,6 @@ static inline int allocate_endpoints(struct fabric_info *info)
...
-    info->p_info->mode = 0;
-    info->p_info->tx_attr->mode = 0;
-    info->p_info->rx_attr->mode = 0;
```
sayantansur: These three lines are trying to remove the ill effects of having agreed to a FI_CONTEXT mode bit? I suppose SHMEM shouldn't be asking for FI_CONTEXT at all, if bounce buffering is turned off ... right?

src/transport_ofi.c
jdinan: IIUC, setting the RMA caps is redundant, because we request the following in ```fi_getinfo```: ```FI_RMA | FI_ATOMICS | FI_RMA_EVENT```.  Or is the intent that maybe the domain has additional caps we don't need, so we want to disable them?  Do we need to set ```FI_REMOTE_WRITE | FI_REMOTE_READ``` or does requesting ```FI_RMA_EVENT``` already have this covered?

src/transport_ofi.c
jdinan: Makes sense -- this is setting info for the counting endpoint.  However, we don't bind a CQ to this endpoint, so does this info change even matter?

jdinan: Scratch that, we do bind a CQ for catching error events.

src/transport_ofi.c
sayantansur: Implicitly, FI_RMA == FI_READ | FI_WRITE | FI_REMOTE_WRITE | FI_REMOTE_READ. Same with FI_ATOMIC. Therefore, the explicit addition of flags in this routine is excessive, but not incorrect.

I think the intent was to disable the RX side caps from the domain. If so, then just remove line 297 and it will be fine.

jdinan:
@sayantansur Updated with your feedback.

jdinan:
Thanks for the review!

PR #248, opened 2016-09-28
Basic thread safety for OFI transport

jpdoyle:
If built with --enable-threads, this adds a global mutex to the OFI transport, which is locked on quiets, fences, and RMA/atomic operations.

### DIFF
--- a/src/transport_ofi.c
+++ b/src/transport_ofi.c
@@ -1074,9 +1074,7 @@ int shmem_transport_init(long eager_size)
     if(ret!=0)
         return ret;
 
-#ifdef ENABLE_THREADS
     SHMEM_MUTEX_INIT(shmem_transport_ofi_lock);
-#endif
 
     shmem_transport_ofi_bounce_buffer_size = eager_size;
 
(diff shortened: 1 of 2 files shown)


sayantansur:
Looks OK, although ideally I'd like to see the basic level just be FI_THREAD_SAFE. This might be better for now as not all providers have gotten around to supporting that (even though it is a requirement).

src/transport_ofi.c
```diff
@@ -1071,6 +1074,10 @@ int shmem_transport_init(long eager_size)
     if(ret!=0)
         return ret;
 
+#ifdef ENABLE_THREADS
```
jdinan: I don't think ```#ifdef ENABLE_THREADS``` is needed here.  The macros should automatically take care of this.

src/transport_ofi.c
```diff
@@ -1236,7 +1243,11 @@ int shmem_transport_fini(void)
     }
 
 #ifdef USE_AV_MAP
-  free(addr_table);
+    free(addr_table);
+#endif
+
+#ifdef ENABLE_THREADS
```
jdinan: Same thing here, ```#ifdef ENABLE_THREADS``` should not be needed.

PR #257, opened 2016-10-19
Add binary search example to "apps" tests

jdinan:


jdinan:
@jpdoyle - this is one of your examples, could you please review?

jdinan:
@jpdoyle I can has a review pls?

test/apps/gups.c
```diff
@@ -412,12 +411,12 @@ SHMEMRandomAccess(void)
 
   if (MyProc == 0) {
     fprintf( outFile, "Running on %d processors\n", NumProcs);
-    fprintf( outFile, "Total Main table size = 2^" FSTR64 " = " FSTR64 " words\n",
+    fprintf( outFile, "Total Main table size = 2^%" PRIu64 " = %" PRIu64 " words\n",
```
jpdoyle: This requires C99, but it seems reasonable to support a standard old enough to drive.

test/apps/binary-search.c
```diff
@@ -0,0 +1,82 @@
...
+    int i, errors = 0;
+
+    shmem_init();
+
+    for (i = 0; i < N_PER_PE; i++)
+        keys[i] = N_PER_PE * shmem_my_pe() + i;
+
+    shmem_barrier_all();
+
+    for (i = 0; i < N_PER_PE * shmem_n_pes(); i++) {
```
jpdoyle: This would be a pretty good stress-test for evaluating a fabric. It might be a bit better for that if the loop is offset by shmem_my_pe()*N_PER_PE so it doesn't bottleneck on a particular node.

Aside from that, it looks correct to me.

test/apps/gups.c
jdinan: This is true, and I'm generally fine with accepting C99 as a baseline for SOS.  In fairness to the PR, the C99 dependence was already there because the fixed-size integer types were added in '99.

PR #259, opened 2016-10-26
OFI: Disable unused caps

jdinan:
Signed-off-by: James Dinan james.dinan@intel.com

### DIFF
--- a/src/transport_ofi.c
+++ b/src/transport_ofi.c
@@ -293,19 +293,6 @@ static inline int allocate_endpoints(struct fabric_info *info)
 	return ret;
     }
 
-    /* In hard polling builds, ask for remote read/write counter updates to
-     * support shmem_wait that blocks on an update arriving from the network */
-#ifndef ENABLE_HARD_POLLING
-    info->p_info->caps |= FI_REMOTE_WRITE | FI_REMOTE_READ;
-#else
-    info->p_info->caps &= ~FI_REMOTE_WRITE;
-    info->p_info->caps &= ~FI_REMOTE_READ;
-#endif
-
-    /* Remove unused capabilities */
-    info->p_info->caps &= ~FI_MSG;
-    info->p_info->caps &= ~FI_TAGGED;
-
     /* Remove FI_CONTEXT flag from this endpoint */
     info->p_info->mode = 0;
     info->p_info->tx_attr->mode = 0;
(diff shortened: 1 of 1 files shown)


jdinan:
@kseager, @sayantansur - please review

src/transport_ofi.c
```diff
@@ -297,7 +297,15 @@ static inline int allocate_endpoints(struct fabric_info *info)
      * support shmem_wait that blocks on an update arriving from the network */
 #ifndef ENABLE_HARD_POLLING
     info->p_info->caps |= FI_REMOTE_WRITE | FI_REMOTE_READ;
+#else
+    info->p_info->caps &= ~FI_REMOTE_WRITE;
+    info->p_info->caps &= ~FI_REMOTE_READ;
```
kseager: if you want a send-only you should instead set FI_WRITE and FI_READ

src/transport_ofi.c
```diff
@@ -297,7 +297,15 @@ static inline int allocate_endpoints(struct fabric_info *info)
...
 #endif
+
+    /* Remove unused capabilities */
+    info->p_info->caps &= ~FI_MSG;
+    info->p_info->caps &= ~FI_TAGGED;
```
kseager: this is unnecessary, we don't need to clear these bits if we never set them in the first place

sayantansur: OTOH, some providers might just provide you with more functionality than you asked for. No harm in turning them off again.

kseager: but then whats the point of turning them off? either its a provider bug if having them on the caps changes functionality from what you asked the provider for or there is nothing to gain from turning them off; that's just my two cents

sayantansur: Yea, ignore this, and probably better to delete this part of the patch.

src/transport_ofi.c
sayantansur: Discussed with Kayla here. Adding permissions to read/write if not hard polling seems wrong to me. You need the permissions to read/write anyways (and already covered under FI_RMA), and hard polling only controls whether you look at the target counter or not e.g. FI_RMA_EVENT.

kseager: yeah guys what did you do in commit: 082c6747e89626bb6facc719e63db040b82388e7 when I was out of town? ;) just kidding. Agreed we should just be keeping the caps=FI_RMA for the ep creation regardless

sayantansur: Yea, it looks like I messed up on that one. Line 295-299 in 082c674 should be deleted.

jdinan: *facepalm*

sayantansur:
This one looks good to me 👍

jdinan:
@kseager Could you please review the updated change set and merge if it look good?

PR #270, opened 2016-11-07
Check return code from asprintf

jdinan:
Quiets build warnings, "ignoring return value of ‘asprintf’, declared
with attribute warn_unused_result", and is generally a good idea.


sayantansur:
👍

Issue #21, opened 2015-10-22
Basic error checking: Local/Remote Buffer


jdinan:
Check that local/remote buffer argument is non-NULL when length > 0.  If length == 0, functions should return immediately.

jdinan:
Also, we should add bounds checking to determine whether the remote variable is symmetric.  It's probably only feasible for the origin process to do this check when remote virtual addressing is enabled.

PR #271, opened 2016-11-07
Improve argument error checking

### LINKED ISSUE #20: Basic error checking: PE argument
Check that PE argument is valid

jdinan:
Closes #20 and #21
Optional feature, must be enabled with --enable-error-checking
Enable UH error unit tests


jdinan:
Right now I'm checking that a given ``ptr`` argument is symmetric.  Should we also check that ```ptr + len``` lies fully within the symmetric region?

jdinan:
Pulling back for now.  I'll macroize the bindings, reapply, and then resubmit the PR.

PR #272, opened 2016-11-07
Update shmemx_wtime

jdinan:
Add higher resolution timing to Fortran bindings and include time.h to
ensure clock_gettime will be used when available.

### DIFF
--- a/src/query_f.c
+++ b/src/query_f.c
@@ -59,20 +59,7 @@ fortran_double_precision_t FC_SHMEMX_WTIME(void);
 fortran_double_precision_t
 FC_SHMEMX_WTIME(void)
 {
-    double wtime = 0.0;
-
     SHMEM_ERR_CHECK_INITIALIZED();
 
-#ifdef CLOCK_MONOTONIC
-    struct timespec tv;
-    clock_gettime(CLOCK_MONOTONIC, &tv);
-    wtime = tv.tv_sec;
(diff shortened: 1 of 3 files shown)


src/query_f.c
```diff
@@ -58,11 +59,20 @@ fortran_double_precision_t FC_SHMEMX_WTIME(void);
...
-     wtime = tv.tv_sec;
-     wtime += (double)tv.tv_usec / 1000000.0;
-     return wtime;
+    SHMEM_ERR_CHECK_INITIALIZED();
+
+#ifdef CLOCK_MONOTONIC
+    struct timespec tv;
+    clock_gettime(CLOCK_MONOTONIC, &tv);
+    wtime = tv.tv_sec;
+    wtime += (double)tv.tv_nsec / 1000000000.0;
```
sayantansur: Does a format like 1.0e9 work? I hate counting zeros.

jdinan: Sure, and as long as I'm editing ...  :)

src/query_f.c
```diff
@@ -12,6 +12,7 @@
 
 #include "config.h"
 
+#include <time.h>
 #include <sys/time.h>
```
kseager: should we drop the extra includes since they have migrated to shmem_internal.h?

jdinan:
@kseager - Updated

PR #274, opened 2016-11-09
Sync PMI with upstream and clean up odds and ends

jdinan:


jdinan:
This PR might be easiest to review patch-by-patch.

jdinan:
Ready for review.

shmem_pmi/simple_pmi.c
```diff
@@ -8,15 +8,17 @@
 /*
  * This file implements the client-side of the PMI interface.
  *
- * Note that the PMI client code must not print error messages (except
+ * Note that the PMI client code must not print error messages (except
```
kseager: would you mind removing the trailing whitespace in some of these files?

jdinan: We're downstream on these files, so the less we edit them the better.  With the changes in this PR, there are a total of 3 edits, to comment out includes of files we don't have.

What if I add a README instead, documenting the upstream source and the update procedure?

kseager: that's true. I'll take the README option instead :)

PR #277, opened 2016-11-14
BOGO: Argument error checking and supermacrohappiness part two!

jdinan:
* Full argument checking, enabled by --enable-error-checking.
 * Refactor C and Fortran bindings to be generated via macros.  All of the "typed" API operations should now be generated.  Note that I did not apply this refactoring in cases where it would have resulted in an increase in code size.  For example, Fortran fixed-size-only operations (e.g. 32/64 collectives) have only a few instances of each function and refactoring these didn't seem sensible.


src/atomic_c.c
```diff
@@ -118,72 +118,21 @@
...
-{
-    long long newval;
-
-    SHMEM_ERR_CHECK_INITIALIZED();
-
-    shmem_internal_swap(target, &value, &newval, sizeof(long long), pe, SHM_INTERNAL_LONG_LONG);
-    shmem_internal_get_wait();
-    return newval;
-}
+SHMEM_EVAL_MACRO_FOR_EXTENDED_AMO_WITH_ITYPE(SHMEM_DEF_SWAP,)
```
kseager: I'm probably missing something but doesn't this macro only include float/double? what about long/int/longlong

src/atomic_c.c
```diff
@@ -118,306 +118,139 @@
...
+    }
+
+SHMEM_BIND_C_AMO(SHMEM_DEF_FINC)
+
+#define SHMEM_DEF_ADD(STYPE,TYPE,ITYPE)                                 \
+    void shmem_##STYPE##_add(TYPE *target, TYPE value, int pe)          \
+    {                                                                   \
+        SHMEM_ERR_CHECK_INITIALIZED();                                  \
+        SHMEM_ERR_CHECK_PE(pe);                                         \
+        SHMEM_ERR_CHECK_SYMMETRIC(target, sizeof(TYPE));                \
```
kseager: you could further refactor...you probably don't want to at this point, by creating an arg_check function to put all these checks into =) save of 2 lines per repeat

src/shmem_internal.h
```diff
@@ -105,11 +121,101 @@ extern int shmem_internal_debug;
...
+            if (ptr_ext > data_ext) {                                                   \
+                fprintf(stderr, "ERROR: %s(): Argument \"%s\" [%p..%p) exceeds "        \
+                                "sym. data region [%p..%p)\n",                          \
+                        __func__, #ptr_in, ptr_base, ptr_ext,                           \
+                        shmem_internal_data_base, data_ext);                            \
+                shmem_runtime_abort(100, PACKAGE_NAME " exited in error");              \
+            }                                                                           \
+        }                                                                               \
+        else if (ptr_base >= shmem_internal_heap_base && ptr_base < heap_ext) {         \
+            if (ptr_ext > heap_ext) {                                                   \
```
kseager: we could just combine this into one check right? since if ptr_ext < heap_ext, then it must follow that ptr_base < heap_ext, since by we know ptr_base < ptr_ext

src/shmem_internal.h
```diff
@@ -246,4 +352,71 @@ static inline double shmem_internal_wtime(void) {
...
+  decl(logical,  LOGICAL,       SIZEOF_FORTRAN_LOGICAL)                 \
+  decl(real,     REAL,          SIZEOF_FORTRAN_REAL)
+
+#define SHMEM_BIND_F_SIZES(decl)        \
+  decl(mem,     MEM,    1)              \
+  decl(4,       4,      4)              \
+  decl(8,       8,      8)              \
+  decl(32,      32,     4)              \
+  decl(64,      64,     8)              \
+  decl(128,     128,    16)
```
kseager: I guess with macros there is not much we can do with getting around these types of size defs...

kseager:
+987 −3,302 👍 
great job on this refactor. It looks great. Its definitely nice to see the repetition shrink and strengthen our input checking.

src/shmem_internal.h
jdinan: Yeah, because 4 and 32 are both 4 bytes.  Fortran bindings are a total 🗑 🔥 .

jdinan:
Thanks for the review!

The error checking identified more lurking memory corruption bugs in unit tests: openshmem-org/tests-uh#14.  😃

Issue #280, opened 2016-11-29
Generate pshmem.h from shmem.h


jdinan:
The ```pshmem.h``` header is a remnant from when the headers were written by hand.  This needs to be updated so that it is generated from the contents of ```shmem.h``` to reduce the potential for errors in maintaining the bindings.

PR #281, opened 2016-11-29
Generate pshmem library interposition interface

jdinan:
In all of the cases where we generate the C bindings, also generate the profiling interface hooks.  This is accomplished through the new ".hin" files that are processed into C header files using a combination of the C preprocessor and a few text processing rules implemented in Perl (see Makefile.am).

A lingering TODO item is to generate the pshmem.h header file from the shmem.h header file to limit the maintenance burden associated with the profiling interface.  This is captured in #280.


jdinan:
Hold on this PR for now.  I am rewriting the bindings generation using M4.  Clem shamed me into doing this properly.

PR #282, opened 2016-12-08
Contexts

jpdoyle:
Adds contexts and domains (i.e. communication and resource handles) to Sandia-OpenSHMEM.

This is probably best to bring in as a separate branch, since it is not tightly optimized yet, and doesn't support bounce-buffering with contexts yet.

### DIFF
--- a/mpp/shmem.h4.in
+++ b/mpp/shmem.h4.in
@@ -72,7 +72,6 @@ extern "C" {
 #define  SHMEM_VENDOR_STRING "Sandia OpenSHMEM"
 #define _SHMEM_VENDOR_STRING SHMEM_VENDOR_STRING
 
-define(`SHFN', `shmem_')dnl
 define(`SHPRE', `')dnl
 include(shmem_c_func.m4)dnl
 
@@ -84,7 +83,6 @@ include(shmem_c_func.m4)dnl
 #endif
 
 /* Blocking block, scalar, and block-strided put */
-dnl args: TYPENAME,TYPE
 define(`SHMEM_CXX_PUT',
 `static inline void shmem_put($2* dest, const $2* source,
                              size_t nelems, int pe) {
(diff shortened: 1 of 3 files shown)


src/shmem_lock.h
```diff
@@ -38,12 +38,12 @@ shmem_internal_clear_lock(volatile long *lockp)
     lock_t *lock = (lock_t*) lockp;
     int curr, cond, zero = 0, sig = SIGNAL_MASK;
 
-    shmem_internal_quiet();
+    shmem_quiet();
```
jdinan: Upcall bad.  What not call shmem_internal_quiet?

src/synchronization_f.c
```diff
@@ -30,7 +30,7 @@ FC_SHMEM_QUIET(void)
 {
     SHMEM_ERR_CHECK_INITIALIZED();
 
-    shmem_internal_quiet();
+    shmem_quiet();
```
jdinan: More upcalls (calling upper/public API from within the implementation)

src/transport_ofi.c
```diff
@@ -255,185 +278,303 @@ typedef enum{
...
-		    &shmem_transport_ofi_put_nb_cqfd->fid, FI_SELECTIVE_COMPLETION | FI_TRANSMIT);
-    if(ret!=0){
-	OFI_ERRMSG("ep_bind cntrep2cq_nb failed\n");
-	return ret;
-    }
+  struct fabric_info* info = &shmem_ofi_cntr_info;
+  info->p_info->ep_attr->tx_ctx_cnt = FI_SHARED_CONTEXT;
+  info->p_info->caps = FI_RMA | FI_WRITE | FI_READ | /*SEND ONLY */
+                        FI_ATOMICS; /* request atomics capability */
+  info->p_info->caps |= FI_REMOTE_WRITE | FI_REMOTE_READ;
```
jpdoyle: Remote read/write are no-ops because there is no receive context.

Issue #285, opened 2016-12-08
remove FI_CONTEXT usage


kseager:
Do not need FI_CONTEXT 

fix: 
1) remove mode setting
2) remove fi_context struct usage from bounce buffering

PR #287, opened 2016-12-10
remove fi_context usage

kseager:
not applicable to RMA fixes issue #285.

### DIFF
--- a/src/transport_ofi.h
+++ b/src/transport_ofi.h
@@ -182,12 +182,10 @@ extern fi_addr_t *addr_table;
 #define GET_DEST(dest) ((fi_addr_t)(dest))
 #endif
 
-typedef struct fi_context fi_context_t;
 
 struct shmem_transport_ofi_frag_t {
     shmem_free_list_item_t item;
     uint8_t mytype;
-    fi_context_t context;
 };
 
 typedef struct shmem_transport_ofi_frag_t shmem_transport_ofi_frag_t;
@@ -240,13 +238,12 @@ void shmem_transport_ofi_drain_cq(void)
 
 		if(ret == 1) {
 			shmem_transport_ofi_frag_t *frag =
-				container_of(buf.op_context,
-				struct shmem_transport_ofi_frag_t,
-				context);
+                (shmem_transport_ofi_frag_t *) buf.op_context;
 
 			if(SHMEM_TRANSPORT_OFI_TYPE_BOUNCE == frag->mytype) {
 				shmem_free_list_free(
-					shmem_transport_ofi_bounce_buffers, frag);
+					shmem_transport_ofi_bounce_buffers,
+                    (shmem_transport_ofi_bounce_buffer_t *) frag);
                         }
                         else {
                             RAISE_ERROR_STR("Unrecognized completion object");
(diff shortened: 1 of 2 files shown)


src/transport_ofi.h
```diff
@@ -240,13 +238,13 @@ void shmem_transport_ofi_drain_cq(void)
 
 		if(ret == 1) {
 			shmem_transport_ofi_frag_t *frag =
-				container_of(buf.op_context,
-				struct shmem_transport_ofi_frag_t,
-				context);
+                (shmem_transport_ofi_frag_t *) buf.op_context;
 
 			if(SHMEM_TRANSPORT_OFI_TYPE_BOUNCE == frag->mytype) {
+			    shmem_transport_ofi_bounce_buffer_t *bb =
```
sayantansur: You could get rid of this by casting directly in the list_free macro?

src/transport_ofi.h
```diff
@@ -460,8 +458,12 @@ static inline
 void
 shmem_transport_put_wait(long *completion) {
 
-    if((*completion)-- > 0)
+    shmem_internal_assertp((*completion) >= 0);
```
jdinan: It should be fine to use a regular assertion here.  The ```p``` in ```assertp``` will cause it to persist even when error checking is disabled.

kseager:
thanks for the feedback/review

jdinan:
A note on this: ```fi_context``` was there to support the BG provider.  This should still work as long as we disable bounce buffering (so we don't generate CQ events).

sayantansur:
Thanks for finding out why we added FI_CONTEXT in the first place!

jdinan:
Does ```FI_CONTEXT``` require a ```struct fi_context``` in every operation, or just the operations that will generate CQ entries?  What I'm wondering is whether it's enough for us to detect ```FI_CONTEXT``` in ```fi_info:mode``` and disable bounce buffering (the only instance where we currently generate CQ entries), or whether we would have to generate CQ entries for every operation to manage ```struct fi_context``` objects at the middleware level?

sayantansur:
It is required for operations that generate a completion entry. The application can remove the generation of completion entry by using the `FI_COMPLETION` flag (I think SHMEM is using that?). In that case the provider does not require `struct fi_context`.

The problem arises when there are two endpoints with differing `FI_CONTEXT` mode bits set that are attached to the same completion queue. When a completion is generated, the provider will get confused whether it can read the `context` parameter or not. Provider can't tell when a completion is generated due to FI_COMPLETION being a run time flag that can be controlled by the application. Therefore, this will be disallowed in ABI 1.1.

I'm not certain that detection of `FI_CONTEXT` is sufficient reason to disable bounce buffering, as they are really different things. Bounce buffering is used when memory copy bw > network draining bw. And FI_CONTEXT indicates whether or not the provider requires additional bits to track the transfer.

SHMEM can support `FI_CONTEXT` pretty easily since the only operations you require completions for (bounce buffers) have an easy place to store the struct.

I wonder if this should become its own compile time/configure time parameter?

jdinan:
SOS _used_ to support FI_CONTEXT for the CQ events generated in the bounce buffering path.  This PR removed that support.

The stakes here are pretty low..  We only generate CQ events in the bounce buffering code (where we used to -- but as of this PR no longer -- support FI_CONTEXT).  If the provider requires FI_CONTEXT, we can easily achieve compatibility by disabling bounce buffering.  Or we can back out this PR.  I tend to favor just detecting the FI_CONTEXT requirement and disabling bounce buffering (most users want this anyway) when we bring up the OFI transport.

sayantansur:
Yea, if that makes life easier, I'm all for it.

PR #290, opened 2016-12-12
OFI: Fix missing fi_wait_obj attr on cntrs

jdinan:


jdinan:
@kseager @sayantansur - We always wait through the FI interface, so ```FI_WAIT_UNSPEC``` looks correct to me.

@kseager, you had wanted to also do some performance tuning of the wait behavior.  Do you want to log an issue for that, or propose a change?

sayantansur:
Yea, this looks correct to me.

kseager:
I guess given the urgency of this fix this works for now. I just tested it on gni and we are again compliant. thanks for this fix and yes lets file a performance tuning todo.

jdinan:
Partially addresses #290

PR #291, opened 2016-12-14
Perf tests

kseager:
fixed a bug in the atomic bindings, fixed a bug in the latency IO, and enabled ease of use with SHMEM version 1.0 (IFS)

### DIFF
--- a/test/performance/shmem_perf_suite/bw_common.h
+++ b/test/performance/shmem_perf_suite/bw_common.h
@@ -196,7 +196,9 @@ void static command_line_arg_check(int argc, char *argv[],
                     "[-k (kilobytes/second)] [-b (bytes/second)] \n"\
                     "[-v (validate data stream)] \n");
         }
+#ifndef VERSION_1.0
         shmem_finalize();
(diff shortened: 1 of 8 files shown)


test/performance/shmem_perf_suite/bw_common.h
```diff
@@ -196,7 +196,9 @@ void static command_line_arg_check(int argc, char *argv[],
                     "[-k (kilobytes/second)] [-b (bytes/second)] \n"\
                     "[-v (validate data stream)] \n");
         }
+#ifndef VERSION_1.0
```
jdinan: You may want this to be a proper number, like 100 (where XXYY is XX is the major number and YY is the minor number), in case you have multiple of these checks down the line.

jdinan: Actually, nevermind that comment.  I see now that this is for OpenSHMEM 1.0 compatibility, which is even before we had SHMEM_MAJOR/MINOR_VERSION..  :(  Ok, this is fine..

PR #293, opened 2016-12-14
Add barrier to atexit handler

jdinan:
Necessary for compliance with OpenSHMEM v1.3 spec.


jdinan:
@kseager - this should be ready to go

PR #294, opened 2016-12-14
Add --enable-debug and cleanup --enable-picky

jdinan:


kseager:
what about asserts? Should we turn those on in debug only? or have an option for turning them off?

jdinan:
Right now, asserts are enabled with ```--enable-error-checking``` (and assertp is always enabled).  Debugging symbols don't tend to impact performance (or at least not in the way that asserts do), so it seems useful for these to be separate options.

PR #295, opened 2016-12-16
Contexts

jpdoyle:
Adds contexts and domains (i.e. communication and resource handles) to Sandia-OpenSHMEM.

This is probably best to bring in as a separate branch, since it is not tightly optimized yet, and doesn't support bounce-buffering with contexts yet.


jdinan:
Replace by #302.  My plan is to reduce the delta from master using that PR, then merge in changes from master and push the contexts work to the SOS repository as a separate branch.

PR #300, opened 2017-01-03
Fix collect bug, improve bandwidth efficiency, and add unit test

jdinan:


jdinan:
@kseager

kseager:
Could you do a quick sanity check on gni? I highly doubt there will be an issue but its good to verify.  

Otherwise this looks clean to me. Additionally, I double checked the bug find/resolution and that also checks out. So from my POV this looks great. Good job, so glad you found this! 👍

jdinan:
There shouldn't be any change here that impacts the GNI provider build and it's pretty time consuming to run tests there.  Can we check that platform periodically, rather than per PR?  (it would be so nice if we had nightlies running on that platform...)

kseager:
ok fair enough

PR #301, opened 2017-01-04
Make the "no transport" build work again

jdinan:
None of the tests will actually work if you try to run them (all will exit with a no transport error), but build should always complete successfully and is a useful canary for finding other bugs.

 * Fix errors preventing the "no transport" build from completing successfully
 * Add the "no transport" build to the Travis matrix
 * Prevent the "no transport" code from bit rotting


jdinan:
Apparently, we introduced a dependence on Linux librt when we added ``clock_gettime```:
```
CCLD   circular_shift
../../src/..libs/libsma.so: undefined. /reference .to `.clock_gettime/'
src/.libs/libsma.so: undefined reference to `clock_gettime'
```
Looks like we've been getting lucky, because OFI and Portals have been pulling this in for us.  Should just require a build system fix.

jdinan:
Commit ce198aa seems to have fixed that problem.  Now for all of the compiler warnings ...

jdinan:
This should be good to merge once Travis signs off.

FYI, these changes are in support of the contexts branch (because of the transport updates involved in contexts).

src/shmem_internal.h
```diff
@@ -337,7 +337,7 @@ void* shmem_internal_get_next(intptr_t incr);
 static inline double shmem_internal_wtime(void) {
     double wtime = 0.0;
 
-#ifdef CLOCK_MONOTONIC
+#ifdef HAVE_CLOCK_GETTIME
```
kseager: don't we additionally need to check if CLOCK_MONOTONIC is also defined? or was that picked up by your configure.ac check? Also we run into this same issue in the perf tests

.travis.yml
```diff
@@ -111,6 +111,28 @@ install:
...
+    - make $TRAVIS_PAR_MAKE check TESTS=
+    - make install
+    ###
+    ### Build with only CMA
+    ###
+    - cd $SOS_SRC
+    - mkdir cma-build
+    - cd cma-build
+    - export PATH=$TRAVIS_INSTALL/hydra/bin:$BASE_PATH
+    - ../configure --prefix=$TRAVIS_INSTALL/sandia-shmem-cma $SOS_GLOBAL_BUILD_OPTS $SOS_BUILD_OPTS
```
jdinan: Good catch!  XPMEM requires a kernel patch, so it would be hard to test in Travis.  You could potentially try to fake the user space part of XPMEM without the kernel patch to exercise the SOS code, but it would be a hack.  The most practical option is probably to test the XPMEM code on a local cluster that is patched or by hand on an XC.  The better option would be to have this run nightly somewhere..

src/shmem_internal.h
jdinan: I don't think it's necessary to check.  This is one of the standard clocks, so it should be supported by any modern Linux platform (or another way to look at this is that if you get a build failure here, there is probably an issue with your platform and it's better to raise an error than to silently give a low resolution timer).  We were using ```CLOCK_MONOTONIC``` as a gate in the old code, because it was the easiest way to check if the function exists.  The new configure check does two things -- it looks for ```clock_gettime``` and if it finds it, it defines ```HAVE_CLOCK_GETTIME``` and it also adds ```librt``` to the linker flags.

jdinan:
Oh, crud.  The CMA builds are broken in Travis:
```
[000] ERROR: ../../src/transport_cma.h:108: process_vm_writev() failed
```
We had --enable-cma instead of --with-cma (fixed in 6e249cb) in the Travis build matrix.  This might just be a Travis issue; I'll look into it.  I'm pretty sure that I've been building with CMA enabled on our Linux cluster.

jdinan:
CMA works on our cluster with Linux kernel 2.6.32-431.11.2.el6.x86_64.  Does not work in Travis with Ubuntu 14.04.5 LTS and Linux Kernel 3.13.0-29-generic.

jdinan:
This is an Ubuntu security feature: https://wiki.ubuntu.com/SecurityTeam/Roadmap/KernelHardening#ptrace_Protection

We need to ```echo 0 > /proc/sys/kernel/yama/ptrace_scope```.  Unfortunately, this may mean requiring sudo, which moves us out of the Travis container environment.

jdinan:
@kseager This should be ready to merge now.  If we notice that the builds are taking much longer (now that we are back to using sudo, which moves us out of the containers environment), we can disable CMA testing in the Travis build matrix.

src/shmem_internal.h
kseager: agreed on letting it give the user a build failure

PR #308, opened 2017-01-12
Fix "./shmem_prog" support

jdinan:


jdinan:
Note that I made this change only to the PMI-1 runtime interface, which we use with Hydra.  I don't know whether PMI-2 process managers also have this same issue and I didn't want to break something that's not broken.

kseager:
👍

Issue #309, opened 2017-01-12
C11 Support


jdinan:
We have ```AC_PROG_C_C99``` in ```configure.ac```, which causes ```oshcc``` to pass ```--std=gnu99``` in the wrapper.  This is bad, because we want the wrapper to support C11 whenever it's available.

jdinan:
Autoconf does not have a ```AC_PROG_C_C11``` macro.  The most sensible thing may be to remove the C99 check.

PR #310, opened 2017-01-12
Remove AC_PROG_CC_C99 from configure

davidozog:
Doesn't seem to be necessary to force -std=c99.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -165,7 +165,7 @@ fi
 AX_GCC_FUNC_ATTRIBUTE([noreturn])
 
 if test "$enable_picky" = "yes" -a "$GCC" = "yes" ; then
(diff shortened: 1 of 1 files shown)


jdinan:
Closes #309

jdinan:
It looks like we need to split the CFLAGS used to build SOS from the CFLAGS that we embed in the oshcc wrapper.  The problem is that -pedantic (enabled via the --enable-picky configure option) is assuming ISO C90 when we remove the C99 flag.

jdinan:
I added "-std=c99" to the picky CFLAGS.  Turns out that we don't propagate CFLAGS to the wrapper, so this should be all that is needed to close out this change.

jdinan:
This branch was a ways behind master.  Merging in those changes to see if it fixes the build errors.

Issue #284, opened 2016-12-08
Optimize cntr_wait behavior


kseager:
SHMEM-OFI currently asks cntr_wait to block without a wait object. This is a bug caught by the gni provider and we currently do not pass tests on this provider as a result: 
https://github.com/ofi-cray/libfabric-cray/issues/1038#issuecomment-265783399

Fix:

1) (Fixed by #290)  set fi_wait_object to type FI_WAIT_UNSPEC
2) (Not yet addressed) since cntr_wait is expensive, mix cntr_read/cntr_wait usage
while(under spin threshold) {cntr_read}
if still not at threshold, block on fi_cntr_wait

jdinan:
Close by #398

PR #312, opened 2017-01-13
Add support for local/remote completion polling

jdinan:

### DIFF
--- a/src/transport_ofi.h
+++ b/src/transport_ofi.h
@@ -26,11 +26,12 @@
 #if USE_PORTALS4
 #include <portals4.h>
 #endif
-#include "shmem_free_list.h"
 #include <string.h>
-#include "shmem_internal.h"
 #include <unistd.h>
 #include <stddef.h>
+#include "shmem_free_list.h"
+#include "shmem_internal.h"
+#include "shmem_atomic.h"
 
 extern struct fid_ep*			shmem_transport_ofi_epfd;
 extern struct fid_ep*			shmem_transport_ofi_cntr_epfd;
@@ -293,7 +294,11 @@ static inline void shmem_transport_put_quiet(void)
     do {
         success = fi_cntr_read(shmem_transport_ofi_put_cntrfd);
         fail = fi_cntr_readerr(shmem_transport_ofi_put_cntrfd);
-        if (fail) {
+
+        if (success < shmem_transport_ofi_pending_put_counter && fail == 0) {
+            SPINLOCK_BODY();
+        }
+        else if (fail) {
             struct fi_cq_err_entry e = {0};
             fi_cq_readerr(shmem_transport_ofi_put_nb_cqfd, (void *)&e, 0);
             SHMEM_MUTEX_UNLOCK(shmem_transport_ofi_lock);
@@ -558,7 +563,11 @@ shmem_transport_get_wait(void)
     do {
         success = fi_cntr_read(shmem_transport_ofi_get_cntrfd);
         fail = fi_cntr_readerr(shmem_transport_ofi_get_cntrfd);
-        if (fail) {
+
+        if (success < shmem_transport_ofi_pending_put_counter && fail == 0) {
+            SPINLOCK_BODY();
+        }
+        else if (fail) {
             struct fi_cq_err_entry e = {0};
             fi_cq_readerr(shmem_transport_ofi_put_nb_cqfd, (void *)&e, 0);
             SHMEM_MUTEX_UNLOCK(shmem_transport_ofi_lock);
--- a/src/transport_ofi.h
+++ b/src/transport_ofi.h
@@ -564,7 +564,7 @@ shmem_transport_get_wait(void)
         success = fi_cntr_read(shmem_transport_ofi_get_cntrfd);
         fail = fi_cntr_readerr(shmem_transport_ofi_get_cntrfd);
 
-        if (success < shmem_transport_ofi_pending_put_counter && fail == 0) {
+        if (success < shmem_transport_ofi_pending_get_counter && fail == 0) {
             SPINLOCK_BODY();
         }
         else if (fail) {
--- a/src/transport_ofi.h
+++ b/src/transport_ofi.h
@@ -280,26 +280,37 @@ static inline shmem_transport_ofi_bounce_buffer_t * create_bounce_buffer(const v
 
 static inline void shmem_transport_put_quiet(void)
 {
-  SHMEM_MUTEX_LOCK(shmem_transport_ofi_lock);
+    SHMEM_MUTEX_LOCK(shmem_transport_ofi_lock);
 
-	int ret = 0;
-
-	/* wait until all outstanding queue operations have completed */
-	while(shmem_transport_ofi_pending_cq_count) {
-		shmem_transport_ofi_drain_cq();
-	}
-
-	/* wait for put counter to meet outstanding count value    */
-	ret = fi_cntr_wait(shmem_transport_ofi_put_cntrfd,
-			shmem_transport_ofi_pending_put_counter,-1);
-    if(ret) {
-	    struct fi_cq_err_entry e = {0};
-        fi_cq_readerr(shmem_transport_ofi_put_nb_cqfd,
-		           (void *)&e, 0);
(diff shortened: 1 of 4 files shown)


jdinan:
@kseager

kseager:
you checked this on gni right?

jdinan:
No, only locally using the sockets provider.

kseager:
a step towards #284, @jswaro for gni review

jdinan:
Enable using --enable-completion-polling configure argument.

kseager:
@jswaro yes if you could test it on gni we would appreciate it. The turn around time on CORI can be pretty rough. 

Unfortunately though, looking at the gni provider I don't think this is avoiding the wait object like you wanted since we are still asking for WAIT_UNSPEC, and cntr_read picks up the wait object and just puts in a nonblocking poll() vs the blocking poll we had before. Are you trying to avoid the unix poll and just do the __gnix_cntr_progress? If so then I guess using FI_WAIT_NONE would not create the wait object. Either way (cntr_wait/not) I'm seeing it timeout on CORI for the perf tests. Let me know what you find.

jswaro:
Yes, we are trying to avoid the unix poll. It is incredibly slow for the GNI provider.

jswaro:
In this particular PR, using my build options given below, I encounter an error that you might want to look into in a different issue. It is easily fixed by include shmem_internal.h in the runtime-pmi2.c file. 

```bash 
#!/bin/sh

./autogen.sh
./configure --prefix=$HOME/sos --with-ofi=$HOME/libfabric/install --enable-lengthy-tests --disable-fortran --disable-mr-scalable --with-pmi=/opt/cray/pmi/default --with-xpmem=/opt/cray/xpmem/default
make -j12 && make install
```
runtime-pmi2.c: In function 'shmem_runtime_init':
runtime-pmi2.c:98:26: warning: unused variable 'my_node' [-Wunused-variable]
     int spawned, appnum, my_node;
                          ^~~~~~~
runtime-pmi2.c: In function 'shmem_runtime_abort':
runtime-pmi2.c:144:9: error: 'shmem_internal_trap_on_abort' undeclared (first use in this function)
     if (shmem_internal_trap_on_abort)
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

**Comment by jswaro** [comment]
Nevermind, I saw the other PRs and you are already aware of the issue.

**Comment by jswaro** [comment]
If you set the FI_WAIT* objects to FI_WAIT_NONE, this appears to resolve the performance issues (w.r.t. the wait set performance). 

```
Results for 2 PEs 500 trials with window size 64 using incast communication style

Operation           Uni-directional Bandwidth           Message Rate                 Latency
                    in megabytes per second         in Million ops/second        in microseconds
shmem_add
int                         5.83                              1.46                    0.69
long                       11.01                              1.38                    0.73
longlong                   11.22                              1.40                    0.71

```


as compared with pre-patch performance of about ~1.2 MB/s for shmem_add with int.

**Comment by jdinan** [comment]
Thanks!  I'll add ```FI_WAIT_NONE``` to the ```--enable-completion-polling``` build.

**Comment by jswaro** [comment]
@jdinan There are two locations in the transport_ofi.c file where the wait object type is specified. It doesn't look like you changed both of them.

**Comment by jdinan** [comment]
The completion polling patch I added handles the cases where the runtime is waiting for local/remote completion of operations issued by the local PE.  The other counter that's there is used to wait for remotely issued operations to complete (e.g. in a call to ```shmem_wait```).  I thought that you guys are building with ```--enable-hard-polling``` (this causes SOS to poll memory and doesn't touch the network, so we don't even create the target counter), since enabling XPMEM/CMA will switch on hard polling.  Does GNI need a third option to create the target counter with ```FI_WAIT_NONE``` and tap the counter during the polling loop?  Or is there something else we should do in the polling loop to stimulate network progress?

That would give us three modes:
1) Wait for a network event, then wake up and look at memory to see if wait condition is met
2) Poll memory until wait condition is met (does not require ```FI_RMA_EVENT``` support)
3) Poll memory and tap network until wait condition is met

**Comment by jdinan** [comment]
@kseager @jswaro Could you weigh in on the target-side counter question above?

**Comment by kseager** [comment]
so essentially you are asking if wait_until (incoming messaging completion) needs a third option which is to cntr_read and poll similar to the implementation you just added for the fence/quiet (outgoing messaging completion)? In terms of the performance tests you won't see wait_until changes improve results since it isn't used for completion in any of the tests except for round trip latency. However this is another performance tuning parameter gni provider should consider. @jswaro any thoughts? Would it help gni to poll into the fabric periodically on the target counter vs the current hard polling where we never call into the fabric?

File: src/transport_ofi.h [hunk 1]
```diff
@@ -539,19 +555,36 @@ static inline
...
+        fail = fi_cntr_readerr(shmem_transport_ofi_get_cntrfd);
 
-	/* wait for get counter to meet outstanding count value    */
-	ret = fi_cntr_wait(shmem_transport_ofi_get_cntrfd,
-			shmem_transport_ofi_pending_get_counter,-1);
-    if(ret) {
-	    struct fi_cq_err_entry e = {0};
-        fi_cq_readerr(shmem_transport_ofi_put_nb_cqfd,
-		           (void *)&e, 0);
+        if (success < shmem_transport_ofi_pending_put_counter && fail == 0) {
```
kseager: get_cntrfd

jdinan: I don't think there is a mixup at the line you flagged, but I did have a mixup in get_wait().  Pushed a patch to fix this, although, I'm a little concerned that it didn't show up in testing...

kseager: ack, yes I meant get_wait

jswaro:
Periodically polling into the provider is generally useful, especially if using the manual progress mode. Polling into the provider moves the internals of the provider, which is why you must poll if using manual completion. Since you are using auto progress, it only serves to progress things if the progress thread hasn't had a chance to move a task forward. 

Auto-progress is slower than manual progress, but allows you to perform other tasks while the provider progresses transactions if you have other work to do.

jdinan:
@jswaro What's the best way to poll the provider?  If the ```shmem_wait``` operation is polling on memory (this is necessary once we enable shared memory optimizations or shared memory access via ```shmem_ptr```), then the target-side (i.e. RX) event counter is not necessary so we disable it.  Assuming we don't enable that counter, we're left with poking a TX-side event object for progress.  Does it matter which OFI completion object gets poked?  We we have several TX-side completion counters and a CQ for catching errors.  Is the progress the same no matter which of these event objects we poll?  Thanks!

jswaro:
Well... that is a really tough question. Some application see the best performance from hard polling of the provider if they don't have a progress thread. If there is a progress thread, hard polling of the provider can interfere with locking and cause unnecessary latency increases due to contention. In general, hard polling of the provider is best for performance. It will probably take some experimenting to find what is best for SHMEM. 

There is a fundamental problem with `--enable-hard-polling` as you've described it to me. It assumes some measure of byte order delivery that isn't guaranteed by the GNI provider. I didn't use it initially, because RMA transactions can deliver bytes out of order to target memory. This can cause the perception of false completions if the other flits in a packet take too long to arrive AND if the library just happened to be polling a byte that wasn't delivered last. 

I've been checking the documentation for libfabric and the only two pieces of ordering information that are published by providers are message ordering and completion ordering. Unfortunately, there is nothing about byte ordering. I'd guess that both Infiniband and sockets are both in-order byte delivery providers, but GNI is not. This seriously complicates a memory-polling approach.

jswaro:
Nevermind the comment about hard polling on memory. Howard commented to the fact that it is still correct if the app is following openshmem models for synchronization.

jdinan:
OpenSHMEM doesn't guarantee byte ordering either, so the user can't call ```shmem_wait``` on a flag that is updated as part of a larger transfer they are waiting for.  We are taking a little liberty and assuming that the networking layer won't show us a partial update to the short/int/long variable we're spinning on.

jdinan:
@kseager Can we merge this and file an issue to add fabric polling to the hard polling path (e.g. turn it on by default and add an ENV var to throttle/disable it)?  We need this patch now on the contexts dev branch.

PR #314, opened 2017-01-20
Use -std=gnu11 when available, gnu99 otherwise

davidozog:
Addresses #309 & #310 by compile-checking for gnu11/gnu99 support and setting -std in CFLAGS instead of CC.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -159,51 +159,33 @@ AC_PROG_CXX dnl required even with --disable-cxx due to automake conditionals
 m4_ifdef([AM_PROG_AR],[AM_PROG_AR])
 AC_PATH_PROG([PERL],[perl])
 
-# start -std check
+# start gnu11/gnu99 check
 ORIG_CFLAGS=$CFLAGS
+WRAPPER_CC=$CC
+AC_SUBST(WRAPPER_CC)
 
-AC_CACHE_CHECK([if -std=gnu11 works], [shmem_cv_c11_works],
-               [AC_LANG_PUSH([C])
-                CFLAGS="-pedantic-errors -std=gnu11"
-                AC_COMPILE_IFELSE(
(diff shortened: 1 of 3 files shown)


configure.ac
```diff
@@ -159,6 +159,35 @@ AC_PROG_CXX dnl required even with --disable-cxx due to automake conditionals
...
+                   #define f(X) _Generic((X), default: g)(X)
+                   ]], [[
+                        f(1);
+                        ]]
+                   )],
+                   [shmem_cv_c11_works="yes"], [shmem_cv_c11_works="no"])
+AS_IF([test "$shmem_cv_c11_works" = "no"],
+      [AC_MSG_RESULT([no])
+      AC_MSG_WARN([No C11 support detected, unable to test _Generic bindings.])
+      CFLAGS="-std=gnu99 $ORIG_CFLAGS"
```
jdinan: Do we want to use ```AC_PROG_CC_C99``` here instead of ```-std=gnu99```?  Not all C99 compilers support gnu99 or the ```-std``` argument.

davidozog: Oops, I think we do.  I'll fix that.

src/Makefile.am
```diff
@@ -130,7 +130,7 @@ EXTRA_DIST = shmem_compiler_script.in \
              shmem_launcher_script.in
 
 do_subst = sed -e 's|[@]PERL[@]|$(PERL)|g' \
-               -e 's|[@]CC[@]|$(CC)|g' \
+               -e 's|[@]WRAPPER_COMPILER_CC[@]|$(CC)|g' \
```
jdinan: I think you need to replace ```$(CC)``` with ```$(WRAPPER_COMPILER_CC)```.

davidozog: Haha - oops again.  Definitely.

PR #315, opened 2017-01-23
Move/add barriers in shmem_free and shmem_realloc

davidozog:
Moving the exit barrier call in shmem_free to the entry eliminates a nasty race condition.  Similarly, shmem_realloc needs a barrier at entry.

### DIFF
--- a/src/symmetric_heap_c.c
+++ b/src/symmetric_heap_c.c
@@ -293,11 +293,11 @@ shmem_free(void *ptr)
     SHMEM_ERR_CHECK_INITIALIZED();
     SHMEM_ERR_CHECK_SYMMETRIC_HEAP(ptr);
 
+    shmem_internal_barrier_all();
+
     SHMEM_MUTEX_LOCK(shmem_internal_mutex_alloc);
     dlfree(ptr);
     SHMEM_MUTEX_UNLOCK(shmem_internal_mutex_alloc);
-
-    shmem_internal_barrier_all();
 }
 
 
@@ -307,6 +307,8 @@ shmem_realloc(void *ptr, size_t size)
     void *ret;
 
     SHMEM_ERR_CHECK_INITIALIZED();
+
+    shmem_internal_barrier_all();
 
     SHMEM_MUTEX_LOCK(shmem_internal_mutex_alloc);
     ret = dlrealloc(ptr, size);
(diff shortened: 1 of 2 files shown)


jdinan:
This is weird ... some tests are failing during transport finalization with an error freeing the target counter:
```
=  ../../src/transport_ofi.c:1231: 
=  Target CT close failed (Operation now in progress)../../src/transport_ofi.c:1242: 
=  Domain close failed (Operation now in progress)../../src/transport_ofi.c:1231: 
=  Target CT close failed (Operation now in progress)../../src/transport_ofi.c:1242: 
=  Domain close failed (Operation now in progress)
```

jdinan:
Appears to have been caused by a recent change to libfabric: https://github.com/ofiwg/libfabric/commit/ec8936f1367de720425c45d58aecd5ef9e00b2f4

@shefty -- Any idea whose bug, ours our yours?

shefty:
That commit was added to catch out of order destruction, which could lead to application crashes (accessing invalid memory).  It's possible I wrote the patch wrong.  :)  Can you check your destruction ordering and ensure that all EPs are closed prior to closing the associated counters?

shefty:
A (very) quick inspection of the shmem code looked okay.  I will look at the ofi code.  It may not be handling mr to counter bindings correctly.  Regardless, I'll help research this more.

shefty:
After a more significant look at the shmem code (transport_ofi.c), I still don't see an issue.  Nor do I see anything in the ofi code that looks incorrect.  These tell me that I'm obviously overlooking the problem...

jdinan:
I'm scratching my head, too.  I confirmed that the failure occurs with the SOS master starting with libfabric commit ec8936f1.

jdinan:
The ref count on the target counter is -1 when we are trying to free it.

jdinan:
It looks like the problem is here in ```sock_rx_ctx_close```:
```
    if (rx_ctx->comp.rem_read_cntr)
        sock_cntr_remove_rx_ctx(rx_ctx->comp.rem_read_cntr, rx_ctx);

    if (rx_ctx->comp.rem_write_cntr)
        sock_cntr_remove_rx_ctx(rx_ctx->comp.rem_write_cntr, rx_ctx);
```
The same remote counter is being removed twice; however in ```sock_ctx_bind_cntr``` the counter is only added once, in spite of potentially having been referenced three times (for FI_READ, FI_REMOTE_READ, and FI_REMOTE_WRITE events).

Proposed fix posted at: https://github.com/ofiwg/libfabric/pull/2671

shefty:
Thanks for finding this.  Apparently the code was more broken than I thought.  It's a wonder that this didn't lead to crashes found earlier in development.

PR #321, opened 2017-02-03
Add Cray test suite to Travis

davidozog:
- The script in `scripts/cray-tests.sh` can be run independently or as a part of Travis CI.  
- Most tests currently pass, but a couple don't, presumably because of bugs in the "tree" and "recursive doubling" reduction algorithms.  It seems like they might be with weird edge cases, for example, `shmem_int_and_to_all` when `PE_Size` is 1.
- Note the NULL pointer check when `--enable-error-checking` is set, which is required for a few Cray tests to pass, because they call `shmem_free(NULL)`.
- In Travis, this commit currently clones my [openshmem branch](https://github.com/davidozog/tests-cray/tree/openshmem) of [tests-cray](https://github.com/openshmem-org/tests-cray), but I'll submit a pull request there as well.

### DIFF
--- a/scripts/cray_tests.sh
+++ b/scripts/cray_tests.sh
@@ -6,14 +6,14 @@ export CC=oshcc
 export CXX=oshCC
 export FTN=oshfort
 export LAUNCHER=oshrun
-export CFLAGS="-DOPENSHMEM"
-export FFLAGS="-DOPENSHMEM -DOPENSHMEM_FORT_SHORT_HEADER -fcray-pointer"
+export CFLAGS="-DQUICK_TEST -DOPENSHMEM"
+export FFLAGS="-DQUICK_TEST -DOPENSHMEM -DOPENSHMEM_FORT_SHORT_HEADER -fcray-pointer"
 export NPES=4
 
 cd $CRAY_TESTS_DIR
 source configure.sh
-make sma1 MAKE_FLAGS=$TRAVIS_PAR_MAKE 
-make sma2 MAKE_FLAGS=$TRAVIS_PAR_MAKE 
+make sma1 MAKE_FLAGS="$TRAVIS_PAR_MAKE"
+make sma2 MAKE_FLAGS="$TRAVIS_PAR_MAKE"
 source $CRAY_TESTS_DIR/sma1/sma1_run 2>&1 | tee cray-tests-sma1.log
 source $CRAY_TESTS_DIR/sma2/sma2_run 2>&1 | tee cray-tests-sma2.log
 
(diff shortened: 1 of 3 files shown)


kseager:
just to follow up on your comment about the collectives: recursive doubling has a case specifically for npes=1 that is just doing a memcpy so I'm surprised to hear that's not working. It might be worth verifying what is going wrong there

davidozog:
Sorry @kseager, I should have been more specific.  I only see the npes=1 deadlock problem with SMA_BCAST_ALGORITHM=tree and SMA_REDUCE_ALGORITHM=tree.  This occurs in [the `all_and_int_ext` test](https://github.com/davidozog/tests-cray/blob/pr/openshmem/sma1/all_and_int_ext.c#L186) - when the source and target buffer pointers are the same, I believe.  I could definitely be wrong about that npes=1 claim, but it looks to me like the tree-based code path could use a memcpy that's similar to the recursive doubling?

The recursive doubling error looks to be a more complicated race that I so far do not understand.  I can reproduce it with SMA_REDUCE_ALGORITHM=recdbl on that same `all_and_int_ext` test.  Unfortunately it's a non-deterministic error.  Any ideas what could be causing that?

kseager:
ah, nice test, overlapping buffers. I would say this isn't getting handled correctly in the collective algorithm; they aren't uniformly handling this case. Tree is not doing a memcpy, its going along its merry way creating a tree and going through OFI to complete the transfer. 

As for recdbl memcpy is a bad choice if buffers overlap.

scripts/cray_tests.sh
```diff
@@ -0,0 +1,45 @@
...
+source $CRAY_TESTS_DIR/sma1/sma1_run 2>&1 | tee cray-tests-sma1.log
+source $CRAY_TESTS_DIR/sma2/sma2_run 2>&1 | tee cray-tests-sma2.log
+
+# Increment by 1 when a set of test fails (max is 3 for C/C++/Fortran)
+tests_failed=0
+
+#Check for failures in the C/C++ tests
+if grep "FAIL" cray-tests-sma1.log; then ((tests_failed+=1)); fi
+if grep "FAIL" cray-tests-sma2.log; then ((tests_failed+=1)); fi
+if [ ! $SOS_DISABLE_FORTRAN ]; then
```
jdinan: It might be safer to test ```-z "$SOS_DISABLE_FORTRAN"``` since we aren't actually setting the variable when Fortran is enabled.  Actually, this may be a bug in ```.travis.yml```...

jdinan:
Nice work!  Apart from a minor comment, this looks ready to me.  Let me know when you're ready to merge.

jdinan:
Looks good -- merging.  @davidozog Could you create issues to track the failures identified by the Cray tests?  Thanks!

PR #322, opened 2017-02-10
perf tests: enable 1 PE on bw tests

kseager:
-only bw tests (bibw will still abort if not even PE set)
-verified change on atomic/get/put bw tests


jdinan:
Looks ok to me. :+1:

PR #323, opened 2017-02-14
Add transport hook for spinlock body

jdinan:
OFI, check the CQ for errors while polling.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -63,6 +63,11 @@ AC_ARG_ENABLE([completion-polling],
 AS_IF([test "$enable_completion_polling" = "yes"],
       [AC_DEFINE([ENABLE_COMPLETION_POLLING], [1], [Enable polling for local/remote completions])])
 
+AC_ARG_ENABLE([progress-assist],
+    [AC_HELP_STRING([--enable-progress-assist],
+                    [Enable progress assistance in hard polling loop. (default:disabled)])])
+AS_IF([test "$enable_progress_assist" = "yes"], [AC_DEFINE([ENABLE_PROGRESS_ASSIST], [1], [Enable progress assistance])])
+
 AC_ARG_ENABLE([nonblocking-fence],
     [AC_HELP_STRING([--enable-nonblocking-fence],
                     [When total data ordering is not available, make shmem_fence a non-blocking operation.  shmem_fence will return immediately, but the next communication call will block until all previous communications finish. (default:disabled)])])
--- a/src/shmem_synchronization.h
+++ b/src/shmem_synchronization.h
@@ -83,7 +83,10 @@ shmem_internal_fence(void)
 
 #define SHMEM_WAIT(var, value)                           \
     do {                                                 \
-        while (*(var) == value) { SPINLOCK_BODY(); }     \
+        while (*(var) == value) {                        \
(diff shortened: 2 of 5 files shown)


jdinan:
Feedback: @hppritcha @kseager

kseager:
We would additionally need to add to ep_bind " | FI_RECV" to notify the provider we are now looking for inbound transfer cq error reporting.  As you mentioned earlier, since we are looking at inbound error checking we could add the ret + cq_error checking to received_cntr_wait , mirroring our local completion usage. 

In terms of the hard-polling pathway I would personally prefer someway to turn off this feature in order to preserve a pathway that doesn't call into the fabric.

jdinan:
Oh ... FI_RECV may interfere with other code that assumes the CQ will only contain TX errors.  What about splitting the CQ endpoint into separate RX and TX endpoints?  RX goes into the AV and TX endpoints don't.  This seems like cleaner software engineering, but is it less efficient from an OFI perspective?

jdinan:
So, we set ```FI_SELECTIVE_COMPLETION | FI_TRANSMIT``` on the counter-only TX EP to catch transmit errors.  This is interesting, given that we use RMA operations, which can never set ```FI_COMPLETE``` since there's no ```flags``` argument.  I guess I could use this same trick at the RX EP to set ```FI_SELECTIVE_COMPLETION | FI_RECV``` so that we collect RX errors.  Although, we already set ```FI_TRANSMIT``` when we bind this CQ to that EP.  Can I bind the same CQ again with different flags?

BTW, we seem to have lost the event slots tracking code from the Portals transport that ensure we have EQ space available before issuing ops that generate completion events (bounce buffered puts).  Should we log a ticket to fix this?

kseager:
Good point about the API calls missing flags; MPI must've not run into this limitation 😉. For our purposes we don't need it for getting error events and want to suppress the successful completion anyways. 

Yes its the same trick on the RX side. Its seems like a waste to create two CQs per process just for splitting up the tx/rx error handling. In theory we check the queue right after the potentially erroneous operation so we should be getting the associated error event. The ret value of the operation should signal if there is an error to check. This code addition is the only time we are checking for an error via the cq vs. checking the return first.  

If its the same associate ep2eq then we should just OR on the extra flag like this:

     ret = fi_ep_bind(shmem_transport_ofi_cntr_epfd,
-                   &shmem_transport_ofi_put_nb_cqfd->fid, FI_SELECTIVE_COMPLETION | FI_TRANSMIT);
+                   &shmem_transport_ofi_put_nb_cqfd->fid, FI_SELECTIVE_COMPLETION | FI_TRANSMIT | FI_RECV );

Sure log and issue for the Portals code.

jdinan:
PRK strikes again.  They changed the compiler flags upstream causing a build failure.  @davidozog will push a fix.

hppritcha:
bot:lanl:retest

hppritcha:
bot:lanl:retest

hppritcha:
okay try again
bot:lanl:retest

hppritcha:
bot:lanl:retest

jdinan:
@hppritcha says that the GNI provider no longer needs this.  Closing.

Issue #138, opened 2016-03-01
External component locations (e.g. PMI) not passed to wrappers


tonycurtis:
My configure stanza passes LDFLAGS settings (-L and -rpath) for the location of libfabric and PMI libraries.

These don't get passed on to oshcc and friends, meaning use of LD_LIBRARY_PATH in my login environment required to find these libraries at execution time (I am allergic to LD_LIBRARY_PATH).

hppritcha:
this is an annoying thing.  taking this issue.

jdinan:
@davidozog Can we close this issue?

davidozog:
@jdinan Yes, I'd say so.

jdinan:
Fixed in #327

PR #327, opened 2017-02-14
Upstream OPAL update

davidozog:
This PR would update the SOS build setup to use the upstream (OMPI) check for libfabric, while still supporting the `--with-ofi` flag.  It also updates to some other upstream OPAL m4 scripts, such as check_package, check_withdir, and a setup wrapper.  This wrapper does a good job of addressing Issue #138, by explicitly setting rpath to include all external library directories.  The SOS configure then passes that rpath string to oshcc and friends via LDFLAGS.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -24,8 +24,6 @@ m4_ifndef([AS_VAR_COPY],
     [m4_define([AS_VAR_COPY],
        [AS_LITERAL_IF([$1[]$2], [$1=$$2], [eval $1=\$$2])])])
 
-OPAL_TOP_BUILDDIR="`pwd`"
-
 dnl Work around change in how the rules were implemented between
 dnl 1.11.2 and 1.11.3.
 AM_SILENT_RULES([yes])
@@ -455,6 +453,7 @@ LIBS="$LIBS $portals4_LIBS $libfabric_LIBS $ofi_LIBS $XPMEM_LIBS $pmi_LIBS"
 # Need the libtool binary before setting up rpath
 LT_OUTPUT
 
+OPAL_TOP_BUILDDIR="`pwd`"
 OPAL_SETUP_RPATH
 
 # wrapper compiler gorp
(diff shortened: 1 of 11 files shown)


configure.ac
```diff
@@ -24,6 +24,8 @@ m4_ifndef([AS_VAR_COPY],
     [m4_define([AS_VAR_COPY],
        [AS_LITERAL_IF([$1[]$2], [$1=$$2], [eval $1=\$$2])])])
 
+OPAL_TOP_BUILDDIR="`pwd`"
```
jdinan: What's this for?  I don't see it referenced anywhere else?

configure.ac
```diff
@@ -234,13 +236,22 @@ OMPI_CHECK_PORTALS4([portals4],
...
-OMPI_CHECK_OFI([ofi],
+OPAL_CHECK_LIBFABRIC([libfabric],
     [AC_DEFINE([USE_OFI], [1], [Define if OFI transport active])
      transport="ofi"
      transport_ofi="yes"],
     [transport_ofi="no"])
 AM_CONDITIONAL([USE_OFI], [test "$transport_ofi" = "yes"])
 
+if test "$transport_ofi" = "no" ; then
+  OMPI_CHECK_OFI([ofi],
```
jdinan: Oh, ick.  It would be nice to get rid of this old macro.  Could you modify the ```AC_HELP_STRING``` in ```ompi_check_ofi.m4``` and prefix the message with "(deprecated)"?

jdinan:
This looks very reasonable.  Just a couple comments for you to consider before we merge.

configure.ac
davidozog: OPAL_SETUP_RPATH needs this to be set to the build path because it calls `libtool --config` just before the rpath parsing stuff (OMPI does the same thing in [their configure.ac file](https://github.com/open-mpi/ompi/blob/master/configure.ac#L185)).  Maybe this line should be moved right before the OPAL_SETUP_RPATH call in our configure.ac to make it seem less out of place?

jdinan: Sure, moving it is a good idea.

jdinan: Oh, I see where it's used now.  Github was hiding part of the diff.

jdinan:
Looks good! /crosses fingers/

PR #331, opened 2017-02-18
Travis: Add -std=c99 to PRK compiler flags

jdinan:


davidozog:
Yup - too bad there's no CFLAGS caught, but this will do.  Thanks.

jdinan:
I thought I had a fix for some of the tree-based collectives that I wanted
to test out.  Looks like that's going to take a bit more work.

Issue #325, opened 2017-02-14
SMA_BCAST_ALGORITHM=tree deadlock when PE_Size == 1


davidozog:
The Cray reduction tests deadlock when SMA_BCAST_ALGORITHM=tree.  For example, [this program](https://github.com/openshmem-org/tests-cray/blob/master/sma1/all_and_int_ext.c).

**Possible solution:**  An appropriately placed `memcpy(target, source, ...); return;` when PE_Size==1.  I've tried this to no avail in `shmem_internal_op_to_all_tree()`.

PR #332, opened 2017-02-18
Fix hang in bcast tree algorithm

jdinan:
For single PE calls, the root PE was falling into the leaf PE code and
waiting for a broadcast message that would never arrive.


jdinan:
Fixes #325

jdinan:
Needs work...

Issue #326, opened 2017-02-14
SMA_REDUCE_ALGORITHM=recdbl error


davidozog:
The Cray reduction tests occasionally result in incorrect buffers when SMA_REDUCE_ALGORITHM=recdbl. For example, [this program](https://github.com/openshmem-org/tests-cray/blob/master/sma1/all_and_int_ext.c).

Some observations:

- I can reproduce this bug with 3 PEs or more on my laptop.
- The error is non-deterministic, with a higher rate of failure with more PEs.
- The test iterates over all reasonable values for (`PE_start, logPE_stride,` and `PE_Size`).  For a given number of PEs at launch, failures occur _consistently_ for various 3-tuple values.  For example, with (`PE_start=2, logPE_stride=0, PE_size=2`), and SHMEM_N_PES=4, failure rate is quite high, but so far there are zero failures for other configurations, such as (1, 0, 3).

jdinan:
This will be an error in the newer ```shmem_internal_op_to_all_recdbl_sw``` collective that we fall back on when the provider doesn't have atomic support.  I don't think we have been thorough about testing the active set arguments with this routine.  That's the first place I would look.

PR #335, opened 2017-02-24
Fix bugs in tree-based bcast and barrier

jdinan:
This fixes several bugs, but seems to have also caused a bug in the Cray ```shmem_broadcast_all.c``` test to fail nearly every time.

### DIFF
--- a/src/collectives.c
+++ b/src/collectives.c
@@ -501,6 +501,8 @@ shmem_internal_bcast_tree(void *target, const void *source, size_t len,
         int i;
 
         if (parent != shmem_internal_my_pe) {
+            send_buf = target;
+
             /* wait for data arrival message if not the root */
             SHMEM_WAIT(pSync, 0);
 
--- a/src/collectives.c
+++ b/src/collectives.c
@@ -645,6 +645,11 @@ shmem_internal_op_to_all_tree(void *target, const void *source, int count, int t
     /* need 2 slots, plus bcast */
     shmem_internal_assert(SHMEM_REDUCE_SYNC_SIZE >= 2 + SHMEM_BCAST_SYNC_SIZE);
 
+    if (PE_size == 1) {
+        memcpy(target, source, type_size*count);
+        return;
+    }
+
     if (PE_size == shmem_internal_num_pes) {
         /* we're the full tree, use the binomial tree */
         parent = full_tree_parent;
--- a/src/collectives.c
+++ b/src/collectives.c
@@ -717,14 +717,13 @@ shmem_internal_op_to_all_recdbl_sw(void *target, const void *source, int count, 
 {
     int stride = 1 << logPE_stride;
     int my_id = ((shmem_internal_my_pe - PE_start) / stride);
-    long one = 1, neg_one = -1, buff = 0;
     int log2_proc = 1, pow2_proc = 2;
     int i = PE_size >> 1;
     int wrk_size = type_size*count;
     void * const current_target = malloc(wrk_size);
-    int peer = 0;
     long completion = 0;
     long * pSync_extra_peer = pSync + SHMEM_REDUCE_SYNC_SIZE - 2;
+    const long ps_target_ready = 1, ps_data_ready = 2;
 
     if (PE_size == 1) {
         memcpy(target, source, type_size*count);
@@ -745,7 +744,8 @@ shmem_internal_op_to_all_recdbl_sw(void *target, const void *source, int count, 
     if (current_target)
         memcpy(current_target, (void *) source, wrk_size);
     else
-        RAISE_ERROR(1);
+        RAISE_ERROR_MSG("Failed to allocate current_target (count=%d, type_size=%d, size=%dB)\n",
+                        count, type_size, wrk_size);
 
     /* Algorithm: reduce N number of PE's into a power of two recursive
      * doubling algorithm have extra_peers do the operation with one of the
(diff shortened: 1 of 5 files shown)


jdinan:
@davidozog Here's a sample failure from the Cray test:
```
FAIL: NPES: 4 shmem_broadcast_all
Sandia OpenSHMEM 1.3.1
SMA_VERSION             Not set
	If set, print library version at startup
SMA_INFO                Set
	If set, print this help message at startup
SMA_SYMMETRIC_SIZE      536870912
	Symmetric heap size
SMA_SYMMETRIC_HEAP_USE_MALLOC Not set
	If set, allocate the symmetric heap using malloc
SMA_SYMMETRIC_HEAP_USE_HUGE_PAGES No
	Symmetric heap use large pages
SMA_COLL_CROSSOVER      4
	Cross-over between linear and tree collectives
SMA_COLL_RADIX          4
	Radix for tree-based collectives
SMA_BOUNCE_SIZE         2048
	Maximum message size to bounce buffer
SMA_BARRIER_ALGORITHM   AUTO
	Algorithm for barrier.  Options are auto, linear, tree, dissem
SMA_BCAST_ALGORITHM     AUTO
	Algorithm for broadcast.  Options are auto, linear, tree
SMA_REDUCE_ALGORITHM    AUTO
	Algorithm for reductions.  Options are auto, linear, tree, recdbl
SMA_COLLECT_ALGORITHM   AUTO
	Algorithm for collect.  Options are auto, linear
SMA_FCOLLECT_ALGORITHM  AUTO
	Algorithm for fcollect.  Options are auto, linear, ring, recdbl
SMA_DEBUG               Off
	Enable debugging messages
SMA_TRAP_ON_ABORT       Off
	Generate trap if the program aborts or calls shmem_global_exit
Network transport:      OFI
SMA_OFI_PROVIDER        sockets
	Provider that should be used by the OFI transport
SMA_OFI_FABRIC          AUTO
	Fabric that should be used by the OFI transport
SMA_OFI_DOMAIN          AUTO
	Fabric domain that should be used by the OFI transport
SMA_OFI_ATOMIC_CHECKS_WARN Not set
	Display warnings about unsupported atomic operations
shmem_broadcast(/home/travis/travis/src/tests-cray/testbin/build/bin/sma1/shmem_broadcast_all) n_pes=4
shmem_broadcast32             max_elements = 10
===================================================================================
=   PID 4838 RUNNING AT testing-gce-f5903183-f9d3-4ea5-8b7a-65278abda2e7
=   EXIT CODE: 139
=   CLEANING UP REMAINING PROCESSES
=   YOU CAN IGNORE THE BELOW CLEANUP MESSAGES
===================================================================================
YOUR APPLICATION TERMINATED WITH THE EXIT STRING: Segmentation fault (signal 11)
This typically refers to a problem with your application.
Please see the FAQ page for debugging suggestions
```

src/collectives.c
```diff
@@ -59,10 +59,30 @@ shmem_internal_build_kary_tree(int radix, int PE_start, int stride,
     for (i = 1 ; i <= radix ; ++i) {
         int tmp = radix * my_id + i;
         if (tmp < PE_size) {
-            children[(*num_children)++] = (PE_start + ((tmp + PE_root) * stride)) % (PE_size * stride);
+            children[(*num_children)++] = (PE_start + ((tmp + PE_root) * stride));
```
jdinan: This change fixed one bug and introduced another new one.  Re-working it...

jdinan:
Closes #325 and #326.

jdinan:
Woo hoo!!  The board is green!!!

jdinan:
@kseager @davidozog Could you please review this today?  I'd like to get out a release candidate ASAP.

src/collectives.c
```diff
@@ -675,149 +709,134 @@ shmem_internal_op_to_all_tree(void *target, const void *source, int count, int t
...
-         shmem_internal_put_small(pSync_extra_peer, &buff,
-                                  sizeof(long), peer);
-         shmem_internal_fence();
-      }
+    } else {
+        if (my_id < PE_size - pow2_proc) {
+            int peer = (my_id + pow2_proc) * stride + PE_start;
+            shmem_internal_put_small(pSync_extra_peer, &ps_target_ready, sizeof(long), peer);
+
+            SHMEM_WAIT_UNTIL(pSync_extra_peer, SHMEM_CMP_EQ, ps_data_ready);
```
kseager: nice catch

jdinan: Sorry for the churn in the recdbl_sw code.  I reformatted the code as a separate commit.  If you want to view just the changes, take a look at 320e834.

src/collectives.c
```diff
@@ -759,80 +759,75 @@ shmem_internal_op_to_all_recdbl_sw(void *target, const void *source, int count,
...
-        long *step_psync;
-
         for (i = 0; i < log2_proc; i++) {
+            long *step_psync = &pSync[i];
+            int peer = (my_id ^ (1 << i)) * stride + PE_start;
 
-            peer = (my_id ^ (1 << i)) * stride + PE_start;
-            step_psync = &pSync[i];
+            if (shmem_internal_my_pe < peer) {
+                shmem_internal_put_small(step_psync, &ps_target_ready,
```
kseager: do we need this sync for every iteration? I agree that at the first iteration we want to avoid corrupting source during the memcpy if we put too quickly on "target" when buffers are the same/overlap.  But what about the rest?

jdinan: This isn't related to the source == target case.  We need this sync in every iteration to avoid overwriting the target buffer before the remote PE has consumed the data from the previous iteration (including in the first iteration to avoid overwriting data sent by one of the "extra" PEs).

kseager: nevermind I misread the diff, yes you are right thanks

src/collectives.c
davidozog: That helps a lot, thanks!

kseager:
👍 thanks Jim this looks great

PR #342, opened 2017-03-01
Do not request OFI mr_key_size > 0 if not needed

ledettwy:
The OFI transport was requesting the maximum available mr_key_size that
OFI supports (value returned in domain_attr->mr_key_size from
fi_getinfo), when depending on compile-time options it may only need a
key size of 1 or 0.  Logic was added to request the appropriate value
depending on these compile time options.

Signed-off-by: Lisanna Dettwyler <levi.e.dettwyler@intel.com>

### DIFF
--- a/src/transport_ofi.c
+++ b/src/transport_ofi.c
@@ -990,6 +990,10 @@ static inline int query_for_fabric(struct fabric_info *info)
 #else
     domain_attr.mr_mode       = FI_MR_BASIC; /* VA space is pre-allocated */
 #endif
+#if !defined(ENABLE_MR_SCALABLE) || !defined(ENABLE_REMOTE_VIRTUAL_ADDRESSING)
+    domain_attr.mr_key_size   = 1; /* Heap and data use different MR keys, need
+                                      at least 1 byte */
+#endif
     domain_attr.threading     = FI_THREAD_ENDPOINT; /* we promise to serialize access
                                                        to endpoints. we have only one
                                                        thread active at a time */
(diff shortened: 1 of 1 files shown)


lanl-ompi:

jdinan:
Thanks for the patch!  Is it necessary to set ```mr_key_size``` in the domain attributes, given that we're providing the value we want in the fabric query?

sayantansur:
I agree, it is better to just provide the hint once, and let the provider give you want it wants. It seems like we are querying for 1B, but then potentially later changing to 0? Why don't we ask for 0B in the first place (if possible).

jdinan:
```domain_attr``` is initialized to all zeroes, so we should be giving the same hint in both places.

jdinan:
@ledettwy What do you think about removing the second block of code (lines 1057-1064)?

ledettwy:
I don't think that's in line with the OFI spec. fi_getinfo's man page states the following for domain_attr:

`Optionally supplied domain attributes. Domain attributes may be specified and returned as part of fi_getinfo. When provided as hints, requested values of struct fi_domain_attr should be set. On output, the actual domain attributes that can be provided will be returned. Output values will be greater than or equal to requested input values. See fi_domain(3) for details.`

fi_getinfo, as its name implies, is a query function. It is a query of what sorts of providers are available and what their abilities are. It's not a way for the application to tell OFI what it wants. That happens at the actual initialization time, and it's why you pass in another fi_info struct to fi_fabric and fi_domain. In general an application should not request more resources than it needs.

@sayantansur If SOS is compiled such that it only uses 1 MR, then it leaves mr_key_size 0, but if it needs multiple MRs, then it sets mr_key_size to 1 in the query, to make sure it only gets providers that are capable of providing at least a mr_key_size of 1. Later in the fi_domain step, we know for sure that we have gotten a provider that will meet our needs, so again just have a compile-time switch to determine what those needs are. This way SOS will not request more resources than it needs from the particular OFI provider it has selected.

jdinan:
Ok, thanks for the discussion!

PR #345, opened 2017-03-04
Fix distribution build

jdinan:


jdinan:
@davidozog Could you review this morning?  I have another PR that I'll post after this one that requires careful review.

The gist of this change is to exclude generated files from the distribution tarball and to ensure that the sources of generated files are included in the distribution tarball.  Autoconf includes all C source files by default and excludes everything else, so there's some extra work needed to ensure everything gets packaged up properly.

PR #346, opened 2017-03-06
Pull Simple PMI into SOS Build

jdinan:
Pull the PMI sources into the main SOS build instead of building them as a subconfigure.  There are several benefits, including:
 * The build is simpler and faster (it runs configure once instead of twice)
 * The distribution tarball is much cleaner
 * Since the subconfigure is gone, the SOS configure script is now able to give errors and warnings on unrecognized arguments, e.g.
```
$ ../configure --invalid-option
configure: error: unrecognized option: `--invalid-option'
Try `../configure --help' for more information
```
```
$ ../configure --enable-invalid-option
...
configure: WARNING: unrecognized options: --enable-invalid-option
```

SOS had also been including the simple PMI files in libsma.la.  As a result, it was building libshmem_pmi.la but not actually using it.  I fixed this by removing the PMI objects from libsma and in the process renamed the PMI library to libpmi_simple.la for consistency.  I added libpmi_simple to the compiler wrappers and also to the rules for building the unit tests.  In the process, I simplified the build rules for the tests, which allowed us to delete a lot of junk from the testing makefiles.


jdinan:
@davidozog and @kseager I'd appreciate a review from both of you on this change.  Hopefully this should resolve the last of the distribution build problems that Matias ran into.

jdinan:
Looks like mpich.org is back online again.  Restarting Travis..

test/unit/Makefile.am
```diff
@@ -128,255 +128,33 @@ LOG_COMPILER = $(TEST_RUNNER)
 AM_CPPFLAGS = -I$(top_builddir)/mpp
 AM_FCFLAGS = -I$(top_builddir)/mpp
 AM_LDFLAGS = $(LIBTOOL_WRAPPER_LDFLAGS)
+LDADD = $(top_builddir)/src/libsma.la
 
-pi_SOURCES = pi.c
-pi_LDADD = $(top_builddir)/src/libsma.la
```
jdinan: Sure.  Pushed a fix for this and a few other things I found along the way.

kseager:
question, why not just build simple-pmi into libsma and not create the separate pmi library? Won't you end up adding simple-pmi to every libsma dependency anyways? I can't think of a use-case for their separation. ..also its nice to see the refactoring of Makefile.am for the test suites

jdinan:
I thought about this, too.  I don't have a very strong opinion, but I think it's better software packaging to build PMI as a separate library since it's an external component in most of our builds and not technically a part of SOS.  MPICH may also soon move PMI into a separate library (https://github.com/pmodels/mpich/pull/2550), so keeping PMI external might help down the road.

kseager:
sure, I'm good with this PR

PR #351, opened 2017-03-09
Reduce default table size in GUPS

jdinan:
The verification phase repeats the GUPS experiment with each update performed while holding a mutex.  This part is really slow...


jdinan:
@davidozog Please review.  @kseager Heads-up on the FIXME that I added.

PR #360, opened 2017-03-16
Add --disable-bounce-buffers configure option

jdinan:


jdinan:
@davidozog Could you please review?

PR #367, opened 2017-03-25
Perf put target

kseager:
@jdinan 

-measures the "target" side bw/mr for put at 8bytes & 4KB and compares it to initator 
-look at last commit for put target implementation in the target_put.h file
-enabled via -t option through shmem_bw_put_perf

sample output:
[target_sample.txt](https://github.com/Sandia-OpenSHMEM/SOS/files/869549/target_sample.txt)

### DIFF
--- a/test/performance/shmem_perf_suite/Makefile.am
+++ b/test/performance/shmem_perf_suite/Makefile.am
@@ -30,7 +30,8 @@ noinst_HEADERS = \
 	int_element_latency.h \
 	bw_common.h \
 	uni_dir.h \
-	bi_dir.h
+	bi_dir.h \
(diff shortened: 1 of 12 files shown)


jdinan:
What's the reason for using ```uint32_t``` for PE id?  PE id is an ```int```, which may lose precision when you convert to ```uint32_t```.

jdinan:
Does this update add support for using the first M PEs as initiators and the remaining PEs as targets?  IIUC, we wanted to support something like a 2 node run, with M PEs on node 0 and N-M on node 1 and take the aggregate bandwidth across all initiators and all targets.

PR #375, opened 2017-04-04
Add shmemx_pcontrol extension

jdinan:
Add function to allow application to interact with profiling tools
(c.f. MPI_Pcontrol).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,7 +68,8 @@ before_install:
     ## Run the icc installation script:
     - >
       if [ "$CC" = "icc" ]; then
-          if [ "$TRAVIS_REPO_SLUG" = "Sandia-OpenSHMEM/SOS" ]; then
+          if [ "$TRAVIS_REPO_SLUG" = "Sandia-OpenSHMEM/SOS" -a \
+               "$TRAVIS_PULL_REQUEST" = "false"]; then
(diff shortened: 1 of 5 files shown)


davidozog:
PR #377 should fix the icc installation problem you're seeing in Travis.

jdinan:
Feedback from MPI Forum: "const" qualifier on level in ```MPI_Pcontrol``` looks like a transcription error and is not necessary.

Feedback from @naveen-rn: Do we want this function to have a return value?  All MPI routines return an error code; however, SHMEM routines do not.

naveen-rn:
I just saw your question on the MPI Forum, it looks removing "const" seems to be valid and we don't have Fortran to worry about;) AFA the return value - Unless there are some specific request from the tools group(nothing yet here internally), I would suggest to go without any return value for now and add later when we have a solid error reporting framework.

naveen-rn:
@jdinan I spoke with the internal Tools time and it seems they don't actually see huge usage from the users on MPI_Pcontrol. Bit curious, do you have any expected usage/stats that you are expecting from this routine.

jdinan:
I don't personally have any expected usage for this function, but some tools folks asked me to implement it and it seems like a generally useful thing to have.  For example, the NERSC IPM tool uses it (https://github.com/nerscadmin/IPM).

PR #377, opened 2017-04-05
Only run icc in Travis CI if not from a Git fork

davidozog:
Because encrypted variables are not available to forks, only test the
icc compiler when TRAVIS_REPO_SLUG matches the upstream
value, 'Sandia-OpenSHMEM/SOS', otherwise exit with a "pass".

Note that we could also default to a "fail" for icc on forks.  But my (slight) preference is to assume it passes, then make sure the icc tests pass in Travis after we merge a PR into 'Sandia-OpenSHMEM/SOS'.  The encrypted variables shouldn't be a problem for that merge commit.

Signed-off-by: David Ozog <david.m.ozog@intel.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,8 +66,16 @@ before_install:
     ## Disable security protection so CMA will work
     - sudo sysctl -w kernel.yama.ptrace_scope=0
     ## Run the icc installation script:
-    - ./scripts/install-icc.sh --components icc,ifort
-    - source ~/.bashrc
+    - >
+      if [ "$CC" = "icc" ]; then
+          if [ "$TRAVIS_REPO_SLUG" = "Sandia-OpenSHMEM/SOS" ]; then
+              ./scripts/install-icc.sh --components icc,ifort
+              source ~/.bashrc
+          else
+              echo "Cannot test Intel compiler on a fork.  Exiting now."
+              exit 0
+          fi
+      fi
     ## Build libev
     - cd $TRAVIS_SRC
     - wget http://dist.schmorp.de/libev/Attic/libev-4.22.tar.gz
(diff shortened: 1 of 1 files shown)


jdinan:
Aww..  Does that mean we won't be able to test icc on PRs from forks?

Thanks for the fix -- I'll go ahead and merge.

davidozog:
Yes, forks will not work unless we do something wild and crazy, like share the license serial number on a case-by-case basis and tell people [how to generate the appropriate encrypted var](https://docs.travis-ci.com/user/environment-variables/#Encrypting-environment-variables), _**then**_ replace it with the original value when we merge into master.  Or even crazier, make the license number unencrypted ;)  @jeffhammond was certainly right when he warned that this might be a bit of a hassle.  It is really nice to have icc checked though...

jdinan:
Aww...  After we merge, Travis re-runs testing on master, so we'll just have to check on icc manually.  Still better than not testing it.  Thanks for slogging though this!

davidozog:
Exactly - icc will always be tested, we just won't know if there is a problem until _after_ the merge...

jeffhammond:
If you support ICC in the upstream Travis, it should be run as part of PRs, no?  You should know before the merge.

I never got it to work, but https://github.com/ParRes/Kernels/blob/master/.travis.yml#L37 shows **two** secure env vars, one for https://github.com/ParRes/Kernels and the other for https://github.com/jeffhammond/PRK, which I attempted to parse in https://github.com/ParRes/Kernels/blob/master/travis/install-intel.sh#L4.

Obviously, we have a reason to prefer ICC, but if the goal is testing a range of compilers, then https://github.com/nemequ/pgi-travis is an option that appears to avoid the license issues ICC has.

davidozog:
Yeah, I could be wrong about all this, especially since I now see that Jim's latest PR failed, as though his "slug" is the upstream one.  I also think your parsing approach above should work just fine... Do you see the "Activation does not work in silent mode..." warning when you try that?  :/

PR #387, opened 2017-04-20
Collect and All-to-All Sync Update

jdinan:
Update collect and all-to-all to use barrier instead of linear flags to complete the exchange.  Also update the distribution order to do the local process last to overlap local copy with remote writes.


jdinan:
SOS collect unit test failed with: ```0: Test 3 error, dst[0] == 0, expected 2```

jdinan:
I can reproduce the hang in the ```alltoalls``` test locally by setting ```SHMEM_BARRIER_ALGORITHM=dissem``` and running with 2 PEs.

jdinan:
Turns out that Mario was right; we do need a quiet before exiting dissemination barrier.

jdinan:
@davidozog This is ready for review.

Issue #103, opened 2016-01-21
Portals build problem with system installed libfabric


jhrobichaux:
When trying to build Portal4 transport on a system running CentOS Linux release 7.1.150, which has libfabric v1.1.1 installed by default, problems are encountered with trying to build the various 'osh*' commands. It appears that the configure script identifies libfabric is installed on the system and somehow tries to support both (see output of Transport stanza). However, in the resulting config.h only the OFI/libfabric  macros are defined.  

I am able to work around this by running configure with '--without-ofi', however wanted to bring this feature to your attention. 

Below are the salient output from the output of the original configure command, and make. Let me know if need more information. 

=== Config Errors ===
./configure --with-portals4=$BASE/portals4/  --disable-long-fortran-header --prefix=$BASE/sandia-shmem-portals4 \
 --enable-error-checking --enable-remote-virtual-addressing --enable-picky \
 --enable-pmi-simple FCFLAGS="-fcray-pointer" \
 CC='gcc' CFLAGS='-g -O0'

...
Looking for library in lib
checking for PtlLEAppend in -lportals... yes
checking --with-ofi value... simple ok (unspecified)
checking --with-ofi-libdir value... simple ok (unspecified)
checking rdma/fabric.h usability... yes
checking rdma/fabric.h presence... yes
checking for rdma/fabric.h... yes
looking for library without search path
checking for fi_getinfo in -lfabric... yes
...

Transports:
  **Portals 4:      yes**
  **OFI:            yes**
  XPMEM:          no
  CMA:            no
...

=== Make Errors ===
...
cp ./shmem_launcher_script.in oshrun
  CC       transport_portals4.lo
  SED      oshcc
  SED      oshc++
  SED      oshCC
  SED      oshfort
  CC       shmem_free_list.lo
  CC       runtime-pmi.lo
  CC       malloc.lo
  CC       init.lo
  CC       collectives.lo
  CC       init_c.lo
In file included from collectives.c:21:0:
shmem_internal_op.h: In function ‘shmem_internal_reduce_local’:
config.h:20:21: error: ‘FI_INT16’ undeclared (first use in this function)
 #define DTYPE_SHORT FI_INT16
                  ^
...

jdinan:
Thanks for the bug report.

It looks like we need to not enable OFI if the user has explicitly enabled Portals.  Presently, we don't support building for both at the same time.  We should also flag an error if we reach then end of the configuration and both are enabled, since it will cause a build failure.

[...]

PR #388, opened 2017-04-21
Remove "--with-ofi" from configure

davidozog:
The "--with-ofi" configure flag has been deprecated for about 2 months, and has caused recent confusion when building on certain cluster environments.  This commit removes the option completely and replaces it with the upstream OpenMPI option, "--with-libfabric", introduced in commit #327.

This commit also attempts to disallow detection of _both_ portals4 and libfabric for transport.  The same logic also applied to XPMEM versus CMA transports.  This should resolve issue #103.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -241,42 +241,44 @@ AC_SEARCH_LIBS([clock_gettime], [rt],
 dnl check for libraries
 
 OMPI_CHECK_PORTALS4([portals4],
-    [AC_DEFINE([USE_PORTALS4], [1], [Define if Portals 4 transport active])
-     transport="portals4"
+    [transport="portals4"
      transport_portals4="yes"],
     [transport_portals4="no"])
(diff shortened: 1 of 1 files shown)


hppritcha:
hmmm... looks like LANL-CI isn't happy with this.
Let's try again.
bot:lanl:retest

hppritcha:
funny, we switched to --with-ofi in the Open MPI project.

hppritcha:
bot:lanl:retest

hppritcha:
it looks like something's changed with write permissions to the SOS repo.

davidozog:
Hah... that's funny.  I didn't see that change, unfortunately.  My intent was to stay upstream with OMPI, so I'll update to the latest scripts, which means we'll actually go with --with-ofi.  That should nullify this PR...  Sorry about that!

hppritcha:
okay some funny business was going on with aws jenkins.  Let's try again:
bot:lanl:retest

davidozog:
I've replaced this PR with #390.

PR #389, opened 2017-04-23
Add shmemx BXOR support

agrippa:
This commit adds a non-fetching bitwise xor atomic to SoS as a shmemx extension. The change is pretty simple, and makes use of the existing shmem_internal_atomic_small API.


hppritcha:
okay I think we have jenkins fixed now.
bot:lanl:retest

mpp/shmemx_c_func.m4
```diff
@@ -18,3 +18,7 @@ void SHPRE()shmemx_ct_set(shmemx_ct_t ct, long value);
 void SHPRE()shmemx_ct_wait(shmemx_ct_t ct, long wait_for);
 
 void SHPRE()shmemx_pcontrol(int level, ...);
+
+void SHPRE()shmemx_int_bxor(int *target, int value, int pe);
+void SHPRE()shmemx_long_bxor(long *target, long value, int pe);
+void SHPRE()shmemx_longlong_bxor(long long *target, long long value, int pe);
```
jdinan: Almost.  Take a look at ```mpp/shmem_c_func.m4``` for inspiration.  You want to define the function prototype as an m4 macro and then ```SHMEM_DEFINE_FOR_AMO(...)```.

Another comment -- these aren't the right types.  You will also need to add a new type category in ```src/shmem_bind_c.m4``` for the bitwise AMO types.

agrippa: Gotcha. Before I do that, how do you define the right types for bitwise atomics (given there isn't one defined the OSHMEM spec)? Looking around for inspiration, it looks like the MPI spec says bxor works for any variant of integer or byte?

jdinan: They are listed in the OpenSHMEM proposal.

agrippa: Ah sorry, forgot you had a proposal out for bitwise atomics.

jdinan:
I think it's Nick's proposal. Should show up under pull requests on the
specification repo.

agrippa:
The latest push should be in line with Nick's latest draft at https://github.com/openshmem-org/specification/pull/31.

The only thing I wasn't sure about was how to translate the unsigned int, unsigned long, and unsigned long long types in Table 4 of https://github.com/openshmem-org/specification/files/839217/proposal-bitwise-amos-20170313.pdf into OFI datatypes. For example, should it be:

unsigned int -> FI_UINT32
unsigned long -> FI_UINT64
unsigned long long -> ?

Once we've got that sorted out, I could also add AND, OR, FETCH_AND, FETCH_OR, and FETCH_XOR to this pull request too. Sounds good?

jdinan:
In ```configure.ac```, take a look at:
```
dnl check for types

SHMEM_FIND_INT_TYPE([char], [$transport])
SHMEM_FIND_INT_TYPE([short], [$transport])
SHMEM_FIND_INT_TYPE([int], [$transport])
SHMEM_FIND_INT_TYPE([long], [$transport])
SHMEM_FIND_INT_TYPE([long long], [$transport])
```

I think we may need to add a ```SHMEM_FIND_UINT_TYPE``` macro so that it spits out the right types (since it only looks at the size).

agrippa:
Latest updates add in FETCH_XOR, AND, FETCH_AND, OR, FETCH_OR and the unsigned int, unsigned long, and unsigned long long data types.

Also added a basic unit test for these new APIs, though it could probably be cleaned up with some preprocessor magic.

jdinan:
Looks like you might be missing an include somewhere:

```
atomic_c.c: In function ‘shmemx_uint_atomic_xor’:
atomic_c.c:283:1: error: ‘SHM_INTERNAL_UINT’ undeclared (first use in this function)
atomic_c.c:283:1: note: each undeclared identifier is reported only once for each function it appears in
atomic_c.c: In function ‘shmemx_ulong_atomic_xor’:
atomic_c.c:284:1: error: ‘SHM_INTERNAL_ULONG’ undeclared (first use in this function)
atomic_c.c: In function ‘shmemx_ulonglong_atomic_xor’:
atomic_c.c:285:1: error: ‘SHM_INTERNAL_ULONG_LONG’ undeclared (first use in this function)
atomic_c.c: In function ‘shmemx_uint32_atomic_xor’:
atomic_c.c:288:1: error: ‘SHM_INTERNAL_UINT32’ undeclared (first use in this function)
atomic_c.c: In function ‘shmemx_uint64_atomic_xor’:
atomic_c.c:289:1: error: ‘SHM_INTERNAL_UINT64’ undeclared (first use in this function)
atomic_c.c: In function ‘shmemx_uint_atomic_and’:
...
```

agrippa:
The complex_reductions_f.f90 example doesn't compile on any Fortran compiler I have access to (and it seems like it shouldn't from reading the code) but whatever version of gfortran Travis is using doesn't seem to mind redundant definitions of shmem_my_pe/shmem_n_pes so reverted that change...

jdinan:
This is one of the many horrors of the Fortran API.  It is actually the user's job to declare the SHMEM API functions at the application level (seriously, check out the shmem_my_pe Fortran API definition in the spec; it's a subtle kind of awful).  SOS also supports an alternative Fortran header that is less horrible (--enable-long-fortran-header), but the default build does what the spec says.

agrippa:
Ah, gotcha. That explains why I couldn't reproduce this, I forgot I was configuring with --enable-long-fortran-header.

agrippa:
I actually just wrote my first FORTRAN OpenSHMEM program last week. It made me very, very sad.

davidozog:
Seems like this example could check for FORTRAN_LONG_HEADER?

jdinan:
This is unfortunately a pain, because the preprocessor is not part of the Fortran spec.  The ```include shmem.fh``` is already nonstandard.  I have no idea if you can reasonably expect things like ifdefs to work, given than the include worked.

What is obnoxious is that Cray's header includes the function declarations, but SGI's doesn't.  IIUC, OpenSHMEM kinda sorta adopted the SGI semantic, which you have to sleuth from the API declarations and horribly busted Fortran example programs in the spec.

davidozog:
Yeah.  Then maybe we should just move it into the "short header only tests", as in #395.

jdinan:
FYI, I added a Fortran section to the SOS troubleshooting wiki page.  :)

agrippa:
Okay, this should be ready for review now. I had to fix an oversight in my bitwise atomics test.

agrippa:
Is this ready to merge?

jdinan:
I'm swamped with SC reviews at the moment.  I should free up to review again shortly.

test/unit/atomic_bitwise.c
```diff
@@ -0,0 +1,233 @@
...
+        fprintf(stderr, "ERR - Requires > 1 PEs\n");
+        shmem_finalize();
+        return 0;
+    }
+    if (num_ranks % 2 != 0) {
+        fprintf(stderr, "ERR - Requires even number of PEs\n");
+        shmem_finalize();
+        return 0;
+    }
+
```
jdinan: Could you add another error check: ```if (sizeof(unsigned int) < 4) { ... }```?

agrippa: is my latest change good? Do you want me to squash?

jdinan: Looks good.  Once this clears testing we should be good to merge.  Thanks!!

jdinan: Oh, no need to squash this.  Clean history is a sign of a wasted life.

Issue #328, opened 2017-02-15
Remove --with-ofi build flag


davidozog:
The `--with-ofi` and `--with-ofi-libdir` flags are now deprecated, but they're still supported so as to not annoy those who depend on them.  When the time is right, these flags should be removed for the favored (and upstream) `--with-libfabric` and `--with-libfabric-libdir` from OPAL.

jdinan:
Since we removed the subconfigure for shmem_pmi, autoconf will generate a warning after we remove --with-ofi.  I suggest we remove it in 1.3.3.

jdinan:
@davidozog Can we close this issue?

davidozog:
Yes.  It was resolved with #390.  Now we should use --with-ofi and ***not*** --with-libfabric, though both do the same thing.

PR #390, opened 2017-04-24
Update to upstream OPAL_CHECK_OFI script

davidozog:
OMPI recently switched from --with-libfabric to --with-ofi.  This commit pulls in those changes (making --with-ofi the preferred flag and deprecating --with-libfabric).  I also added some logic that separates ofi and portals4 if both are detected (as well as XPMEM and CMA).


configure.ac
```diff
@@ -241,38 +241,44 @@ AC_SEARCH_LIBS([clock_gettime], [rt],
 dnl check for libraries
 
 OMPI_CHECK_PORTALS4([portals4],
-    [AC_DEFINE([USE_PORTALS4], [1], [Define if Portals 4 transport active])
-     transport="portals4"
+    [transport="portals4"
```
jdinan: Setting ```transport``` here and in the ofi check below seems redundant.  Can we remove it?

configure.ac
```diff
@@ -241,38 +241,44 @@ AC_SEARCH_LIBS([clock_gettime], [rt],
...
+# assure that only ofi is enabled if portals4 is also detected
+if test -n "$with_ofi" ; then
+    transport="ofi"
+    transport_portals4="no"
+    AC_DEFINE([USE_OFI], [1], [Define if OFI transport active])
+elif test -n "$with_portals4" ; then
+    transport="portals4"
+    transport_ofi="no"
+    AC_DEFINE([USE_PORTALS4], [1], [Define if Portals 4 transport active])
 fi
```
jdinan: What happens in the case where the user neither requested OFI nor Portals (or both)?  Is it worth trying to prevent this from happening or do we attribute this to operator error?

configure.ac
```diff
@@ -241,38 +241,44 @@ AC_SEARCH_LIBS([clock_gettime], [rt],
...
 CHECK_CMA(
-   [AC_DEFINE([USE_CMA], [1], [Define if Cross Memory Attach transport is active])
-    transport_cma="yes"],
-   [transport_cma="no"])
+    [transport_cma="yes"],
+    [transport_cma="no"])
+
+if test -n "$with_xpmem" ; then
+    transport_cma="no"
+    AC_DEFINE([USE_XPMEM], [1], [Define if XPMEM transport is active])
```
jdinan: Do we need to check ```transport_xpmem = "yes"``` before enabling ```USE_XPMEM``` to see whether the check actually succeeded?  Or are we not able to get this far if the check failed?

configure.ac
davidozog: Yes, definitely.

configure.ac
davidozog: Bad things... so I've added a warning.  If (in addition) neither XPMEM nor CMA are defined, the user will get two warnings about transport.  We could change these to errors if need be, though my gut says warnings may suffice.

configure.ac
davidozog: On second thought, I think $with_X is a poor choice here, because I think it's *possible* for $with_X to equal "no" in most, if not all of these scripts.  Also, we should try to avoid introducing variables that are defined in m4 scripts outside of configure.ac.  Does my latest commit fix this?

configure.ac
```diff
@@ -241,24 +241,24 @@ AC_SEARCH_LIBS([clock_gettime], [rt],
...
 
 OPAL_CHECK_OFI([ofi],
-    [transport="ofi"
-     transport_ofi="yes"],
+    [transport_ofi="yes"],
     [transport_ofi="no"])
 
 # assure that only ofi is enabled if portals4 is also detected
-if test -n "$with_ofi" ; then
+if test "$transport_ofi" = "yes" ; then
```
jdinan: I think we actually want ```$with_ofi``` here to check if the user asked for it.  With the new logic, an auto detected OFI library will prevent the user from creating a Portals build.

jdinan:
Closes #328

configure.ac
davidozog: Ah, right.  How about ```if test -n $with_ofi -a $with_ofi -ne "no"```?

davidozog: Yeah, we do need ```$with_X != "no"``` in these conditionals.  For example, if the user passes ```--without-ofi --with-portals4``` then ```$with_ofi``` is set to "no" and ```$with_portals4``` to "yes".  Without the check, this would choose OFI.

configure.ac
```diff
@@ -241,38 +241,44 @@ AC_SEARCH_LIBS([clock_gettime], [rt],
...
-AM_CONDITIONAL([USE_OFI], [test "$transport_ofi" = "yes"])
 
-if test "$transport_ofi" = "no" ; then
-  OMPI_CHECK_OFI([ofi],
-      [AC_DEFINE([USE_OFI], [1], [Define if OFI transport active])
-       transport="ofi"
-       transport_ofi="yes"],
-      [transport_ofi="no"])
-  AM_CONDITIONAL([USE_OFI], [test "$transport_ofi" = "yes"])
+# assure that only ofi is enabled if portals4 is also detected
```
jdinan: Would it be helpful to update this comment to something like "Check if user requested OFI or Portals 4"?

configure.ac
```diff
@@ -241,38 +241,44 @@ AC_SEARCH_LIBS([clock_gettime], [rt],
...
+if test -n "$with_ofi" -a "$with_ofi" != "no" ; then
+    transport="ofi"
+    transport_portals4="no"
+    AC_DEFINE([USE_OFI], [1], [Define if OFI transport active])
+elif test -n "$with_portals4" -a "$with_portals4" != "no" ; then
+    transport="portals4"
+    transport_ofi="no"
+    AC_DEFINE([USE_PORTALS4], [1], [Define if Portals 4 transport active])
+else
+    AC_MSG_WARN([Neither OFI nor portals4 transport found])
```
jdinan: I think it would be more accurate for the message to say "requested" rather than "found".  It's possible to reach this stage with both OFI and Portals having been detected, but neither requested by the user.  Should we add an error case here for: ```test "$transport_ofi" = "yes" -a "$transport_portals4" = "yes"```?

configure.ac
```diff
@@ -241,38 +241,44 @@ AC_SEARCH_LIBS([clock_gettime], [rt],
...
+    [transport_cma="yes"],
+    [transport_cma="no"])
+
+if test -n "$with_xpmem" -a "$with_xpmem" != "no" ; then
+    transport_cma="no"
+    AC_DEFINE([USE_XPMEM], [1], [Define if XPMEM transport is active])
+elif test -n "$with_cma" -a "$with_cma" != "no" ; then
+    transport_xpmem="no"
+    AC_DEFINE([USE_CMA], [1], [Define if Cross Memory Attach transport is active])
+fi
```
jdinan: Same as above.  Do we want an error test for both having been auto detected, but neither having been requested?

configure.ac
jdinan: This error message should ask the user to pass "--with-..." to pick one.

configure.ac
```diff
@@ -248,17 +248,24 @@ OPAL_CHECK_OFI([ofi],
...
     transport="portals4"
     transport_ofi="no"
     AC_DEFINE([USE_PORTALS4], [1], [Define if Portals 4 transport active])
+# If neither, set transport to "none" and print a warning:
 else
-    AC_MSG_WARN([Neither OFI nor portals4 transport found])
+    transport="none"
+    transport_ofi="no"
+    transport_portals4="no"
+    AC_MSG_WARN([Neither OFI nor portals4 transport requested])
```
jdinan: I think we already have a warning for this near the bottom of configure.ac.  Is it still correct?  Do we want another warning?

configure.ac
```diff
@@ -271,12 +278,21 @@ CHECK_CMA(
...
     transport_cma="no"
     AC_DEFINE([USE_XPMEM], [1], [Define if XPMEM transport is active])
 elif test -n "$with_cma" -a "$with_cma" != "no" ; then
     transport_xpmem="no"
     AC_DEFINE([USE_CMA], [1], [Define if Cross Memory Attach transport is active])
+# If neither, disable XPMEM and CMA:
+else
+    transport_xpmem="no"
+    transport_cma="no"
+    AC_MSG_RESULT([Neither XPMEM nor CMA transport requested])
```
jdinan: This is also printed as the final step in configure.ac.  Probably not necessary to print it here.

configure.ac
```diff
@@ -248,17 +248,24 @@ OPAL_CHECK_OFI([ofi],
     [transport_ofi="yes"],
     [transport_ofi="no"])
 
-# assure that only ofi is enabled if portals4 is also detected
-if test -n "$with_ofi" -a "$with_ofi" != "no" ; then
+# If both OFI and portals4 requested, user needs to choose one:
+if test -n "$with_ofi" -a "$with_ofi" != "no" -a -n "$with_portals4" -a "$with_portals4" != "no" ; then
+    AC_MSG_ERROR([Cannot choose both OFI and portals4 transports, use --with to choose one])
```
jdinan: Should we say "use --help for details"?  In case we ever change the arguments.

configure.ac
davidozog: Yeah, there are two, but I kinda think it's helpful for determining/debugging what happens with the OFI/Portals checks, as well as printing at that point of the configuration output.  The final warning also includes XPMEM and CMA too, so it wouldn't get printed if XPMEM (or CMA) _only_ is enabled.

configure.ac
davidozog: Same here.  Probably not necessary, but helpful imho to separate from OFI/portals.

configure.ac
davidozog: Sure.

configure.ac
davidozog: I believe the final check prints only when no transports at all are found.  For example, if XPMEM is enabled but neither OFI nor Portals, that warning will not print.  This separate warning has been helpful for me, but I can remove it if you want.

configure.ac
davidozog: Similar to the above.

jdinan: Ahh, ok

configure.ac
```diff
@@ -241,38 +241,60 @@ AC_SEARCH_LIBS([clock_gettime], [rt],
...
 CHECK_CMA(
-   [AC_DEFINE([USE_CMA], [1], [Define if Cross Memory Attach transport is active])
-    transport_cma="yes"],
-   [transport_cma="no"])
+    [transport_cma="yes"],
+    [transport_cma="no"])
+
+# If both XPMEM and CMA requested, user needs to choose one:
+if test -n "$with_xpmem" -a "$with_xpmem" != "no" -a -n "$with_cma" -a "$with_cma" != "no" ; then
+    AC_MSG_ERROR([Cannot choose both XPMEM and CMA transports, use --with to choose one])
```
jdinan: Update this one to --help, too?

davidozog: Yup!

jdinan:
Sweet!  Once it clears testing, feel free to merge.  :shipit:

davidozog:
Note: this addresses issue #103.

PR #391, opened 2017-04-24
Add triggered tree barrier algorithm over portals4

davidozog:
Currently for use with portals4 only.  Enable with:
SMA_BARRIER_ALGORITHM=trigger

### DIFF
--- a/src/collectives.c
+++ b/src/collectives.c
@@ -34,6 +34,7 @@ char *coll_type_str[] = { "AUTO",
                           "LINEAR",
                           "TREE",
                           "DISSEM",
+                          "TRIGGER",
                           "RING",
                           "RECDBL" };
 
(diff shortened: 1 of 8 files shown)


jdinan:
One quick comment - I think we'd like to split the barrier into setup and execution phases.  We want the triggered ops to be already posted when the barrier occurs to reduce latency.  I think we would want to post them initially in init() and then re-post when exiting the barrier_all routine.  This means that we also need to cancel them during finalization.  The local PE increments the CT to register its arrival at the barrier, so the threshold is also one greater.

davidozog:
Yeah, this makes sense.  It may not be a huge win because it just hides the first and last barriers' initialization latency in init() and finalize(), but it seems worth doing.  Is that the goal, or do you foresee this strategy reducing latency for all barriers?

jdinan:
Needs update to allow triggered ops to be pre-posted.

PR #392, opened 2017-05-01
Fix asymmetric allocation bug in GUPS

### LINKED ISSUE #384: GUPS fails with certain npes and memory sizes
GUPS seg. faults when launching certain numbers of PEs (for example, 5, 14, ..., 166, and more) on the 3 providers I've tried so far (libfabric/psm2, libfabric/sockets, portals4/IB).

The smallest job that consistently produces the error is this:

```
oshrun -n 5 ./gups -m 205
```

With 5 PEs, I also notice that all buffer sizes below 204 bytes work fine, and all above 204 seg. fault.

The error over portals4 on IB is:
```
[001] ERROR: ../../src/transport_portals4.h:855 return code 5
[mpiexec@comet-14-04.sdsc.edu] handle_pmi_cmd (pm/pmiserv/pmiserv_cb.c:52): Unrecognized PMI command: abort | cleaning up processes
[mpiexec@comet-14-04.sdsc.edu] control_cb (pm/pmiserv/pmiserv_cb.c:280): unable to process PMI command
[mpiexec@comet-14-04.sdsc.edu] HYDT_dmxu_poll_wait_for_event (tools/demux/demux_poll.c:76): callback returned error status
[mpiexec@comet-14-04.sdsc.edu] HYD_pmci_wait_for_completion (pm/pmiserv/pmiserv_pmci.c:198): error waiting for event
[mpiexec@comet-14-04.sdsc.edu] main (ui/mpich/mpiexec.c:344): process manager error waiting for completion

```

Over libfabric the error is only:
```
=   PID 92899 RUNNING AT node109
=   EXIT CODE: 11
```

jdinan:
OpenSHMEM requires that any asymmetric allocation must be the last call to ```shmem_malloc```.  Since the Table allocation in GUPS can be asymmetric (as reported in #384), ensure it's the last symmetric allocation call.

Closes #384

### DIFF
--- a/test/apps/gups.c
+++ b/test/apps/gups.c
@@ -261,10 +261,10 @@ UpdateTable(uint64_t *Table,
             uint64_t MinLocalTableSize,
             uint64_t Top,
             int Remainder,
-            int64_t niterate,
+            uint64_t niterate,
             int use_lock)
 {
-  int64_t iterate;
+  uint64_t iterate;
   int index;
   uint64_t ran, remote_val, global_offset;
   int remote_pe;
@@ -307,7 +307,7 @@ SHMEMRandomAccess(void)
   int NumProcs, MyProc;
   int Remainder;            /* Number of processors with (LocalTableSize + 1) entries */
   uint64_t Top;               /* Number of table entries in top of Table */
-  int64_t LocalTableSize;    /* Local table width */
+  uint64_t LocalTableSize;    /* Local table width */
   uint64_t MinLocalTableSize; /* Integer ratio TableSize/NumProcs */
   uint64_t logTableSize, TableSize;
 
--- a/test/apps/gups.c
+++ b/test/apps/gups.c
@@ -319,7 +319,7 @@ SHMEMRandomAccess(void)
   uint64_t NumUpdates_Default; /* Number of updates to table (suggested: 4x number of table entries) */
   uint64_t NumUpdates;  /* actual number of updates to table - may be smaller than
                        * NumUpdates_Default due to execution time bounds */
-  int64_t ProcNumUpdates; /* number of updates per processor */
+  uint64_t ProcNumUpdates; /* number of updates per processor */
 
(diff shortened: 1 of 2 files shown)


jdinan:
Ready for review.

test/apps/gups.c
```diff
@@ -389,10 +389,12 @@ SHMEMRandomAccess(void)
 
 
   sAbort = 0;
-  HPCC_Table = shmem_malloc(LocalTableSize * sizeof(uint64_t));
+  /* Ensure the allocation size is symmetric */
+  HPCC_Table = shmem_malloc((Remainder > 0 ? (MinLocalTableSize + 1) : LocalTableSize)
```
davidozog: This fixes the bug.  Nice find!

test/apps/gups.c
```diff
@@ -264,7 +264,7 @@ UpdateTable(uint64_t *Table,
             int64_t niterate,
             int use_lock)
 {
-  uint64_t iterate;
+  int64_t iterate;
```
jdinan: Hmm...I think you're right.  I made this change to quiet a warning on line 278 because the conditional was comparing signed with unsigned, but the indentation suggests that LocalTableSize used to be a uint64_t and was changed to signed.  Changing it back seems like a much better fix for the sign issue.

davidozog: Indentation evidence, huh?  There is no such thing as a perfect crime.

test/apps/gups.c
```diff
@@ -322,10 +322,10 @@ SHMEMRandomAccess(void)
   int64_t ProcNumUpdates; /* number of updates per processor */
```
davidozog: Might as well unsign this one too, right?  It's just 4*LocalTableSize and it's what's passed to UpdateTable.

jdinan: Oh, of course.  :man_facepalming:

Actually, I'm surprised the compiler didn't warn about loss of information passing int64_t into a uint64_t function argument..

PR #394, opened 2017-05-04
Improve error message when symmetric heap setup fails

richardnorth3:
Signed-off-by: Richard North III <richard.north.iii@intel.com>

### DIFF
--- a/src/symmetric_heap_c.c
+++ b/src/symmetric_heap_c.c
@@ -204,7 +204,7 @@ static void *mmap_alloc(size_t bytes)
                fd,
                0);
     if (ret == MAP_FAILED) {
-        RAISE_WARN_STR("mmap for symmetric heap failed");
+        RAISE_WARN_MSG("symmetric heap setup failed (%zuB), reduce SHMEM_SYMMETRIC_SIZE\n", bytes);
         ret = NULL;
     }
     if (fd) {


richardnorth3:
Isayah,

This error will only occur if the user allocates too much memory on the
symmetric. Hence, they will have to decrease the amount they have requested
to allocate on the symmetric because it is not enough memory available on
their current system.

Regards,

Richard

jdinan:
The message contains the symmetric heap size that SOS attempted to set up.  It will likely be larger than what the user requested with SMA_SYMMETRIC_SIZE.

jdinan:
Isayah, could you review the comments here and let us know if this is ready to merge?

ireed:
Ok. I was confused by the byte output - if it is to show how much a user requested, and thus user should decrease?
Im assuming you've checked that the message doesn't occur in the case where not enough heap is requested, since (I think) the original code shows the same output for both (hence the motivation for this change?).

jdinan:
There is a different message that gets printed by ```shmem_malloc``` if the requested allocation size is greater than the amount of space available in the symmetric heap.  In that case, we suggest increasing ```SHMEM_SYMMETRIC_SIZE```.

davidozog:
An unsuccessful call to mmap sets errno, so maybe we could look at that at some point in the future.  Seems like a number of different things could go wrong with this call to mmap, not only too big a symmetric heap request (which possibly sets errno to EAGAIN?).  Just a thought.

jdinan:
File an issue for it.  :)

PR #396, opened 2017-05-05
uni-dir perf suite addition: target put/get with variable PE sets

kseager:
This test suite adds the following input variables: -t, -l, -r

-t:  enables timing of the target side of a transfer (will output initiator timing as well, side-by-side)
-l: PEs stream across two nodes for this suite, the "left side" is the initiator side, you can reduce the PEs streaming on this side, X <= PPN 
-r: same ideas as above: "right side" is the target side of the transfer stream, you can reduce the PEs streaming on this side, X <= PPN 

feature notes: 
-only for sizes 8bytes and 4KB
-you cannot use -l/r together, if you want less on both sides, you should launch less
-window size is set at 1 and cannot be altered for -t 

Review notes:
target_put.h is best read on its own at head commit (was overhauled since initial commit)

@jdinan please review

### DIFF
--- a/test/performance/shmem_perf_suite/bw_common.h
+++ b/test/performance/shmem_perf_suite/bw_common.h
@@ -85,7 +85,7 @@ typedef struct perf_metrics {
     unsigned long int window_size, warmup;
     int validate;
     int target_data;
-    int my_node, num_pes;
+    int my_node, num_pes, sztarget, midpt;
     bw_units unit;
     char *src, *dest;
     const char *bw_type;
@@ -109,6 +109,8 @@ void static data_init(perf_metrics_t * data) {
     data->target_data = false;
     data->my_node = shmem_my_pe();
     data->num_pes = shmem_n_pes();
+    data->midpt = data->num_pes/2;
+    data->sztarget = data->midpt;
     data->src = NULL;
     data->dest = NULL;
     data->cstyle = COMM_PAIRWISE;
@@ -134,10 +136,11 @@ int static inline partner_node(perf_metrics_t my_info)
         return 0;
 
     if(my_info.cstyle == COMM_PAIRWISE) {
-        int pairs = my_info.num_pes / 2;
+        int pairs = my_info.midpt;
 
-        return (my_info.my_node < pairs ? my_info.my_node + pairs :
-                    my_info.my_node - pairs);
+        return (my_info.my_node < pairs ?
+            ((my_info.my_node + pairs) % (my_info.sztarget + pairs)) :
+            my_info.my_node - pairs);
     } else {
         assert(my_info.cstyle == COMM_INCAST);
         return INCAST_PE;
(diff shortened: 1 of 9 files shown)


test/performance/shmem_perf_suite/bw_common.h
```diff
@@ -100,8 +106,12 @@ void static data_init(perf_metrics_t * data) {
     data->warmup = WARMUP; /*number of initial iterations to skip*/
     data->unit = MB;
     data->validate = false;
+    data->target_data = false;
     data->my_node = shmem_my_pe();
     data->num_pes = shmem_n_pes();
+    data->midpt = data->num_pes/2;
```
jdinan: Is there any potential issue when num_pes is odd?

test/performance/shmem_perf_suite/bw_common.h
```diff
@@ -127,10 +137,10 @@ int static inline partner_node(perf_metrics_t my_info)
         return 0;
 
     if(my_info.cstyle == COMM_PAIRWISE) {
-        int pairs = my_info.num_pes / 2;
+        int pairs = my_info.midpt;
 
-        return (my_info.my_node < pairs ? my_info.my_node + pairs :
-                    my_info.my_node - pairs);
+        return (my_info.my_node < pairs ? (my_info.my_node + pairs) :
+            (my_info.my_node - pairs));
```
jdinan: If ```num_pes``` is odd, the ```num_pes-1``` PE is an "extra PE" and do the wrong thing here.  Or is this problem already elimiated earlier on in the benchmark?

test/performance/shmem_perf_suite/bw_common.h
```diff
@@ -198,25 +222,62 @@ void static command_line_arg_check(int argc, char *argv[],
...
+            ((metric_info->trials + metric_info->warmup) * TARGET_SZ_MIN)) ||
+            (metric_info->max_len <
+            ((metric_info->trials + metric_info->warmup) * TARGET_SZ_MAX))) {
+                error = true;
+            }
+    }
+
     if (error) {
         if (metric_info->my_node == 0) {
             fprintf(stderr, "Usage: \n[-s start_length] [-e end_length] "\
```
jdinan: The backslashes at the end of these lines are not necessary.  C will paste adjacent string literals together, even when they are separated by line breaks.

test/performance/shmem_perf_suite/bw_common.h
```diff
@@ -271,10 +332,10 @@ void static print_atomic_results_header(perf_metrics_t metric_info) {
...
-            "max message size %lu with multiple of %lu increments\n",
-            metric_info.num_pes, metric_info.trials, metric_info.window_size,
-            metric_info.max_len, metric_info.size_inc);
-
+            "max message size %lu with multiple of %lu increments, "\
+            "\ntargeting %d remote PEs initiated from %d PEs\n", metric_info.num_pes,
+            metric_info.trials, metric_info.window_size, metric_info.max_len,
+            metric_info.size_inc, metric_info.sztarget, metric_info.szinitiator);
     printf("\nLength           %s           "\
             "Message Rate\n", metric_info.bw_type);
```
jdinan: Same here, trailing backslashes can be removed.  What you have is perfectly valid C, so this is not critical, but it may make the code a little easier to maintain.

jdinan:
This looks reasonable, although the code is not easy to review.  Have you run some experiments to determine that it's working as expected?  You could do something like N->1 (initiators->targets) and verify that the bandwidth saturates at the BW of a single port, N->2 two ports, and also the reverse.

test/performance/shmem_perf_suite/bw_common.h
kseager: the base model (aka everything but the target usage) assumes you have a symmetric or even number of PEs.  Even with the target usage, you launch with an even set of PE's and then reduce the number you want on the left or right side -l/-r via the command line arguments

test/performance/shmem_perf_suite/bw_common.h
kseager: I catch launching an odd number of PEs as invalid usage and will throw an error and exit

PR #398, opened 2017-05-19
Add quiet/wait polling limits

davidozog:
Calling fi_cntr_wait incurs significant overhead on some systems. Instead, mix polling with waiting by setting a limit on the number of iterations in the polling loop.  This is done with some environment variables.

There are 3 new environment variables: SMA_OFI_GLOBAL_POLL_LIMIT, SMA_OFI_PUT_POLL_LIMIT, and SMA_OFI_GET_POLL_LIMIT.  The GLOBAL limit sets the max number of polls in both put (quiet) and get (wait), and if set, overrides the other 2 environment variables.  The PUT and GET values set these poll limits individually.

Also note that --enable-completion-polling sets the default poll limit to -1, which leads to an infinite poll loop.  The default limit is 0, which simply leads to a call to fi_cntr_wait.


src/transport_ofi.h
```diff
@@ -286,30 +288,30 @@ void shmem_transport_put_quiet(void)
...
             SPINLOCK_BODY();
+        } else if (fail) {
+            struct fi_cq_err_entry e = {0};
+            fi_cq_readerr(shmem_transport_ofi_put_nb_cqfd, (void *)&e, 0);
+            OFI_CQ_ERROR(shmem_transport_ofi_put_nb_cqfd, &e);
         }
-        else if (fail) {
+        poll_count++;
+    }
+    if (success < shmem_transport_ofi_pending_put_counter) {
```
jdinan: Out of scope for this PR, but this particular sequence looks like it will break in a thread-multiple job.

jdinan: That is, if a put is issued by another thread after the above polling loop and before this check, it will cause the thread that was trying to exit to instead call ```fi_cntr_wait```.  This will be an error if the counter has ```FI_WAIT_NONE``` set.

jdinan: On second thought, I think it's in-scope for this PR, since this is introducing the race.

src/transport_ofi.c
```diff
@@ -1197,6 +1223,17 @@ void shmem_transport_print_info(void)
            (NULL != shmem_util_getenv_str("OFI_ATOMIC_CHECKS_WARN")) ?
            "Set" : "Not set");
     printf("\tDisplay warnings about unsupported atomic operations\n");
+
+    if (NULL != shmem_util_getenv_str("OFI_GLOBAL_POLL_LIMIT")) {
+        printf("SMA_OFI_GLOBAL_POLL_LIMIT      %ld\n", shmem_util_getenv_long("OFI_GLOBAL_POLL_LIMIT", 0, DEFAULT_POLL_LIMIT));
+        printf("\tMax # of polls for put/get completions is %"PRIu64"\n", shmem_transport_ofi_put_poll_limit);
```
jdinan: I think we want to handle this one differently, because if it's set it will override the two below.  However, if it's not set, we will still print the default poll limit.

src/transport_ofi.c
```diff
@@ -1197,6 +1223,17 @@ void shmem_transport_print_info(void)
...
+    } else {
+        printf("SMA_OFI_PUT_POLL_LIMIT  %ld\n", shmem_util_getenv_long("OFI_PUT_POLL_LIMIT", 0, DEFAULT_POLL_LIMIT));
+        printf("\tMax # of polls for put (quiet) completions is %"PRIu64"\n", shmem_transport_ofi_put_poll_limit);
+        printf("SMA_OFI_GET_POLL_LIMIT  %ld\n", shmem_util_getenv_long("OFI_GET_POLL_LIMIT", 0, DEFAULT_POLL_LIMIT));
+        printf("\tMax # of polls for get (wait) completions is %"PRIu64"\n", shmem_transport_ofi_get_poll_limit);
+    }
```
jdinan: Can we print ```shmem_transport_ofi_put/get_poll_limit``` directly in place of the calls to getenv?

src/transport_ofi.c
davidozog: If it's not set, doesn't getenv_str(...) return NULL?

src/transport_ofi.c
davidozog: Yup

src/transport_ofi.h
davidozog: Yeah, I see the race.  Any thoughts on how best to fix it?

jdinan:
Some proposed updates: https://github.com/davidozog/sandia-shmem/pull/1

Also, I don't see ```SMA_OFI_GLOBAL_POLL_LIMIT``` in the PR?

davidozog:
Cool - the updates are good.  Nice catch on the bugfix.

I believe we decided to remove the global limit in commit 6e13ee8.  Should I edit my original comment?

jdinan:
Ahh, ok, forgot about that change.  No need to update the message.

PR #401, opened 2017-05-26
Remove less-than-zero arg checks for unsigned args

jdinan:
Don't check args for less-than-zero errors when the type of the args is unsigned (e.g. ```size_t```).

Fixes coverity issues 169688-169781.


jdinan:
@ireed Ready for you to review.

jdinan:
I can has review pls?  @ireed kthxbai?

PR #405, opened 2017-06-08
Updates for v1.3.3rc1 release

jdinan:
* Update NEWS
* Bump version to v1.3.3rc1
* Last minute fixes to memory management API

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -2,6 +2,18 @@
 
 Sandia OpenSHMEM NEWS -- history of user-visible changes.
 
+v1.3.3rc1
+---------
+- Fixed dissemination barrier race that could lead to incorrect synchronization.
+- Improved collect and all-to-all collectives performance.
+- Improved error reporting and debugging output.
+- Added support for proposed bitwise atomics (see shmemx.h).
+- Added shmemx_pcontrol function to profiling interface extension (see shmemx.h).
+- Added symbol hiding support in SHMEM library.
+- Fixed symmetric heap corruption bug in GUPS benchmark (see test/apps/gups.c).
+- Added target-side measurement and asymmetric configuration support to
+  performance suite benchmarks.
+
 v1.3.2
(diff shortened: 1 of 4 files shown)


jdinan:
@ireed Could you check the realloc/align/free fixes against the Mellanox tests?

davidozog:
`shmemalign` unit tests are failing though...

jdinan:
Should be fixed with f1fb86c.

jdinan:
Cray tests require implementation to round up to nearest power of two.  Latest patch should fix this.

ireed:
If shmem_align is a fancy wrapper for memalign then that may cause issues with the mellanox tests. memalign(0,100) and memalign(3,100) return valid pointers on my system, thus failing the tests. But, posix_memalign() does work.

jdinan:
We will return NULL in both those cases since the alignment is less than size(void*) (which is 8).

ireed:
(update) The merged code is not passing the malloc tests, BUT it works if i manually write the test code in a new program (it fails if i do this without the merge).

ireed:
Passes the original failing malloc tests, but fails a different test now. 
`    /* test that we get desired alignments */
    for (align = 1; align <= 4096; align *= 2)
    {
       p = shmem_align(align, 37);
       if (!p)   log_debug(OSH_TC, "failed to alloc with align = %d\n", align);

       if ((unsigned long)p & (align-1))     log_debug(OSH_TC, "failed to alloc with align = %d\n", align);
       shmem_free(p);
    }
`

(fails when align is 1, 2, and 4; passes in subsequent iterations)
This test passed before the merge. 
My guess is that the new cap on 8byte alignment affected this test.

jdinan:
Ok, we can back out that change.  Since the spec doesn't place any restrictions on alignment, I suppose that we shouldn't either.

PR #407, opened 2017-06-09
Include transport_none in sources

jdinan:
Ensure that transport_none.h is included in the distribution tarball.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -88,6 +88,13 @@ libsma_la_SOURCES += \
 	transport_ofi.c
 endif
 
+if !USE_PORTALS4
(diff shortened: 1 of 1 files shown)


src/Makefile.am
```diff
@@ -88,6 +88,13 @@ libsma_la_SOURCES += \
 	transport_ofi.c
 endif
 
+if !USE_PORTALS4
+if !USE_OFI
+libsma_la_SOURCES += \
+	transport_none.h
+endif
+endif
+
```
jdinan: Only came up during RPM packaging (make dist wasn't including this file).

Issue #414, opened 2017-07-14
pthread link error on Ubuntu


davidozog:
On Ubuntu 16.04 and 16.10, I'm getting the following error in "make check" when configured with ```--enable-threads```:

```
  CC       mt_a2a-mt_a2a.o
  CCLD     mt_a2a
/usr/bin/ld: mt_a2a-mt_a2a.o: undefined reference to symbol 'pthread_create@@GLIBC_2.2.5'
//lib/x86_64-linux-gnu/libpthread.so.0: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
Makefile:1427: recipe for target 'mt_a2a' failed
make[3]: *** [mt_a2a] Error 1
...
```

jdinan:
Fixed by #415

PR #415, opened 2017-07-14
Pthread cflags fix for issue #414

davidozog:
Adding PTHREAD_CFLAGS to mt_a2a LDADD to address issue #414.

This is a pretty minimal change to fix the problem.  We might want to consider setting PTHREAD_LDFLAGS in the [OPAL upstream m4](https://github.com/Sandia-OpenSHMEM/SOS/blob/master/config/opal_config_pthreads.m4#L608).  To my eye, it will never be set to anything.


configure.ac
```diff
@@ -571,6 +571,7 @@ WRAPPER_COMPILER_EXTRA_LDFLAGS=`echo "$WRAPPER_COMPILER_EXTRA_LDFLAGS" | sed -e
 AC_MSG_RESULT([$WRAPPER_COMPILER_EXTRA_LDFLAGS])
 AC_SUBST(WRAPPER_COMPILER_EXTRA_LDFLAGS)
 AC_SUBST(WRAPPER_COMPILER_EXTRA_LIBS)
+AC_SUBST(PTHREAD_CFLAGS)
```
jdinan: Is this necessary?

jdinan: Bueller?

davidozog:
I'm not sure why my comments aren't being posted... is this one pending?

configure.ac
davidozog: Perhaps the confusion is due to the fact that the PTHREAD_* variables are not "preset output variables" like CFLAGS, LDFLAGS, and friends, so they are not defined in any Makefile.in by default. With AC_SUBST(PTHREAD_*), the @PTHREAD_*@ var is placed in Makefile.in and set at configure time.  On Ubuntu 16.04, opal_config_pthreads.m4 _only_ sets PTHREAD_CFLAGS for some reason, and the mt_a2a unit test needs -pthread at _link_ time to build.  So, another option is to include PTHREAD_CFLAGS in LDFLAGS in configure.ac.  Yet another option is to figure out why PTHREAD_LIBS and/or PTHREAD_LDFLAGS aren't set correctly.  I hope that makes sense...

jdinan: Right, what's surprising is that ```AC_SUBST(PTHEAD_CFLAGS)``` does anything at all.  The current usage of ```$PTHREAD_CFLAGS/LIBS``` in ```Makefile.am``` is definitely wrong.  Past Jim had not wanted to include these in ```$CFLAGS``` and ```$LDFLAGS```, but present Jim does not remember why...  I do think I prefer the ```$CFLAGS/$LDFLAGS``` approach to an autoconf substitution (e.g. ```@PTHREAD_CFLAGS```).

jdinan: It may have been because I was concerned that ```$CFLAGS/$LDFLAGS``` would propagate to the wrappers.

davidozog:
bot:lanl:retest

jdinan:
When I was reading up on ```AC_SUBST```, I missed the fact that in addition to doing the @var@ substitution, autoconf also defines a variable with the given name in each Makefile that it generates.  Thus, I think the correct fix is actually to add:
```
AC_SUBST(PTHREAD_CFLAGS)
AC_SUBST(PTHREAD_LIBS)
```

jdinan:
Woo hoo!

PR #417, opened 2017-07-26
Setting symmetric heap size env var for ISx tests to account for rece…

ireed:
…nt ISx changes that use more dynamic memory.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -255,6 +255,7 @@ script:
     ###
     - export PATH=$TRAVIS_INSTALL/sandia-shmem-ofi/bin:$TRAVIS_INSTALL/hydra/bin:$BASE_PATH
     - export OSHRUN_LAUNCHER="mpiexec.hydra"
+    - export SMA_SYMMETRIC_SIZE='4G'
     - cd $TRAVIS_SRC/ISx/SHMEM
     - make CC=oshcc LDLIBS=-lm
     - oshrun -np 4 ./bin/isx.strong 134217728 output_strong


lanl-ompi:

davidozog:
Does it really need to be 4G?  That seems pretty large to me, but sensible.  Do you know what recent change causes this requirement?  Thanks!

davidozog:
Ah, I think I see the change here:
https://github.com/ParRes/ISx/commit/338cd28c3e261bc840dbfc83209620c675fb51a4

I actually calculate a little over 2GB if KEY_TYPE is 8 bytes, so maybe 4G is ok.

Re-running the tests now to see if that fixes our issue in PR #418.

ireed:
Im slowly increasing the key size to uint 128. It should be complete within 6 weeks. 
The idea behind this pull request is to show that the env var is now needed. I chose 4G so i wont need to make SOS pull request to increase heap size each time i increase ISx key size. 
While i have no problem doing an SOS pull request each time, im assuming you dont want SOS stability dependent on a non-SOS admin/person,

jdinan:
Keep in mind that anything in travis.yml is constrained by the Travis CI
environment.  Can we run 4 PEs each with 4gb symmetric heaps in one vm
instance?  Seems to work, but it might also fall over dead if you try to
use all the memory.  Travis VM specs are online, might want to check.

davidozog:
I see [here](https://docs.travis-ci.com/user/reference/precise/) that sudo-enabled Travis VM's have 7.5GB, but without sudo they have 4GB.  Seems to me like we're playing with fire a bit, but it certainly seems to work...

jdinan:
That already exceeds physical memory, so we may be paging with the current workload.  At a minimum, a comment about memory requirement would be a good breadcrumb in case this bites us in the future (e.g. if we migrate back to the non-sudo environment).  As long as the tests are still running quickly, this may be benign.  If they are not running quickly (this is CI testing, not nightly regression, so we want speed) then we should reduce the workload sizes.

davidozog:
Yeah, I agree.  I just compared some test execution times from just before 1.3.3 and after the 4GB change, and strangely enough, there is essentially no change.  In fact, I see more instances where the new tests run slightly _faster_ than slower (which is counter-intuitive, but it's noisy).  So I agree that we're probably fine for now, but it's definitely something to keep an eye on.

jdinan:
Please add a comment to the effect:
```
# Note: This may exceed the physical memory available in the CI testing environment
```

Apart from that, I'm ok with a merge.

jdinan:
Ahh, I missed the window on this merge.  I'd appreciate a follow-up PR to add the comment.  It's useful to warn future developers about this.

Issue #286, opened 2016-12-08
mt_a2a test failing


jdinan:
The mt_a2a test is failing occasionally in Travis CI with the OFI transport with the following message:

```
Starting multithreaded test on 2 PEs, 8 threads/PE
Atomic test error: [PE = 1 | TID = 7] -- dest = -1 (expected 1), flag = 2 (expected 2)
Success
```

jdinan:
Another failure message:
```
Starting multithreaded test on 2 PEs, 8 threads/PE
Atomic test error: [PE = 1 | TID = 7] -- dest = -1 (expected 1), flag = 2 (expected 2)
Atomic test error: [PE = 1 | TID = 5] -- dest = -1 (expected 1), flag = 2 (expected 2)
Atomic test error: [PE = 1 | TID = 3] -- dest = -1 (expected 1), flag = 2 (expected 2)
Atomic test error: [PE = 1 | TID = 4] -- dest = -1 (expected 1), flag = 2 (expected 2)
Atomic test error: [PE = 1 | TID = 2] -- dest = -1 (expected 1), flag = 2 (expected 2)
```

jdinan:
This bug has not manifested since #352.

jdinan:
Still failing in v1.3.2

jdinan:
Fixed (finally?) by #418

PR #418, opened 2017-07-27
Changed mt_a2a.c such that the test works.

AMcKinley15:
File was missing a pthread barrier and was causing unspecified execution.

Signed-off-by: = <=alex@mckpals.com>

### DIFF
--- a/test/unit/mt_a2a.c
+++ b/test/unit/mt_a2a.c
@@ -78,6 +78,7 @@ static void * thread_main(void *arg) {
         pthread_mutex_unlock(&mutex);
     }
 
+    pthread_barrier_wait(&fencebar);
     if (0 == tid) shmem_barrier_all();
     pthread_barrier_wait(&fencebar);
 


lanl-ompi:

jdinan:
Looks like the failure was caused by an upstream change in ISx:
```
[0003] WARN:  ../../src/symmetric_heap_c.c:137: shmem_internal_get_next
[0003]        Out of symmetric memory, heap size 537919488, overrun 1073745920
[0003]        Try increasing SHMEM_SYMMETRIC_SIZE
ISx v1.1
  Number of Keys per PE: 33554432
  Max Key Value: 32768
  Bucket Width: 8192
  Number of Iterations: 1
  Number of PEs: 4
  WEAK_ISOBUCKET Scaling!
===================================================================================
=   PID 12117 RUNNING AT testing-gce-83ffce2b-d481-4b37-b900-1ca14f3c4817
=   EXIT CODE: 139
=   CLEANING UP REMAINING PROCESSES
=   YOU CAN IGNORE THE BELOW CLEANUP MESSAGES
===================================================================================
YOUR APPLICATION TERMINATED WITH THE EXIT STRING: Segmentation fault (signal 11)
This typically refers to a problem with your application.
Please see the FAQ page for debugging suggestions
```

jdinan:
Thanks; this does look like it fixes a races with accesses to the ```dest``` array.  Might fix #286.

davidozog:
Yes, this change fixes #286 as far as we can reproduce it.  (To reproduce, it helped to have more VM cores than physical cores.)

Not sure yet what's causing the ISx problems.

PR #419, opened 2017-08-02
Update SMA_SYMMETRIC_SIZE for ISx

jdinan:
Move the SMA_SYMMETRIC_SIZE env var adjustment to an oshrun argument to avoid changing the environment variable setting (which may be set elsewhere in the future and would be used by all other tests).


jdinan:
@davidozog Ready to merge?

PR #426, opened 2017-08-18
Fix missing manpages from RPM build

davidozog:
Adds the man pages directory to EXTRA_DIST so that "make dist" includes
them in the packages tarball.  Also adds the manpage files to the
"devel" RPM build and "oshrun" to the standard RPM.

Signed-off-by: David Ozog <david.m.ozog@intel.com>

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -13,7 +13,7 @@
 
 ACLOCAL_AMFLAGS = -I config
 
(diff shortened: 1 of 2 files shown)


davidozog:
@matcabral - Do these changes look ok to you?

matcabral:
👍

PR #427, opened 2017-08-22
Fix handling of zero-length put/get operations

jdinan:
Note that atomics are always length 1, so I added asserts to the corresponding ```atomic/put_small``` routines rather than a length check (the assertions are only included in error checking builds).

### DIFF
--- a/src/shmem_comm.h
+++ b/src/shmem_comm.h
@@ -52,6 +52,8 @@ shmem_internal_put_small(void *target, const void *source, size_t len, int pe)
 {
     int node_rank;
 
+    shmem_internal_assert(len > 0);
+
     if (-1 != (node_rank = SHMEM_GET_RANK_SAME_NODE(pe))) {
 #if USE_XPMEM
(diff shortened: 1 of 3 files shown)


src/shmem_comm.h
```diff
@@ -140,6 +146,8 @@ shmem_internal_get(void *target, const void *source, size_t len, int pe)
 {
     int node_rank;
 
+    if (len == 0) return;
```
davidozog: Should this be an assert?  I imagined 'get' being on the critical path; but now that I think of it, maybe not.

jdinan: Nope, user can provide len == 0 in a get operation.

jdinan:
CT operations always need to be performed to generate the counter increment at the target, regardless of size.

src/shmem_comm.h
davidozog: Oh yeah, I could see that - thanks.

PR #431, opened 2017-09-11
Adding mellanox test suite

ireed:


lanl-ompi:

davidozog:
Looks like the barrier stress test needs fewer iterations or to be skipped entirely for Travis to have a chance to complete.  We could also do something like the [QUICK_TEST macro from the Cray tests](https://github.com/openshmem-org/tests-cray/blob/b9ed1283b99649e8cad3ace82454a4f8820fe058/sma1/shmem_long_finc_swap.c#L43).  Any thoughts?  I'd say skip it, and call the other 'sync' tests manually.

ireed:
There are others that are dangerously close to time-out errors due to long iterations - data:put_blocking is one example. They fail sometimes. I reduce the iterations by a factor of 10, but im not sure if thats something i could/should submit as a mellanox pull-request. 
They may run into the same issues after setting up their jenkins.

davidozog:
Yeah, it does seem like a few other tests will likely need a "QUICK_TEST" number for iterations.  I'll look into getting this through the configury and post a PR.  In the mean time, can you please investigate what # of iterations gets the time-consuming tests to run in just a few seconds on a 2-4 core laptop?

Also, if they fail intermittently (which I haven't seen yet locally), we'll want to figure out what's happening.

ireed:
Intermittently is more for travis. The testing time varies and can sometimes be to long for travis. The analysis suite is the poster-child for this behavior - it doesnt fails (because it doesnt check anything) but it can be very long.
Maybe '_fail intermittently_' should be rephrased to '_times out intermittently_'.

jdinan:
One comment is that the we could also build (but not run, since the sockets version of Portals 4 is busted) the Mellanox tests with the Portals build of SOS.

PR #432, opened 2017-09-13
Incorporate OpenPA for atomics.

davidozog:
OpenPA v1.0.3 from upstream mpich.  This commit includes bare minimum
configury for inclusion in OFI transport, as well as atomics for
get_wait and put_quiet counter accesses.


davidozog:
@jdinan any initial feedback?

src/transport_ofi.c
```diff
@@ -72,9 +72,9 @@ uint8_t**                       shmem_transport_ofi_target_heap_addrs;
 uint8_t**                       shmem_transport_ofi_target_data_addrs;
 #endif /* ENABLE_REMOTE_VIRTUAL_ADDRESSING */
 #endif /* ENABLE_MR_SCALABLE */
-uint64_t                        shmem_transport_ofi_pending_put_counter;
-uint64_t                        shmem_transport_ofi_pending_get_counter;
-uint64_t                        shmem_transport_ofi_pending_cq_count;
```
jdinan: OPA_int_t is 32-bit, and these counters will overflow if the size is reduced to 32-bits.  😞

src/transport_ofi.h
```diff
@@ -299,18 +300,18 @@ void shmem_transport_put_quiet(void)
...
             struct fi_cq_err_entry e = {0};
             fi_cq_readerr(shmem_transport_ofi_put_nb_cqfd, (void *)&e, 0);
             OFI_CQ_ERROR(shmem_transport_ofi_put_nb_cqfd, &e);
         }
-    } while (success < shmem_transport_ofi_pending_put_counter);
+    } while (success < OPA_load_int(&shmem_transport_ofi_pending_put_counter));
 #else
     int ret = fi_cntr_wait(shmem_transport_ofi_put_cntrfd,
-                           shmem_transport_ofi_pending_put_counter, -1);
+                           OPA_load_int(&shmem_transport_ofi_pending_put_counter), -1);
```
jdinan: I think this needs to be in a loop that repeats the call to ```fi_cntr_wait``` if ```pending_put_counter``` increased during the call.

jdinan: Same comment below with ```get_wait```.

src/transport_ofi.h
```diff
@@ -299,18 +300,18 @@ void shmem_transport_put_quiet(void)
...
+        if (success < OPA_load_int(&shmem_transport_ofi_pending_put_counter) && fail == 0) {
             SPINLOCK_BODY();
         }
         else if (fail) {
             struct fi_cq_err_entry e = {0};
             fi_cq_readerr(shmem_transport_ofi_put_nb_cqfd, (void *)&e, 0);
             OFI_CQ_ERROR(shmem_transport_ofi_put_nb_cqfd, &e);
         }
-    } while (success < shmem_transport_ofi_pending_put_counter);
+    } while (success < OPA_load_int(&shmem_transport_ofi_pending_put_counter));
```
jdinan: I think it's possible for a multithreaded run to have ```success``` be temporarily larger than ```pending_put_counter``` if context switches occur at exactly the wrong time since we increment the pending put counter after issuing the OP.  I suggest that we move incrementing the counter to before issuing the OP to resolve this.

I'm also tempted to suggest that we do a single ```OPA_load``` of the counter into a temporary variable.  Currently the first conditional could evaluate to false and the second to true.

src/transport_ofi.c
jdinan: It reduces portability, but I think the best approach will be to rely on C11 atomics: http://en.cppreference.com/w/c/atomic

We can also write our own wrappers so that we can reduce the overhead in builds where the user disables threads.

davidozog:
I've labelled as "invalid".  Please see #441 instead.

PR #436, opened 2017-09-13
Fix memory leak in target_put

davidozog:
Coverity reported this leak, which can be fixed by moving the free()
into the same scope as the assignment.


kseager:
thanks. If we go this route could we add an if(my_PE_partners) free(my...)? In some cases it won't be allocated.

AKA: line 80/81
int *my_PE_partners = (snode ? get_initiators_partners(metric_info, num_partners): NULL);

davidozog:
I don't mind, but I was actually thinking the same thing until I realized the C spec (sec. 7.20.3.2/2 from ISO-IEC 9899) says free(NULL) takes no action.

PR #441, opened 2017-09-21
Use C11 atomics when available and threading is enabled

davidozog:
If threading is enabled, check whether the C11 standard atomics are supported.  If so, then use those for the OFI transport counters within the shmem_internal_atomics interface; if not, then fall back to the legacy "__sync" atomic builtin functions using uint64 types.  If threading is disabled, the shemm_internal_atomic functions do simple initialization/load/store/arithmetic on uint64 types.

### DIFF
--- a/src/shmem_internal.h
+++ b/src/shmem_internal.h
@@ -386,8 +386,6 @@ extern shmem_internal_mutex_t shmem_internal_mutex_alloc;
 #   define SHMEM_MUTEX_UNLOCK(_mutex)
 #endif /* ENABLE_THREADS */
 
-typedef shmem_atomic_uint64_t shmem_internal_atomic_uint64_t;
-
 void shmem_internal_start_pes(int npes);
 void shmem_internal_init(int tl_requested, int *tl_provided);
 void shmem_internal_finalize(void);
--- a/src/shmem_internal.h
+++ b/src/shmem_internal.h
@@ -384,8 +384,10 @@ extern shmem_internal_mutex_t shmem_internal_mutex_alloc;
 #   define SHMEM_MUTEX_DESTROY(_mutex)
 #   define SHMEM_MUTEX_LOCK(_mutex)
 #   define SHMEM_MUTEX_UNLOCK(_mutex)
-
+typedef uint64_t shmem_atomic_uint64_t;
 #endif /* ENABLE_THREADS */
+
+typedef shmem_atomic_uint64_t shmem_internal_atomic_uint64_t;
 
 void shmem_internal_start_pes(int npes);
 void shmem_internal_init(int tl_requested, int *tl_provided);
--- a/src/shmem_internal.h
+++ b/src/shmem_internal.h
@@ -384,7 +384,6 @@ extern shmem_internal_mutex_t shmem_internal_mutex_alloc;
 #   define SHMEM_MUTEX_DESTROY(_mutex)
 #   define SHMEM_MUTEX_LOCK(_mutex)
 #   define SHMEM_MUTEX_UNLOCK(_mutex)
-typedef uint64_t shmem_atomic_uint64_t;
 #endif /* ENABLE_THREADS */
 
 typedef shmem_atomic_uint64_t shmem_internal_atomic_uint64_t;
--- a/src/transport_ofi.h
+++ b/src/transport_ofi.h
@@ -49,9 +49,9 @@ extern uint8_t**                       shmem_transport_ofi_target_heap_addrs;
 extern uint8_t**                       shmem_transport_ofi_target_data_addrs;
 #endif /* ENABLE_REMOTE_VIRTUAL_ADDRESSING */
 #endif /* ENABLE_MR_SCALABLE */
-extern uint64_t                         shmem_transport_ofi_pending_put_counter;
-extern uint64_t                         shmem_transport_ofi_pending_get_counter;
-extern uint64_t                         shmem_transport_ofi_pending_cq_count;
+extern shmem_internal_atomic_uint64_t   shmem_transport_ofi_pending_put_counter;
+extern shmem_internal_atomic_uint64_t   shmem_transport_ofi_pending_get_counter;
+extern shmem_internal_atomic_uint64_t   shmem_transport_ofi_pending_cq_count;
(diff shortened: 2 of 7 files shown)


src/transport_ofi.h
```diff
@@ -299,18 +299,20 @@ void shmem_transport_put_quiet(void)
...
             struct fi_cq_err_entry e = {0};
             fi_cq_readerr(shmem_transport_ofi_put_nb_cqfd, (void *)&e, 0);
             OFI_CQ_ERROR(shmem_transport_ofi_put_nb_cqfd, &e);
         }
-    } while (success < shmem_transport_ofi_pending_put_counter);
+    } while (success < shmem_internal_atomic_read(&shmem_transport_ofi_pending_put_counter));
 #else
     int ret = fi_cntr_wait(shmem_transport_ofi_put_cntrfd,
-                           shmem_transport_ofi_pending_put_counter, -1);
+                           shmem_internal_atomic_read(&shmem_transport_ofi_pending_put_counter), -1);
```
jdinan: I think we need to check if the pending put counter increased while we were blocked.  I'm thinking of something like:
```
uint64_t cnt, cnt_new;
cnt_new = shmem_internal_atomic_read(&shmem_transport_ofi_pending_put_counter), -1);
do {
  int ret;
  cnt = cnt_new;
  ret = fi_cntr_wait(shmem_transport_ofi_put_cntrfd, &cnt, -1);
  ...
  cnt_new = shmem_internal_atomic_read(&shmem_transport_ofi_pending_put_counter), -1);
} while (cnt != cnt_new);
```

src/transport_ofi.h
```diff
@@ -554,18 +557,20 @@ void shmem_transport_get_wait(void)
...
             struct fi_cq_err_entry e = {0};
             fi_cq_readerr(shmem_transport_ofi_put_nb_cqfd, (void *)&e, 0);
             OFI_CQ_ERROR(shmem_transport_ofi_put_nb_cqfd, &e);
         }
-    } while (success < shmem_transport_ofi_pending_get_counter);
+    } while (success < shmem_internal_atomic_read(&shmem_transport_ofi_pending_get_counter));
 #else
     int ret = fi_cntr_wait(shmem_transport_ofi_get_cntrfd,
-                           shmem_transport_ofi_pending_get_counter, -1);
+                           shmem_internal_atomic_read(&shmem_transport_ofi_pending_get_counter), -1);
```
jdinan: Same comment as above, we need to check if the pending counter increased while we were blocked.

src/transport_ofi.h
```diff
@@ -299,18 +299,20 @@ void shmem_transport_put_quiet(void)
...
+        if (success < \
+            shmem_internal_atomic_read(&shmem_transport_ofi_pending_put_counter) && \
+            fail == 0) {
             SPINLOCK_BODY();
         }
         else if (fail) {
             struct fi_cq_err_entry e = {0};
             fi_cq_readerr(shmem_transport_ofi_put_nb_cqfd, (void *)&e, 0);
             OFI_CQ_ERROR(shmem_transport_ofi_put_nb_cqfd, &e);
         }
```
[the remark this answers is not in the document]
jdinan: It reduces the number of atomic reads (which are expensive).  Same comment below in get-wait.

davidozog:
Thanks for the review.  I agree with everything, but I'm also running into very concerning issues with Clang on OSX with this whole approach.  There seem to be several implicit conversions from _Atomic types to non-atomic types that clang does not approve of at all...  So far I don't know of a work-around.

jdinan:
Oh, that's interesting.  So, the ```shmem_internal_atomic_read()``` routines have different return types based whether you're using C11 atomics or compiler intrinsics.  They should definitely both return uint64_t.

jdinan:
Thinking a bit more about the rationale for the counter comparison in the completion ops -- on every operation we increment the put/get issued counter before the FI call (i.e. before the FI event counter is incremented).  When checking for completion, we check the counters in the reverse order -- FI event counter then put/get issued counter.  Might be worthwhile to add a comment to that effect to the completion routines so we remember to preserve that property.

src/shmem_internal.h
```diff
@@ -384,9 +384,11 @@ extern shmem_internal_mutex_t shmem_internal_mutex_alloc;
 #   define SHMEM_MUTEX_DESTROY(_mutex)
 #   define SHMEM_MUTEX_LOCK(_mutex)
 #   define SHMEM_MUTEX_UNLOCK(_mutex)
-
+typedef uint64_t shmem_atomic_uint64_t;
```
jdinan: Delete me

jdinan:
```
In file included from synchronization_c.c:28:0:
../../src/shmem_atomic.h:147:18: warning: redefinition of typedef ‘shmem_atomic_uint64_t’ [-Wpedantic]
 typedef uint64_t shmem_atomic_uint64_t;
                  ^
In file included from synchronization_c.c:27:0:
../../src/shmem_internal.h:387:18: note: previous declaration of ‘shmem_atomic_uint64_t’ was here
 typedef uint64_t shmem_atomic_uint64_t;
                  ^
  CC       collectives_f.lo
In file included from ../../src/shmem_synchronization.h:19:0,
                 from ../../src/shmem_collectives.h:19,
                 from collectives_f.c:32:
../../src/shmem_atomic.h:147:18: warning: redefinition of typedef ‘shmem_atomic_uint64_t’ [-Wpedantic]
 typedef uint64_t shmem_atomic_uint64_t;
                  ^
In file included from collectives_f.c:30:0:
../../src/shmem_internal.h:387:18: note: previous declaration of ‘shmem_atomic_uint64_t’ was here
 typedef uint64_t shmem_atomic_uint64_t;
```

davidozog: Hmm... Travis/Ubuntu gcc doesn't like this when I remove it (and no threading), which is probably why I threw a second typedef in there.  I'll figure something out.

jdinan:
Can we replace ```shmem_atomic_uint64_t``` with ```shmem_internal_atomic_uint64_t``` to get this down to one new type?

jdinan:
```shmem_internal_atomic_read``` doesn't have a consistent return type.  Can we update all of the functions to return ```uint64_t```?

PR #467, opened 2017-09-29
Add support for 1.4 types

jdinan:
* Add support for fixed width and standard integer types
* Add type compatibility detection for populating C11 generic and C++ overloaded interfaces
* Update C11 tests to include new types
* Move bitwise AMOs to shmem_ namespace
* Add generic selection interface for bitwise AMOs (testing coverage still needed)
* Add profiling interfaces foe bitwise AMOs


mpp/shmem_c_func.m4
```diff
@@ -156,6 +156,32 @@ define(`SHMEM_C_SET',
...
+SHMEM_DECLARE_FOR_BITWISE_AMO(`SHMEM_C_FETCH_XOR')
+
+define(`SHMEM_C_FETCH_AND',
+`SHMEM_FUNCTION_ATTRIBUTES $2 SHPRE()shmem_$1_atomic_fetch_and($2 *target, $2 value, int pe)')dnl
+SHMEM_DECLARE_FOR_BITWISE_AMO(`SHMEM_C_FETCH_AND')
+
+define(`SHMEM_C_FETCH_OR',
+`SHMEM_FUNCTION_ATTRIBUTES $2 SHPRE()shmem_$1_atomic_fetch_or($2 *target, $2 value, int pe)')dnl
+SHMEM_DECLARE_FOR_BITWISE_AMO(`SHMEM_C_FETCH_OR')
+
```
davidozog: I don't see any evidence of fetch_add and fetch_inc.  Are those accounted for in `SHMEM_C_FETCH`?   Elsewhere?

davidozog: Maybe all of the atomic_* routines are not address in this commit...  Just the bitwise?

bindings/shmem_bind_c.m4
```diff
@@ -11,11 +11,27 @@ define(`SHMEM_BIND_C_RMA', dnl args: macro_name, end
 `$1(float,      float)$2
 $1(double,     double)$2
 $1(longdouble, long double)$2
+$1(schar,      signed char)$2
```
davidozog: Extremely minor, but perhaps `char` could be before `signed char` to match the spec table and the unit tests...

mpp/shmem_c_func.m4
jdinan: This is just the new types.  I pulled bitwise AMOs in since they were already done and I needed them to test part of the bindings generators.

Issue #439, opened 2017-09-18
complex_reductions_f imaginary test failure


jdinan:
The complex_reductions_f test fails with 10 processes:
```
$ mpiexec -np 10 test/unit/complex_reductions_f
 fail : incorrect imaginary component   -10421575.0      expected   -10421576.0      on process            9 test #           5
```

jdinan:
Additional failures with 11-16 PEs:
```
$ mpiexec -np 11 test/unit/complex_reductions_f
 fail : incorrect real component   -187999168.      expected   -187999184.      on process            0 test #           5

$ mpiexec -np 12 test/unit/complex_reductions_f
 fail : incorrect real component    3.17739981E+09  expected    3.17740006E+09  on process            9 test #           5

$ mpiexec -np 13 test/unit/complex_reductions_f
 fail : incorrect imaginary component   -4.34649825E+10  expected   -4.34649866E+10  on process            0 test #           5
 fail : incorrect imaginary component   -4.34649825E+10  expected   -4.34649784E+10  on process           12 test #           5
 fail : incorrect imaginary component   -4.34649825E+10  expected   -4.34649866E+10  on process            3 test #           5

$ mpiexec -np 14 test/unit/complex_reductions_f
 fail : incorrect real component    2.29804474E+12  expected    2.29804501E+12  on process            0 test #           5

$ mpiexec -np 15 test/unit/complex_reductions_f
 fail : incorrect real component    1.33558111E+13  expected    1.33558132E+13  on process            0 test #           5

$ mpiexec -np 16 test/unit/complex_reductions_f
 fail : incorrect real component   -6.32191537E+14  expected   -6.32191470E+14  on process            0 test #           5
```

jdinan:
This appears to be an associativity issue -- the SHMEM reduction is reducing elements in a different order than the unit test.  Need to update the validation check to fail only of the percent difference is greater than 1%.

PR #468, opened 2017-09-29
Loosen threshold to pass Fortran complex reduction

davidozog:
Instead of `epsilon`, use 1/1000th of a percent (1e-5) tolerance when checking for correctness in the Fortran complex reduction unit test.  We need this because there's a tiny bit of error at relatively large scale, likely because of the numerical non-associativity of parallel reductions.

Addresses issue #439.

### DIFF
--- a/test/unit/complex_reductions_f.f90
+++ b/test/unit/complex_reductions_f.f90
@@ -202,20 +202,21 @@
 
       contains
 
+      ! Checks that real and imaginary components are within 1/1000th of
+      ! a percent of the expected value:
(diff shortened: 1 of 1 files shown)


jdinan:
Did you verify that the issue flagged in #439 is resolved?  We don't check for it in CI testing, since the PE counts used there are too small.

davidozog:
Yup.  I was able to reproduce the error and fix it on my Mac.  The test now passes with a large # of PEs (I've tested up to 50).

Issue #460, opened 2017-09-28
Configury check for data_start


jdinan:
We have an implicit requirement on Linux that the compiler defines a ```data_start``` symbol.  We should really be checking that this code is reasonable at build time (otherwise, the library will build successfully but fail testing).  A ```AC_TRY_RUN``` test should do the trick.  Something like:

```
#include <stdio.h>
#ifdef __APPLE__
#include <mach-o/getsect.h>
#else
extern char data_start;
extern char end;
#endif

int main(void) {
  void *base;
  unsigned long length;
#ifdef __APPLE__
    base = (void*) get_etext();
    length = get_end() - get_etext();
#else
    base = &data_start;
    length = (unsigned long) &end  - (unsigned long) &data_start;
#endif
    printf("%p %lu\n", base, length);
    return 0;
}
```

PR #469, opened 2017-09-29
Configury: Add check for data segment symbol

jdinan:
Check for presence of 'data_start' and 'end' symbols on Linux.  Error
output looks like:

  checking for data segment pointer... not found
  configure: error: Could not locate data segment


jdinan:
Closes #460

PR #479, opened 2017-10-10
 Add m4/macros for _ctx_ routines to C bindings

davidozog:
Adds preliminary C bindings for contexts with dummy definitions for `SHMEM_CTX_DEFAULT` and `shmem_ctx_t`.

### DIFF
--- a/src/atomic_c.c4
+++ b/src/atomic_c.c4
@@ -37,10 +37,20 @@ define(`SHMEM_PROF_DEF_SWAP',
 #define shmem_$1_swap pshmem_$1_swap')dnl
 SHMEM_DEFINE_FOR_EXTENDED_AMO(`SHMEM_PROF_DEF_SWAP')
 
+define(`SHMEM_PROF_DEF_CTX_SWAP',
+`#pragma weak shmem_ctx_$1_swap = pshmem_ctx_$1_swap
+#define shmem_ctx_$1_swap pshmem_ctx_$1_swap')dnl
+SHMEM_DEFINE_FOR_EXTENDED_AMO(`SHMEM_PROF_DEF_CTX_SWAP')
+
 define(`SHMEM_PROF_DEF_ATOMIC_SWAP',
 `#pragma weak shmem_$1_atomic_swap = pshmem_$1_atomic_swap
 #define shmem_$1_atomic_swap pshmem_$1_atomic_swap')dnl
 SHMEM_DEFINE_FOR_EXTENDED_AMO(`SHMEM_PROF_DEF_ATOMIC_SWAP')
+
+define(`SHMEM_PROF_DEF_CTX_ATOMIC_SWAP',
+`#pragma weak shmem_ctx_$1_atomic_swap = pshmem_ctx_$1_atomic_swap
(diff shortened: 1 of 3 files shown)


src/atomic_c.c4
```diff
@@ -54,107 +64,209 @@ define(`SHMEM_PROF_DEF_CSWAP',
...
 #endif /* ENABLE_PROFILING */
 
+
+
 #define SHMEM_DEF_SWAP(STYPE,TYPE,ITYPE)                        \
     TYPE SHMEM_FUNCTION_ATTRIBUTES                              \
-    shmem_##STYPE##_swap(TYPE *target, TYPE value, int pe)      \
-    {                                                           \
+    SHMEM_FUNC_PROTOTYPE(STYPE##_swap, TYPE *target,            \
+                         TYPE value, int pe)                    \
```
jdinan: I _think_ you we merge the _atomic* generators if you move the STYPE and op pasting into the prototype macro (e.g. invoke as ```SHMEM_FUNC_PROTOTYPE(STYPE, swap, ...)```) and define:

```#define SHMEM_FUNC_PROTOTYPE(TYPE, OP, ...) shmem_##TYPE##_##OP(__VA_ARGS__)```
```#define SHMEM_FUNC_PROTOTYPE(TYPE, OP, ...) shmem_##TYPE##_atomic_##OP(__VA_ARGS__)```
```#define SHMEM_FUNC_PROTOTYPE(TYPE, OP, ...) shmem_ctx_##TYPE##_##OP(__VA_ARGS__)```
```#define SHMEM_FUNC_PROTOTYPE(TYPE, OP, ...) shmem_ctx_##TYPE##_atomic_##OP(__VA_ARGS__)```

File: src/shmem_internal.h [hunk 2]
```diff
@@ -44,6 +44,13 @@ extern long shmem_internal_data_length;
 /* Note: must be accompanied by shmem_internal_my_pe in arguments */
 #define RAISE_PE_PREFIX "[%04d]        "
 
+
+/* Function prototype for standard routines with the default context: */
+#define SHMEM_FUNC_PROTOTYPE(FUNCNAME,  ...)        \
+  shmem_##FUNCNAME(__VA_ARGS__) {                   \
+  const shmem_ctx_t ctx = SHMEM_CTX_DEFAULT;
+
+
```
jdinan: I'd prefer to move this into the .c files to where it is used (and has limited lifespan between define/undefine), rather than defining it at a top level across all files.

davidozog: Yeah I think so too.  I originally had this, but thought it was slightly weird to have it duplicated in `atomic_c.c4` and `data_c.c4`...

src/atomic_c.c4
davidozog: Yeah, I was thinking that too.  I'll give it a shot.

davidozog:
Whoops, that commit is wrong.  I'll fix that if'n you don't mind.

src/atomic_c.c4
davidozog: I think this works, but `finc`/`fetch_inc` and `cswap`/`compare_swap` are exceptions, right?

davidozog:
@jdinan Is this one good to merge?

PR #480, opened 2017-10-12
adding optimization for self-access

ireed:
Add another intranode option (in addition to XPMEM and CMA) to enable self-communication optimization. 

Side note: i dont understand how or why it shows commits for the mellanox tests, but the commits seem to only add an extra blank line to the travis script. If someone knows how to remove or ignore that commit then i'd love to know how.


lanl-ompi:

jdinan:
Replaced by #484

PR #484, opened 2017-10-12
Add memcpy "transport"

jdinan:
Replaces #480


davidozog:
If the user passes --with-xpmem (or --with-cma) as well as --enable-memcpy, then this will (kinda quietly) drop the memcpy, right?  Is that ok?  Perhaps a warning would be good during configure, but it's no big deal.  Otherwise looks good.

ireed:
If xpmem or cma is enabled then the new code should never be reached.

davidozog:
Right but if both xpmem (or cma) _and_ memcpy are enabled, then the new code is not reached either.  I'm just saying a warning in configure might be good so users know that xpmem/cma are prioritized if they accidentally request memcpy as well.

ireed:
.... well, unless you use compile tricks such as setting -DUSE_SELF_OPTIMIZATION.

davidozog:
On second thought, we could just print "transport_memcpy = yes" at the end of configure if any of the the shared memory transports are enabled, but only `USE_MEMCPY` if it alone is enabled (Jim's idea).  I'll push the change now and you guys can review.

PR #489, opened 2017-10-19
Add new sync routines

jdinan:


jdinan:
@davidozog This is ready for review.

I'll also file an issue to circle back and deal with the quiet before exiting dissemination barrier.  This can't be replaced by a wait because of back-to-back barriers.  Once we have contexts, we can add a collectives context that will solve this problem.  Several other collectives also use quiet and this would help them, too.

jdinan:
Created issue #490 to remind us to fix quiet in collectives.

PR #494, opened 2017-10-20
Add unit tests and fill out contexts API

jdinan:
* Add OpenMP support
* Move thread safety API from shmemx_ to shmem_
* Add contexts constants
* Add unit tests from jdinan/contexts

TODO: Enable threading configury by default


test/unit/web.c
```diff
@@ -0,0 +1,227 @@
...
+    free(threads);
+
+    for(i = 0; i < NUM_CONTEXTS; ++i) {
+      shmem_ctx_quiet(contexts[i]);
+      shmem_ctx_destroy(contexts[i]);
+      int err;
+      err = pthread_mutex_destroy(&ctx_locks[i]);
+      CHECK_ASSERT(!err);
+    }
+
```
davidozog: Do you need a `shmem_sync_all` here?  If not, I don't quite understand why it's needed at the end of  `many-ctx.c` and not here...

test/apps/mandelbrot.c
```diff
@@ -0,0 +1,507 @@
...
+#include <utmpx.h>
+#include <unistd.h>
+#include <assert.h>
+
+#ifdef ENABLE_PAPI
+#include <papi.h>
+#endif
+
+#include <shmem.h>
+#include "../unit/pthread_barrier.h" /* FIXME -- this is for MacOS */
```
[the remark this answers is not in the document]
jdinan: They are both ok, but the double quotes are [slightly better](https://stackoverflow.com/questions/3162030/difference-between-angle-bracket-and-double-quotes-while-including-heade).  I'll fix this for both cases and also fix the include paths.

test/apps/mandelbrot.c
```diff
@@ -0,0 +1,507 @@
...
+    int out_file = 0;
+    pthread_t *threads;
+    struct th_arg *t_arg;
+
+#ifdef ENABLE_PINNING
+    int p4_cpu;
+    pthread_t thread = pthread_self();
+    cpu_set_t cpuset;
+#endif
+
```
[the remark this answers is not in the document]
jdinan: I'd prefer to leave this as-is, since there are different ways of achieving the pinning (e.g. via ```mpiexec``` or explicitly via ```pthreads_setaffinity```) and the right solution may depend on whether threads are spawned lower down in the networking stack.

test/unit/web.c
jdinan: Yup, good catch!

jdinan: Scratch that.  Web takes care of this via the ```received``` that is waited on in web.c:124.

jdinan:
@davidozog Thanks for the feedback.  Updates pushed; should be ready to go.

PR #473, opened 2017-10-03
Generic C++ bindings for bitwise AMOs

jdinan:

PR #499, opened 2017-10-20
Generic C++ bindings for bitwise AMOs

jdinan:


jdinan:
Reincarnation of #473

PR #502, opened 2017-10-23
Update atomic_fetch_add to use AMO type class

jdinan:
Was using extended AMO.  Spec calls for AMO.


jdinan:
Merging with minimal testing.

PR #504, opened 2017-10-23
Propagate context argument through the transport layer

davidozog:
Add the `shmem_ctx_t ctx` argument to each routine in the transport layer (for ofi, portals, and none).

### DIFF
--- a/src/transport_none.c
+++ b/src/transport_none.c
@@ -10,5 +10,6 @@
  */
 
 #include "transport_none.h"
+#include "shmem.h"
 
 void *SHMEM_CTX_DEFAULT = NULL;
--- a/src/transport_none.c
+++ b/src/transport_none.c
@@ -11,5 +11,4 @@
 
 #include "transport_none.h"
 
-void *SHMEMX_CTX_DEFAULT = NULL;
-void *SHMEMX_DOMAIN_DEFAULT = NULL;
+void *SHMEM_CTX_DEFAULT = NULL;
--- a/src/transport_none.c
+++ b/src/transport_none.c
@@ -1,0 +1,15 @@
+/* -*- C -*-
+ *
+ * Copyright (c) 2016 Intel Corporation. All rights reserved.
+ * This software is available to you under the BSD license.
+ *
+ * This file is part of the Sandia OpenSHMEM software package. For license
+ * information, see the LICENSE file in the top level directory of the
+ * distribution.
+ *
+ */
+
+#include "transport_none.h"
+
+void *SHMEMX_CTX_DEFAULT = NULL;
+void *SHMEMX_DOMAIN_DEFAULT = NULL;
--- a/src/transport_none.h
+++ b/src/transport_none.h
@@ -58,7 +58,6 @@ typedef int shmem_transport_ct_t;
 struct shmem_transport_ctx_t{ int dummy; };
 
 typedef struct shmem_transport_ctx_t shmem_transport_ctx_t;
-extern void * SHMEM_CTX_DEFAULT;
 
 static inline
 int
--- a/src/transport_none.h
+++ b/src/transport_none.h
@@ -55,10 +55,9 @@ typedef int shm_internal_datatype_t;
 typedef int shm_internal_op_t;
 typedef int shmem_transport_ct_t;
 
-struct shmem_transport_ctx_t{};
+struct shmem_transport_ctx_t{ int dummy };
 
 typedef struct shmem_transport_ctx_t shmem_transport_ctx_t;
-shmem_transport_ctx_t shmem_transport_ctx_default;
 
 static inline
 int
--- a/src/transport_none.h
+++ b/src/transport_none.h
@@ -55,7 +55,7 @@ typedef int shm_internal_datatype_t;
 typedef int shm_internal_op_t;
 typedef int shmem_transport_ct_t;
 
-struct shmem_transport_ctx_t{ int dummy };
+struct shmem_transport_ctx_t{ int dummy; };
 
 typedef struct shmem_transport_ctx_t shmem_transport_ctx_t;
 
--- a/src/transport_none.h
+++ b/src/transport_none.h
@@ -55,6 +55,11 @@ typedef int shm_internal_datatype_t;
 typedef int shm_internal_op_t;
 typedef int shmem_transport_ct_t;
 
+struct shmem_transport_ctx_t{};
+
+typedef struct shmem_transport_ctx_t shmem_transport_ctx_t;
+shmem_transport_ctx_t shmem_transport_ctx_default;
+
 static inline
 int
 shmem_transport_init(void)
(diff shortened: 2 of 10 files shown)


src/transport_none.h
```diff
@@ -78,74 +78,74 @@ shmem_transport_fini(void)
 
 static inline
 int
-shmem_transport_quiet(void)
+shmem_transport_quiet(shmem_ctx_t ctx)
```
jdinan: You might want the type to be ```shmem_transport_ctx_t``` here to avoid casting all over the place in the implementation.  In the internal routines, you can cast the ```shmem_ctx_t ``` to ```shmem_transport_ctx_t *```.

src/transport_ofi.c
```diff
@@ -1214,7 +1214,7 @@ int shmem_transport_startup(void)
 int shmem_transport_fini(void)
 {
     /* Wait for acks before shutdown */
-    shmem_transport_quiet();
+    shmem_transport_quiet(SHMEM_CTX_DEFAULT);
```
jdinan: Once you introduce ```shmem_transport_ctx_default```, you'll want to use that here in place of ```SHMEM_CTX_DEFAULT```.  Otherwise, the types won't match (```SHMEM_CTX_DEFAULT``` is ```void *``` and ```shmem_transport_ctx_default``` should be of type ```shmem_transport_ctx_t *``` ).

If you make the type change I suggested, you'll want to also introduce the new variable in this PR to resolve the type issue.

src/transport_ofi.h
```diff
@@ -214,10 +214,10 @@ int shmem_transport_fini(void);
 
 extern size_t SHMEM_Dtsize[FI_DATATYPE_LAST];
 
-static inline void shmem_transport_get_wait(void);
+static inline void shmem_transport_get_wait(shmem_ctx_t ctx);
 
 static inline
-void shmem_transport_ofi_drain_cq(void)
+void shmem_transport_ofi_drain_cq()
```
jdinan: Removing void here will cause GCC to produce a warning about the function not being properly prototyped.

src/transport_portals4.h
```diff
@@ -336,7 +336,7 @@ shmem_transport_portals4_fence_complete(void)
 
 static inline
 void
-shmem_transport_portals4_drain_eq(void)
+shmem_transport_portals4_drain_eq()
```
jdinan: Same comment about removing ```void``` in the arguments.

src/transport_ofi.h
davidozog: Whoops!

src/transport_none.h
```diff
@@ -55,6 +55,11 @@ typedef int shm_internal_datatype_t;
 typedef int shm_internal_op_t;
 typedef int shmem_transport_ct_t;
 
+struct shmem_transport_ctx_t{};
```
jdinan: Woah, does this work?  First time I've seen it.

jdinan: I'm wondering if we should just declare it to be ```void *``` since it's never dereferenced in transport-none (maybe same in the Portals transport)?

jdinan:
Looks like the empty structure has undefined behavior in the C language spec, but is supported as a GCC extension: https://stackoverflow.com/questions/3849334/sizeof-empty-structure-is-0-in-c-and-1-in-c-why

src/transport_none.h
davidozog: Haha, yeah.  I was kinda surprised too, but it seems to work.  But yes, I'll just do `void *` since it simpler/probably more portable.

davidozog:
Oh wow, thanks for the info on empty structs... 
I'll add a dummy variable for now so we can confidently merge this after a few Travis rows pass.  Then I'll do a separate PR to change the empty structs to `void *`, since Travis seems abnormally sluggish at the moment...

src/transport_none.h
```diff
@@ -55,10 +55,9 @@ typedef int shm_internal_datatype_t;
 typedef int shm_internal_op_t;
 typedef int shmem_transport_ct_t;
 
-struct shmem_transport_ctx_t{};
+struct shmem_transport_ctx_t{ int dummy };
 
 typedef struct shmem_transport_ctx_t shmem_transport_ctx_t;
-shmem_transport_ctx_t shmem_transport_ctx_default;
```
jdinan: Depending on how we define SHMEM_CTX_DEFAULT in the public header, we will likely need to add a ```transport_none.c``` file to define a symbol for it.  In the prototype, I think we declared ```extern void * SHMEM_CTX_DEFAULT``` in ```shmem.h``` and then defined ```shmem_transport_ctx_t SHMEM_CTX_DEFAULT``` in the .c file for each transport.

davidozog:
The failing tests are due to and undefined "SHMEM_CTX_DEFAULT" when it's marked as `extern` in the public header.  When I remove `extern`, this error goes away when I build locally.  But the `extern` is what makes sense to me... any idea what's wrong?

src/transport_none.c
```diff
@@ -0,0 +1,15 @@
...
+ * This file is part of the Sandia OpenSHMEM software package. For license
+ * information, see the LICENSE file in the top level directory of the
+ * distribution.
+ *
+ */
+
+#include "transport_none.h"
+
+void *SHMEMX_CTX_DEFAULT = NULL;
+void *SHMEMX_DOMAIN_DEFAULT = NULL;
```
jdinan: Rename SHMEMX -> SHMEM
Also, delete SHMEMX_DOMAIN_DEFAULT

jdinan: It looks like transport_none is also missing the ```extern void * SHMEM_CTX_DEFAULT;``` declaration.

davidozog: Arg.. committed the wrong file.

jdinan:
Looks good now.  Once Travis gives it the ok, go ahead and merge.  Travis should build (but not test) both transport-none and transport-portals.

davidozog:
Hmm... apparently I'm still missing something with the visibility of the SHMEM_CTX_DEFAULT symbol.  The mandelbrot test is still failing in Travis.

src/transport_none.h
```diff
@@ -55,6 +55,11 @@ typedef int shm_internal_datatype_t;
 typedef int shm_internal_op_t;
 typedef int shmem_transport_ct_t;
 
+struct shmem_transport_ctx_t{ int dummy; };
+
+typedef struct shmem_transport_ctx_t shmem_transport_ctx_t;
+extern void * SHMEM_CTX_DEFAULT;
```
jdinan: Delete me -- this shadows the declaration in ```shmem.h``` and has the wrong visibility.

jdinan: Looks like this is the cause of the Travis testing failures.

davidozog: Oh yeah, the problem is in transport_none - at least I can reproduce it now.  

But when I remove that line (61) I still get the undefined reference in Ubuntu locally...  I'll try a push through Travis anyway.

davidozog: If I give it that visibility attribute, however, all is good...  Is that what this needs?

jdinan: Maybe ```#include "shmem.h"``` in ```transport_none.c```?  Both of the other transports do this, so they are pulling in the declaration with the visibility attribute.

davidozog: Oh jeez, that's got to be it.  Thanks!

Issue #501, opened 2017-10-23
Threading should enable hard polling


jdinan:
The transport wait should be disabled when threading is enabled, since threads can wake up other threads from calls to ```shmem_wait``` with a store.  Said the other way, threading should imply hard polling.

jdinan:
We also need a unit test for this.

PR #505, opened 2017-10-23
Multithreaded shmem_wait

jdinan:
* Allow threads to wake each other up from calls shmem_wait
* Add new return value from shmem_init_thread


davidozog:
Looks nice, and closes #501.  The only thing that catches my attention is that none of the other unit tests check the return value of `shmem_init_thread`, which is apparently fine since it can only return 0 or finalize.

jdinan:
Oh, good catch.  I'll update those other tests.

PR #510, opened 2017-10-25
Adding the check for shmem_info_get_name() to be equivaluent with SHM…

wrrobin:
…EM_VENDOR_STRING in the unit test shmem_info


lanl-ompi:

PR #514, opened 2017-10-31
Removed volatile specifier from wait and lock APIs

wrrobin:


lanl-ompi:

jdinan:
That's weird.  I'm not sure why we didn't see this error on the earlier PR that added C++ bindings for ```shmem_wait_until```.  This is easily fixed; I'll push a patch.

```
In file included from ../../../test/unit/cxx_test_shmem_fetch.cpp:37:0:
../../mpp/shmem.h: In function ‘void shmem_wait_until(long int*, shmem_cmp_t, long int)’:
../../mpp/shmem.h:2550:80: error: ‘void shmem_wait_until(long int*, shmem_cmp_t, long int)’ was declared ‘extern’ and later ‘static’ [-fpermissive]
 static inline void shmem_wait_until(long *ivar, shmem_cmp_t cmp, long cmp_value) {
                                                                                ^
../../mpp/shmem.h:1580:32: error: previous declaration of ‘void shmem_wait_until(long int*, shmem_cmp_t, long int)’ [-fpermissive]
 SHMEM_FUNCTION_ATTRIBUTES void shmem_wait_until(long *ivar, shmem_cmp_t cmp, long value);
                                ^
make[3]: *** [cxx_test_shmem_fetch.o] Error 1
make[3]: *** Waiting for unfinished jobs....
```

jdinan:
@wrrobin 70af37d should hopefully resolve the compilation error.

@nspark Looks like the new generic ```shmem_wait_until``` bindings overlap with the legacy ```shmem_wait_until``` routine.  Should we deprecate the old routine?  IIUC, existing programs should be unaffected as long they have access to C11 generic interfaces.

wrrobin:
@jdinan Thanks. Do you think removing the ``volatile`` specifier from the internal functions in shmem_lock.h would have any implications? I was kind of in doubt for these functions. Just making sure if I thought it right.

nspark:
@jdinan Good point. I hadn't thought about how the subsumption of `shmem_wait_until(long*, ...)` by `shmem_wait_until(TYPE*, ...)` doesn't apply under a C99 translation environment. In that case, I think deprecating the explicit `shmem_wait_until(long*, ...)` interface makes sense.

jdinan:
@nspark This is fairly benign for C, as long as the legacy ```shmem_wait_until``` routine is declared prior to the definition of the C11 generic ```shmem_wait_until``` macro.  The problem popped up for us in SOS because we also use function overloading to provide a type-generic C++ interface.  In that case, the C++ bindings try to redefine this routine as ```static inline```, which breaks compilation.

So, technically, it should be fine to leave the ```shmem_wait_until``` declaration in place in the C11 translation environment.  As long as I have to disable it for C++, it seems sensible to disable it in C11 for extra robustness.

In terms of the spec, we would need to delay ratification to deprecate this routine.  However, deprecation doesn't really fix the problem since the routine is still required by the implementation.

PR #519, opened 2017-11-07
First draft of ctx create/destroy

davidozog:
Highlights:
- Adds `shmem_ctx_create()` and `shmem_ctx_destroy()` for OFI, based on the [prototype](https://github.com/jdinan/sandia-shmem/tree/contexts).
- Removes the (presumably) extraneous [cntr_info object](https://github.com/jdinan/sandia-shmem/blob/contexts/src/transport_ofi.c#L1252) and renames it to `shmem_transport_ofi_info`.
- Adds a couple endpoints to the ctx struct

To do:
- Initialize the default context with the same routine as user contexts (`shmem_transport_ofi_ctx_init`).
- Do resource cleanup upon error from shmem_ctx_create and address [JD's comments](https://github.com/Sandia-OpenSHMEM/SOS/compare/v1.4.0-dev...davidozog:pr/ctx_create_destroy?expand=1#diff-02eef8be694d26121d1275651d32e9a5R1303).
- Assure this works with latest libfabric:master (`many-ctx` and `web.c` tests currently seem to have progress issues in the sockets provider up to and including libfabric 1.5.1).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -71,7 +71,6 @@ before_install:
     - export TRAVIS_INSTALL=$HOME/travis/install
     ## Disable security protection so CMA will work
     - sudo sysctl -w kernel.yama.ptrace_scope=0
-    - ulimit -n 2048
     ## Run the icc installation script:
     - >
       if [ "$CC" = "icc" ]; then
--- a/.travis.yml
+++ b/.travis.yml
@@ -71,6 +71,7 @@ before_install:
     - export TRAVIS_INSTALL=$HOME/travis/install
     ## Disable security protection so CMA will work
     - sudo sysctl -w kernel.yama.ptrace_scope=0
+    - ulimit -n 2048
     ## Run the icc installation script:
     - >
       if [ "$CC" = "icc" ]; then
--- a/.travis.yml
+++ b/.travis.yml
@@ -100,7 +100,7 @@ before_install:
     - make install
     ## Build libfabric
     - cd $TRAVIS_SRC
-    - git clone --depth 10 https://github.com/ofiwg/libfabric.git libfabric
+    - git clone -b v1.5.1 --depth 10 https://github.com/ofiwg/libfabric.git libfabric
     - cd libfabric
     - ./autogen.sh
     - ./configure --prefix=$TRAVIS_INSTALL/libfabric
--- a/src/transport_none.h
+++ b/src/transport_none.h
@@ -81,6 +81,13 @@ shmem_transport_fini(void)
 }
 
 static inline
+void
+shmem_transport_ctx_create(shmem_transport_ctx_t **ctx)
+{
+    return;
+}
+
+static inline
 int
 shmem_transport_quiet(shmem_transport_ctx_t* ctx)
 {
--- a/src/transport_none.h
+++ b/src/transport_none.h
@@ -88,6 +88,13 @@ shmem_transport_ctx_create(shmem_transport_ctx_t **ctx)
 }
 
 static inline
+void
+shmem_transport_ctx_destroy(shmem_transport_ctx_t *ctx)
+{
+    return;
+}
+
+static inline
 int
 shmem_transport_quiet(shmem_transport_ctx_t* ctx)
 {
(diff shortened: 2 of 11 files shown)


davidozog:
@jdinan - ready for review.  The commit history is a little messy, so feel free to squash.

src/transport_ofi.c
```diff
@@ -1211,24 +1283,104 @@ int shmem_transport_startup(void)
...
+    if (fi_close(&ctx->get_cntrfd->fid)) {
+        RAISE_ERROR_MSG("Context counter close failed (%s)\n", fi_strerror(errno));
+    }
+
+    if (ctx->id >= 0) {
+        SHMEM_MUTEX_LOCK(shmem_transport_ofi_lock);
+        shmem_transport_ofi_contexts[ctx->id] = NULL;
+        SHMEM_MUTEX_UNLOCK(shmem_transport_ofi_lock);
+        free(ctx);
+    }
```
jdinan: Else: error?

src/transport_ofi.c
```diff
@@ -1211,24 +1283,104 @@ int shmem_transport_startup(void)
...
+    /* FIXME: This does not do resource cleanup 
+     * on error, it just returns. This is consistent with how
+     * initialization worked before, but is worse since context creation
+     * is not necessarily do-or-die.
+     */
+    /* JD: This needs to have a mutex around it to synchronize a possible
+     * update to the contexts array.
+     *
+     * This also does not reuse entries from contexts that were freed.
+     */
```
jdinan: The part about the mutex can be deleted now that this is guarded by ```shmem_transport_ofi_lock```, but the part about leaking entries is still a valid FIXME.

.travis.yml
```diff
@@ -71,6 +71,7 @@ before_install:
     - export TRAVIS_INSTALL=$HOME/travis/install
     ## Disable security protection so CMA will work
     - sudo sysctl -w kernel.yama.ptrace_scope=0
+    - ulimit -n 2048
```
jdinan: Our users are going to run into the same limitation.  Might be better to reduce the number of contexts created in the unit tests instead.

src/transport_ofi.c
```diff
@@ -1107,37 +1115,103 @@ int query_for_fabric(struct fabric_info *info)
...
 }
 
+static int shmem_transport_ofi_ctx_init(shmem_transport_ctx_t *ctx, int id)
+{
+    int ret = 0;
+
+    struct fi_cntr_attr cntr_attr = {0};
+    cntr_attr.events   = FI_CNTR_EVENTS_COMP;
+    /* Set FI_WAIT based on the put and get polling limits defined above */
+    if (shmem_transport_ofi_put_poll_limit < 0) {
```
jdinan: Should the get polling limit also be checked here?

src/transport_ofi.c
```diff
@@ -1107,37 +1115,103 @@ int query_for_fabric(struct fabric_info *info)
...
+    }
+
+    ret = fi_cntr_open(shmem_transport_ofi_domainfd, &cntr_attr,
+                       &ctx->get_cntrfd, NULL);
+    if (ret!=0) {
+        RAISE_ERROR_MSG("context cntr_open failed (%s)\n", fi_strerror(errno));
+        return ret;
+    }
+
+    ret = allocate_endpoints(ctx, info);
```
jdinan: I would pull the cntr_ep creation out of ```allocate_endpoints``` and do it here.  Then only use ```allocate_endpoints``` during initialization.

src/transport_ofi.c
```diff
@@ -1107,37 +1115,103 @@ int query_for_fabric(struct fabric_info *info)
...
+    if (ret!=0) {
+        RAISE_ERROR_MSG("ctx allocate endpoints failed (%s)\n", fi_strerror(errno));
+        return ret;
+    }
+
+    ret = bind_resources_to_and_enable_ep(ctx);
```
jdinan: Might split this function into one that does the CQ EP and a second one that you can use from here.

src/transport_none.h
```diff
@@ -88,6 +88,13 @@ shmem_transport_ctx_create(shmem_transport_ctx_t **ctx)
 }
```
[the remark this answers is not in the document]
davidozog: Done!

Issue #524, opened 2017-11-15
mr_reg heap failed with GNI provider


davidozog:
I get the following error with SOS and libfabric master on Cori@NERSC:

```
ozog@nid00523: ~/Repos/SOS/test/unit $ ./hello
[0000] WARN:  ../../../src/transport_ofi.c:546: allocate_recv_cntr_mr
[0000]        mr_reg heap failed
[0000] ERROR: ../../../src/init.c:238: shmem_internal_init
[0000]        Transport init failed (-22)
```

Here is my configure line:
```
../configure --disable-mr-scalable --with-xpmem=/opt/cray/xpmem/default --enable-completion-polling --enable-threads --with-pmi=/opt/cray/pe/pmi/default --prefix=<my_prefix> --with-ofi=<my_libfabric> --enable-debug CFLAGS=-O0
```

A similar error was reported by @yburette on Edison:
```
libfabric:gni:mr:__gnix_generic_register():404<info> [15403:1] could not allocate nic to do mr_reg, ret=-22
libfabric:gni:mr:__mr_cache_create_registration():1451<info> [15403:1] failed to register memory with callback
fi_mr_reg(): common/shared.c:395, ret=-12 (Cannot allocate memory)
```

davidozog:
I used `git bisect` (for the first time - it's a fantastic tool) to find the first libfabric commit where things went wrong for SOS.  It's here:
https://github.com/ofiwg/libfabric/commit/4a0e5c5ac327d3190262fb0dc15f62873ae328ce

davidozog:
On libfabric master, we're falling into this conditional:
https://github.com/ofiwg/libfabric/blob/master/prov/gni/src/gnix_mr.c#L266

jdinan:
@jswaro Might be able to help.  Looks like that commit added support for ```MR_SCALABLE```.  Have you tried using that (remove --disable-mr-scalable) instead of ```MR_BASIC```?

jswaro:
That conditional suggests that you are using basic registration (FI_MR_PROV_KEY in 1.5 terms), and you have a provided a non-zero key. In that specific case, the provider assumes this is an error on the part of the application and returns FI_EINVAL.

jdinan:
[...]

PR #525, opened 2017-11-20
OFI: fix memory registration for  FI_MR_BASIC

hppritcha:
It turns out the GNI provider's memory registration
implementation has gotten pickier about adhering
to the libfabric man pages.  It rejects memory
registration requests for domains using FI_MR_BASIC
which have a non-zero input key value.

This commit lets SOS again work with the GNI provider.

### DIFF
--- a/src/transport_ofi.c
+++ b/src/transport_ofi.c
@@ -536,11 +536,11 @@ int allocate_recv_cntr_mr(void)
 
 #else
     /* Register separate data and heap segments using keys 0 and 1,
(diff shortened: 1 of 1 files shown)


jdinan:
Unfortunately, the fix isn't this simple.  The same code is also used with MR_SCALABLE when remote virtual-addressing is disabled.  For that case, we need to supply the key values.

jdinan:
@hppritcha I think we've reached agreement that this will be fixed in the provider.  Would you like to close or revise this PR?

jdinan:
Fixed in the GNI provider (for details, see #524).

PR #526, opened 2017-11-28
Move CQ into the context

jdinan:
Eliminate sharing of CQs across contexts; this is needed to support the FI_THREAD_COMPLETION model supported by the PSM provider.  Additional changes:

 * Destroy the default context using common destroy routine
 * Add OFI return code checking macros to clean up/simplify return code checks
 * Remove global endpoint (ofi_epfd), replace usage of this with default context CNTR endpoint
 * Clean up / rewrite CQ error event reading/reporting code (usage should be reviewed)
 * TODO: Move bounce buffer pool to context to prevent deadlock from bounce buffer exhaustion and decrease mutex granularity
 * TODO: Add global target endpoint with FI_REMOTE_READ/WRITE and remove these flags from injection-only endpoints in contexts
 * TODO: Move STX into the context to eliminate STX (PSM context) sharing overheads


jdinan:
Ref: #527

jdinan:
Looks good!

Issue #527, opened 2017-11-28
OFI: Review OFI_CQ_ERROR Usage


jdinan:
Review code to check if we are looking at the CQ for error events when appropriate.

IIUC, we should only expect to find an error event on the CQ when the error code is ```-FI_EAVAIL```, but calls to ```OFI_CQ_ERROR``` are not limited to these cases.  Further, if an error event is not available on the CQ, the error is not being reported well (e.g. the error code is lost in the call to ```OFI_CQ_ERROR```).

Also, is there a race for threads to grab the error event?  If so, is it benign or can it be handled better?

PR #528, opened 2017-11-30
Add FI_THREAD_COMPLETION support

jdinan:
* Establish upper limit on bounce buffer pool size
  * Adds SHMEM_MAX_BOUNCE_BUFFERS env var
  * Cleans up RMA operation error handling and incorrect CQ usage
  * Clean up CQ drain loop
* Move STX into the context (future work, improve STX resource usage)
* Add support for FI_THREAD_COMPLETION via --enable-thread-completion (maybe not the best name, I'm open to suggestions)

### DIFF
--- a/src/transport_ofi.h
+++ b/src/transport_ofi.h
@@ -238,6 +238,7 @@ struct shmem_transport_ctx_t {
   long                            options;
   struct fid_ep*                  cntr_ep;
   struct fid_ep*                  cq_ep;
+  struct fid_stx*                 stx;
   struct fid_cntr*                put_cntr;
   struct fid_cntr*                get_cntr;
   struct fid_cq*                  cq;
--- a/src/transport_ofi.h
+++ b/src/transport_ofi.h
@@ -72,7 +72,7 @@ extern long                             shmem_transport_ofi_max_bounce_buffers;
             if (ret == 1) {                                                     \
                 const char *errmsg = fi_cq_strerror((ctx)->cq, e.prov_errno,    \
                                                     e.err_data, NULL, 0);       \
-                RAISE_ERROR_STR(errmsg);                                        \
+                RAISE_ERROR_MSG("Error in operation: %s\n", errmsg);            \
             } else {                                                            \
                 RAISE_ERROR_MSG("Error reading from CQ (%zd)\n", ret);          \
             }                                                                   \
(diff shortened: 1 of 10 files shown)


jdinan:
Ready for your review, @davidozog!

jdinan:
A few comments:
 * We could disable atomics with USE_CTX_LOCK to improve performance
 * Needs testing with the PSM provider (with --enable-thread-completion)

jdinan:
Closes #527

src/transport_ofi.h
```diff
@@ -389,27 +412,55 @@ int shmem_transport_fence(shmem_transport_ctx_t* ctx)
...
 
     if (ret) {
         if (ret == -FI_EAGAIN) {
-            shmem_transport_ofi_drain_cq(ctx);
+            if (ctx->cq_ep) {
+                shmem_free_list_lock(ctx->bounce_buffers);
+                shmem_transport_ofi_drain_cq(ctx);
+                shmem_free_list_unlock(ctx->bounce_buffers);
+            }
+            else if (0) {
```
davidozog: Is this intentional?  I'd think `else {...` would be ok.

src/transport_ofi.h
```diff
@@ -492,23 +543,22 @@ void shmem_transport_put_nb(shmem_transport_ctx_t* ctx, void *target, const void
 
         shmem_transport_put_small(ctx, target, source, len, pe);
 
-    } else if (len <= shmem_transport_ofi_bounce_buffer_size &&
-               ctx->options & SHMEMX_CTX_BOUNCE_BUFFER) {
+    } else if (len <= shmem_transport_ofi_bounce_buffer_size && ctx->cq_ep) {
```
davidozog: I see that checking for a non-null `cq_ep` is equivalent (and probably fewer cycles?) than checking the options, but it's ever so slightly less clear what's being checked.  No problem from me, just thought I'd mention it.

src/transport_ofi.h
```diff
@@ -279,38 +296,35 @@ void shmem_transport_ofi_drain_cq(shmem_transport_ctx_t *ctx)
...
                                                            const size_t len)
 {
     shmem_transport_ofi_bounce_buffer_t *buff;
 
-    buff = (shmem_transport_ofi_bounce_buffer_t*) shmem_free_list_alloc(pool);
+    shmem_free_list_lock(ctx->bounce_buffers);
+
+    while (ctx->bounce_buffers->nalloc > shmem_transport_ofi_max_bounce_buffers) {
+        shmem_transport_ofi_drain_cq(ctx);
+    }
```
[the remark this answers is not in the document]
jdinan: Hmm, well, one advantage would be that you might give up the lock sooner..

src/transport_ofi.h
jdinan: Oh, good catch!  That's left over from debugging.

src/transport_ofi.h
jdinan: Yeah, I went back and forth on whether to clear ```SHMEMX_CTX_BOUNCE``` from options or to leave it at what was requested by the user.  Ultimately, I cleared it, but I left this new check since it should eliminate an instruction.

PR #529, opened 2017-12-01
Add a separate target endpoint

davidozog:
Each SOS context had its transmit endpoint enable FI_REMOTE_READ and FI_REMOTE_WRITE, which probably consumes extraneous hardware resources.  This change adds a new endpoint, `shmem_transport_ofi_target_ep`, which is inserted into the address vector, and removes FI_REMOTE_READ/WRITE from the other transmitting endpoints.


davidozog:
@jdinan Should be ready for review, thanks!

src/transport_ofi.c
```diff
@@ -980,6 +981,38 @@ int query_for_fabric(struct fabric_info *info)
...
+    info->p_info->tx_attr->mode = 0;
+    info->p_info->rx_attr->mode = 0;
+
+    ret = fi_endpoint(shmem_transport_ofi_domainfd,
+                      info->p_info, &shmem_transport_ofi_target_ep, NULL);
+    OFI_CHECK_RETURN_MSG(ret, "target endpoint creation failed (%s)\n", fi_strerror(errno));
+
+    /* Attach the shared context */
+    ret = fi_ep_bind(shmem_transport_ofi_target_ep, &shmem_transport_ofi_stx->fid, 0);
+    OFI_CHECK_RETURN_STR(ret, "fi_ep_bind STX to target endpoint failed");
```
jdinan: The target EP shouldn't need an STX, since it's never used for transmission.

src/transport_ofi.c
```diff
@@ -980,6 +981,38 @@ int query_for_fabric(struct fabric_info *info)
     return ret;
 }
 
+static int shmem_transport_ofi_target_ep_init() {
+    int ret = 0;
+
+    struct fabric_info* info = &shmem_transport_ofi_info;
+    info->p_info->ep_attr->tx_ctx_cnt = FI_SHARED_CONTEXT;
```
jdinan: I think this should be 0 for the target EP.

src/transport_ofi.c
```diff
@@ -980,6 +981,38 @@ int query_for_fabric(struct fabric_info *info)
...
+    info->p_info->caps = FI_RMA | FI_ATOMICS;
```
jdinan: Do you need to also ```| FI_REMOTE_READ | FI_REMOTE_WRITE```?  Otherwise, I think we automatically get all of ```FI_REMOTE_READ/WRITE | FI_READ/WRITE```.

src/transport_ofi.c
```diff
@@ -980,6 +981,38 @@ int query_for_fabric(struct fabric_info *info)
...
+    ret = fi_endpoint(shmem_transport_ofi_domainfd,
+                      info->p_info, &shmem_transport_ofi_target_ep, NULL);
+    OFI_CHECK_RETURN_MSG(ret, "target endpoint creation failed (%s)\n", fi_strerror(errno));
+
+    /* Attach the shared context */
+    ret = fi_ep_bind(shmem_transport_ofi_target_ep, &shmem_transport_ofi_stx->fid, 0);
+    OFI_CHECK_RETURN_STR(ret, "fi_ep_bind STX to target endpoint failed");
+
+    /* Attach the address vector */
+    ret = fi_ep_bind(shmem_transport_ofi_target_ep, &shmem_transport_ofi_avfd->fid, 0);
```
jdinan: I think this is also unnecessary, since the EP is not used to transmit.

src/transport_ofi.c
```diff
@@ -1246,6 +1278,11 @@ int shmem_transport_fini(void)
 
     shmem_transport_ctx_destroy(&shmem_transport_ctx_default);
 
+    if (shmem_transport_ofi_target_ep &&
+        fi_close(&shmem_transport_ofi_target_ep->fid)) {
+        RAISE_ERROR_MSG("target endpoint close failed (%s)\n", fi_strerror(errno));
+    }
```
jdinan: This conditional should be unnecessary, since the target EP should always exist.

src/transport_ofi.c
```diff
@@ -1007,9 +1040,8 @@ static int shmem_transport_ofi_ctx_init(shmem_transport_ctx_t *ctx, int id)
 
     struct fabric_info* info = &shmem_transport_ofi_info;
     info->p_info->ep_attr->tx_ctx_cnt = FI_SHARED_CONTEXT;
-    info->p_info->caps = FI_RMA | FI_WRITE | FI_READ | FI_ATOMICS
-                         | FI_REMOTE_WRITE | FI_REMOTE_READ;
-    info->p_info->tx_attr->op_flags = FI_DELIVERY_COMPLETE | FI_INJECT_COMPLETE;
+    info->p_info->caps = FI_RMA | FI_WRITE | FI_READ | FI_ATOMICS;
+    info->p_info->tx_attr->op_flags = FI_DELIVERY_COMPLETE;
```
jdinan: Just re-read the descriptions for these flags, and I'm scratching my head.  I think I'd like to understand this field better.  I'm wondering if ```FI_INJECT_COMPLETE``` (which seems too relaxed to support shmem_quiet) combined with bounce buffering was giving us an ordering issue that we were only seeing in the Cray tests?

src/transport_ofi.c
davidozog: Whoops, yup.

src/transport_ofi.c
davidozog: Oh yeah, I forgot we don't want `FI_READ/WRITE`.

src/transport_ofi.c
davidozog: Yeah, I was just keeping with the style of the conditional checks below.  Should I remove those too?  If I'm not mistaken, all these objects should exist?

src/transport_ofi.c
davidozog: Hmm... could be.  You mean the cswap Cray tests?  I haven't seen the bug yet, we'll have more stats with my next push(es).

src/transport_ofi.c
jdinan: If they are for objects that should always exist, we should remove them.  Or potentially add an "else" case to act as an error check.

src/transport_ofi.c
jdinan: ```FI_DELIVERY_COMPLETE``` is correct and is what we're using on master.  Looks like this oddity might have come over from the prototype implementation.

davidozog: Oh right, I knew that.  So this can't be the cswap problem...

src/transport_ofi.c
davidozog: When I remove this, I get a "bad access" seg fault in the RX progress code:
``` 
  frame #0: 0x00000001002959fe libfabric.1.dylib`sock_av_get_addr_index(av=0x0000000000000000, addr=0x0000000101f00000) at sock_av.c:71
    frame #1: 0x00000001002b9528 libfabric.1.dylib`sock_pe_process_rx_conn_msg(pe=0x000000010035b000, rx_ctx=0x0000000100603920, pe_entry=0x00000001003841a0) at sock_progress.c:1474
    frame #2: 0x00000001002b69d1 libfabric.1.dylib`sock_pe_process_recv(pe=0x000000010035b000, rx_ctx=0x0000000100603920, pe_entry=0x00000001003841a0) at sock_progress.c:1542
    frame #3: 0x00000001002b30ee libfabric.1.dylib`sock_pe_progress_rx_pe_entry(pe=0x000000010035b000, pe_entry=0x00000001003841a0, rx_ctx=0x0000000100603920) at sock_progress.c:2035
    frame #4: 0x00000001002b32c9 libfabric.1.dylib`sock_pe_progress_rx_ctrl_ctx(pe=0x000000010035b000, rx_ctx=0x0000000100603920, tx_ctx=0x00000001006036f0) at sock_progress.c:2454
    frame #5: 0x00000001002b3458 libfabric.1.dylib`sock_pe_progress_tx_ctx(pe=0x000000010035b000, tx_ctx=0x00000001006036f0) at sock_progress.c:2487
    frame #6: 0x00000001002b4e05 libfabric.1.dylib`sock_pe_progress_thread(data=0x000000010035b000) at sock_progress.c:2648
```
Am I missing something, or do we indeed need to bind the AV?

jdinan: Looks like I'm wrong -- the fi_ep_bind manpage says "Connectionless endpoints must be bound to a single address vector."

It would be sorta nice if fi_ep_enable did a sanity check on the EP...

jdinan:
Good to merge once Travis gives it the green light!

PR #530, opened 2017-12-04
Manpages1.4

angelali4096:
Updated script to be compatible with version 1.4 of the specification. Changes include:
- Updated list of defined macros
- Supports bullet lists within manpages
- Creates manpage for every possible function (considering sizes and types)
- Tables no longer need to be in separate files

### DIFF
--- a/man/shmem_addr_accessible.3
+++ b/man/shmem_addr_accessible.3
@@ -1,80 +1,124 @@
-.TH SHMEM_ADDR_ACCESSIBLE 3 "Open Source Software Solutions, Inc.""OpenSHMEM Library Documentation"
+.TH SHMEM_ADDR_ACCESSIBLE 3 "Open Source Software Solutions, Inc." "OpenSHMEM Library Documentation"
 ./ sectionStart
 .SH NAME
 shmem_addr_accessible \- 
 Determines whether an address is accessible via OpenSHMEM data transfer
 routines from the specified remote PE.
+
 ./ sectionEnd
+
+
 ./ sectionStart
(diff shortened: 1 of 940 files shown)


lanl-ompi:

davidozog:
add to whitelist

davidozog:
@hppritcha Will that "add to whitelist" get rid of this message on PRs from @angelali4096?

jdinan:
Could you also create another PR to update the manpages in the SOS ```man/``` directory?

angelali4096:
@jdinan 
I didn't do this beforehand because it would track more than 500 more files that could be automatically generated by the script. Also, if the script changes, some if not all the manpages will change. Is this what we want?

jdinan:
Yeah, it's a lot of files, but we want to have them in the repository.  Otherwise, users would also need to have a copy of the spec to get a complete build.

jdinan:
Looks good to me -- @davidozog ?

jdinan:
IIUC, it requires a patched version of the OpenSHMEM 1.4-RC5.

davidozog:
Cool, makes sense.  Just wondering if that's in a branch/tag somewhere for future reference.

jdinan:
Works ok for me on the RC5 branch.

davidozog:
Is that `jdinan:drafts/osh-1.4-rc5`?   I see `Error: shpdeallc.tex is not in the given directory`.

jdinan:
Yeah, you also need this patch: https://github.com/jdinan/specification/pull/3

PR #531, opened 2017-12-04
OFI: Allow provider to select CQ size

jdinan:
Since we are setting FI_RM_ENABLED on the domain, it's preferred to
allow the provider to select the CQ size in order to allow better
resource management.

### DIFF
--- a/src/transport_ofi.c
+++ b/src/transport_ofi.c
@@ -996,9 +996,11 @@ static int shmem_transport_ofi_ctx_init(shmem_transport_ctx_t *ctx, int id)
         cntr_get_attr.wait_obj = FI_WAIT_UNSPEC;
     }
 
+    /* Allow provider to choose CQ size, since we are using FI_RM_ENABLED.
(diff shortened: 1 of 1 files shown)


src/transport_ofi.c
```diff
@@ -999,7 +998,7 @@ static int shmem_transport_ofi_ctx_init(shmem_transport_ctx_t *ctx, int id)
 
     struct fi_cq_attr cq_attr = {0};
     cq_attr.format = FI_CQ_FORMAT_CONTEXT;
-    cq_attr.size = shmem_transport_ofi_queue_slots;
+    cq_attr.size = 0; /* Size chosen by provider */
```
jdinan: @davidozog We can actually delete this line, since it's already set to 0 on line 999.  Should it go?

davidozog: Definitely, that would be a good choose.  :wink:

jdinan: Done.  :)

PR #534, opened 2017-12-07
Eliminate locking, when possible

jdinan:
Eliminate locks in FUNNELED and SERIALIZED threading modes and in MULTIPLE mode when the context is PRIAVATE or SERIALIZED.


jdinan:
Yeah, that's the problem with cross-cutting features like this.  We should definitely keep this gap in mind as we (or maybe @angelali4096) start writing tests for serialized and private contexts.

PR #536, opened 2017-12-07
Clean up OFI contexts array

jdinan:
Clean up the contexts array code and stop leaking array slots when contexts are freed.

### DIFF
--- a/src/transport_ofi.c
+++ b/src/transport_ofi.c
@@ -1073,7 +1073,7 @@ static int shmem_transport_ofi_ctx_init(shmem_transport_ctx_t *ctx, int id)
     ret = bind_enable_cntr_ep_resources(ctx);
     OFI_CHECK_RETURN_MSG(ret, "context bind/enable CNTR endpoint failed (%s)\n", fi_strerror(errno));
 
-    if (ctx->options | SHMEMX_CTX_BOUNCE_BUFFER &&
+    if (ctx->options & SHMEMX_CTX_BOUNCE_BUFFER &&
         shmem_transport_ofi_bounce_buffer_size > 0 &&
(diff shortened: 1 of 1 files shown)


src/transport_ofi.c
```diff
@@ -90,15 +90,13 @@ shmem_internal_mutex_t          shmem_transport_ofi_lock;
 
 struct fabric_info shmem_transport_ofi_info = {0};
 
-static shmem_transport_ctx_t** shmem_transport_ofi_contexts;
-static size_t shmem_transport_ofi_num_ctx = 0;
+static shmem_transport_ctx_t** shmem_transport_ofi_contexts = NULL;
+static size_t shmem_transport_ofi_contexts_size = 0;
```
davidozog: To me, `num_ctx` makes slightly more sense than `contexts_size`, but it's no big deal.   If it a namespace sorta thing, maybe `contexts_num`?

jdinan: "contexts" is the array name and "contexts_size" is the size of the array.  We no longer track the number of context references stored in the array (since we don't use it anywhere, it was unnecessary global state).  So, "contexts_size" seemed clearer than "contexts_num", but I'm happy to rename either or both if you have a better name in mind.

davidozog: Yeah, "contexts_size" is good.  I just tend to first think of "sizeof" and "size_t", and tend towards "num" and "len" for number of elements, but that's personal preference.  No problem either way, just a thought.

This PR looks great, BTW, it's much cleaner.  Ready to merge when you are.

jdinan: Oh, "len" is good.

Issue #452, opened 2017-09-28
OpenSHMEM 1.4 - change log #9 - ordering semantics


davidozog:
Assure the ordering semantics of memory ordering, point-to-point synchronization and collective synchronization.

jdinan:
Check for appropriate memory barriers in quiet, fence, wait, barrier, sync, etc APIs.

jdinan:
Possible to have an optimized build when threading and shmem_ptr (i.e. XPMEM) APIs are disabled.

jdinan:
In ```shmem_atomic.h```, add new routine ```shmem_internal_membar``` that calls ```__sync_synchronize```.  Sprikle calls in all SHMEM synchronization routines.

jdinan:
Related issue: #506

Issue #200, opened 2016-06-07
XPMEM put/get __sync_synchornize


kseager:
## Issue

extra synchronization around a memcpy should not be needed; quiet should instead be the point of using the memory fence to ensure the change is visible to other cores

ref:
http://pgas11.rice.edu/papers/BrightwellPedretti-OpenSHMEM-PGAS11.pdf
## Proposed solution

1) add ifdef syncs with TOTAL_DATA_ORDERING, including NO-OP fence
2) verify that shmem_wait calls have XPMEM syncs

jdinan:
Related to #452

jdinan:
Can't remove memory barriers until internal wait/wait_until calls contain appropriate memory barriers.

PR #537, opened 2017-12-08
Internal memory barrier for ordering semantics

wrrobin:
@jdinan Please review the changes and let me know if it looks ok. Also, this PR does not include any memory barrier for wait of remote operation completion, since that seems to be out-of-scope for issue #452 and #200 to me.


lanl-ompi:

PR #539, opened 2017-12-08
Pr/ordering semantics

wrrobin:


lanl-ompi:

src/shmem_atomic.h
```diff
@@ -74,6 +74,46 @@ shmem_spinlock_fini(shmem_spinlock_t *lock)
...
 }
 
+
+static inline
+void
+shmem_internal_membar(void) {
+#if (defined(__STDC_NO_ATOMICS__) || !defined(HAVE_STDATOMIC_H))
+    __sync_synchronize();
+#else
+#include <stdatomic.h>
```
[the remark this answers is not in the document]
jdinan: Thinking about this a bit more -- stdatomic can contain function definitions and nested functions are invalid in C.  Best to move the include outside of the function body.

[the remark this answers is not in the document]
wrrobin: Yes. I had a similar thought and wanted to have your opinion on this. So, if I put it outside and use gcc, is there any chance that a "missing header" error might happen?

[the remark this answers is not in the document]
wrrobin: Ok. Will make the changes then.

jdinan:
Should we replace the xpmem_transport_quiet/fence with direct calls to membar?  Would this change simplify the code or make it more complicated?

jdinan:
Also, need to ensure that membar is defined in the case where threading is disabled and XPMEM is enabled.

src/shmem_atomic.h
```diff
@@ -25,6 +25,8 @@
 #endif
 
 #define COMPILER_FENCE() do { __asm__ __volatile__ ("" ::: "memory"); } while (0)
+#define LFENCE()  __asm__ __volatile__ ("lfence" ::: "memory")
```
jdinan: Need ```do { ... } while (0)``` around function-like macros to safely eat the trailing semicolon: https://gcc.gnu.org/onlinedocs/cpp/Swallowing-the-Semicolon.html

jdinan: If you structure it this way, the ```LFENCE``` and ```SFENCE``` macros should also be defined within an ```#if defined(__i386__) || defined(__x86_64__)``` block with an else case that defines them to ```__sync_synchronize```.  I think I would prefer to just put this directly into the membar_load/store routines. 
 Since lfence/sfence are only used once, there's limited value to defining macros for them.

wrrobin: Thanks @jdinan. So, similar to this? In this case, we should not need #if defined(__i386__) || defined(__x86_64__).
shmem_internal_membar_load(void) {
    do { 
          __asm__ __volatile__ ("lfence" ::: "memory"); 
     } while (0);
    return;
 }

jdinan: The ```do { ... } while (0)``` business is for creating function-like macros, so not needed if you stick the asm into the membar_load function.  I think the logic might be simpler if you do:
```
shmem_internal_membar_load(void) {
#if defined(__i386__) || defined(__x86_64__)
    __asm__ __volatile__ ("lfence" ::: "memory"); 
#else
    __sync_synchronize();
#endif
    return;
 }
```

wrrobin: Ok. Got it. Thanks.

src/transport_xpmem.h
```diff
@@ -80,7 +80,7 @@ static inline
 int
 shmem_transport_xpmem_quiet(void)
 {
-    __sync_synchronize();
+    shmem_internal_membar();
```
jdinan: This looks like it will cause a circular include, since ```shmem_synchronization.h``` includes ```transport_xpmem.h``` (via ```shmem_comm.h``` -- our headers are such a mess).  You are missing the ```#include "shmem_synchronization.h"``` needed to define this function, but as soon as you include it, the compiler will barf (assuming you test with XPMEM enabled).  Better to leave this as ```__sync_synchrionize()```.

wrrobin: ```shmem_internal_membar()``` is defined in ```shmem_atomic.h```. I am not sure why ```shmem_synchronization.h``` would be needed here? Am I missing anything here?

jdinan: Ok, looks like I got my wires crossed.  I think you do need to ```#include "shmem_atomic.h"``` in ```transport_xpmem.h```.

wrrobin: Ok. I will add it then. But, why dont I see the compiler complains about this in my local or in travis?

jdinan: It's probably already included in the file that includes transport_xpmem.h.

wrrobin:
Thanks @jdinan for the feedbacks. Good to know that this went through.

jdinan:
@wrrobin Could you remind me of the rationale for the membars in put/get_wait?  I'm trying to recall the sequence of SHMEM operations that necessitates this and am coming up blank.

jdinan:
Looks like we are also missing membars in the shmem_wait routines?  Maybe this is a good opportunity to add ```shmem_internal_wait```?

jdinan:
The quiet in ```shmem_ctx_destroy``` is also missing a membar.  Might be easiest to move the quiet up to the ```shmem_ctx_destroy``` level.

Barriers look ok: ```shmem_internal_barrier``` calls ```shmem_internal_quiet``` and all internal barriers call the ```shmem_internal_barrier``` routine.

wrrobin:
@jdinan, Regarding the put/get_wait, the membars are added because of the blocking shmem_puts defined in data_c.c. After each non-blocking put, there is a call to the put/get_wait to make sure of its completion. It adds an extra membar in the case of collectives as there the quiet/fence is also invoking membar. Let me know if there is any better way to handle this.

wrrobin:
Regarding shmem_wait, are you referring to the macros? Yes, we can add shmem_internal_wait in this case.

wrrobin:
Missed the ```shmem_ctx_destroy```; Will fix it in the next round.

wrrobin:
@jdinan, regarding the ```shmem_wait``` macros, the membar is needed before going to the wait, right? Trying to see whether I am understanding it correctly.

wrrobin:
@jdinan, this PR is ready for review now.

src/shmem_atomic.h
```diff
@@ -74,6 +74,76 @@ shmem_spinlock_fini(shmem_spinlock_t *lock)
     shmem_internal_assertp(lock->enter == lock->exit);
 }
 
+
+#if (defined(__STDC_NO_ATOMICS__) || !defined(HAVE_STDATOMIC_H))
+
+static inline
+void
+shmem_internal_membar(void) {
+#if USE_XPMEM || shmem_internal_thread_level != SHMEM_THREAD_SINGLE
```
jdinan: You can't combine a preprocessor conditional and a runtime check like this.  The second part will always evaluate to false at compile time since the preprocessor doesn't have a symbol named shmem_internal_thread_level.

jdinan: I think you can get what you're after by defining a top-level macros like this:
```
#if defined(USE_XPMEM)
#define SHMEM_INTERNAL_NEED_MEMBAR 1
#elif defined (ENABLE_THREADS)
#define SHMEM_INTERNAL_NEED_MEMBAR (shmem_internal_thread_level != SHMEM_THREAD_SINGLE)
#else
#define SHMEM_INTERNAL_NEED_MEMBAR 0
#endif
```

Then in each membar function:
```
  if (SHMEM_INTERNAL_NEED_MEMBAR) __sync_synchronize();
```

src/shmem_lock.h
```diff
@@ -93,6 +94,8 @@ shmem_internal_set_lock(long *lockp)
 
             SHMEM_WAIT(&(lock->data), lock_cur.data);
         }
+    } else {
+        shmem_internal_membar_load();
```
jdinan: I will surely forget the reason for this memory barrier.  Could you add a comment to the effect: Lock was acquired immediately without calling SHMEM_WAIT, which provides memory ordering.  Therefore, issue a load fence to ensure memory ordering.

src/shmem_synchronization.h
```diff
@@ -32,6 +32,8 @@ shmem_internal_quiet(shmem_ctx_t ctx)
 #ifdef USE_XPMEM
     ret = shmem_transport_xpmem_quiet();
     if (0 != ret) { RAISE_ERROR(ret); }
+#else
+    shmem_internal_membar();
 #endif
```
jdinan: I think we can delete the shmem_transport_xpmem_quiet/fence routines and just call shmem_internal_membar, right?  The latter routine already contains the checks for XPMEM.

src/shmem_synchronization.h
```diff
@@ -47,6 +49,8 @@ shmem_internal_fence(shmem_ctx_t ctx)
 #ifdef USE_XPMEM
     ret = shmem_transport_xpmem_fence();
     if (0 != ret) { RAISE_ERROR(ret); }
+#else
+    shmem_internal_membar_store();
 #endif
```
jdinan: Same comment as quiet.  The shmem_transport_xpmem_fence routine is now redundant.

src/shmem_atomic.h
wrrobin: Thanks Jim. To use ```SHMEM_THREAD_SINGLE``` in the macro, we have to include shmem.h header as follows, right?
```
#define SHMEM_INTERNAL_INCLUDE
#include "shmem.h"
```

wrrobin:
@jdinan Thanks for the feedback. I have incorporated the changes. Please review it and let me know if they look ok.

src/shmem_atomic.h
```diff
@@ -29,6 +29,19 @@
 #define likely(x)       __builtin_expect((x),1)
 #define unlikely(x)     __builtin_expect((x),0)
 
+#define SHMEM_INTERNAL_INCLUDE
+#include "shmem.h"
```
jdinan: Why was this added?  We should only need to include ```shmem.h``` in the .c files that define the C language bindings.

src/shmem_synchronization.h
```diff
@@ -46,12 +41,7 @@ shmem_internal_fence(shmem_ctx_t ctx)
     ret = shmem_transport_fence((shmem_transport_ctx_t *)ctx);
     if (0 != ret) { RAISE_ERROR(ret); }
 
-#ifdef USE_XPMEM
-    ret = shmem_transport_xpmem_fence();
```
jdinan: Please also delete these functions from ```transport_xpmem.h```.

jdinan: That is, please delete xpmem_fence/quiet since they are now unused.

src/transport_xpmem.h
```diff
@@ -18,6 +18,7 @@
 
 #include <string.h>
 #include <xpmem.h>
+#include "shmem_atomic.h"
```
jdinan: Once you delete xpmem_quiet/fence, you can also delete this change.

src/shmem_atomic.h
wrrobin: If we do not include ```shmem.h```, then the build reports this error. 
```
./shmem_atomic.h:40:69: error: 'SHMEM_THREAD_SINGLE' undeclared (first use in this function)
 # define SHMEM_INTERNAL_NEED_MEMBAR (shmem_internal_thread_level != SHMEM_THREAD_SINGLE)
```
Please let me know how else I can resolve this.

jdinan: Oh, right!  This #include is ok to keep.

wrrobin:
@jdinan, one of the tests failed, but I am not sure what is the error in the log. https://api.travis-ci.org/v3/job/319308315/log.txt

davidozog:
It's the blasted cswap bug!  #378  So far it's still a mystery.

It's caused us quite a bit of grief, but so far it only manifests in these Cray tests.  It's a nondeterministic bug, so if you re-run you might get lucky.  :)

jdinan:
Yeah, fing_cswap strikes again.  I'm going to ignore it and merge.

wrrobin:
Ok. Sounds good.

Issue #523, opened 2017-11-15
Contexts unit tests failing in Travis w/ OFI master


davidozog:
With SOS 1.4.0-dev and the libfabric master branch, the "many-ctx" and "web" unit tests fail with the following stack trace (and at least 2 processes):
```
#0  0x00007ffff72551ec in sock_pe_progress_tx_entry (pe=0x7ffff7faf010, tx_ctx=0x625c90, pe_entry=0x7ffff7fce520) at ../prov/sockets/src/sock_progress.c:1926
#1  0x00007ffff7257115 in sock_pe_progress_tx_ctx (pe=0x7ffff7faf010, tx_ctx=0x625c90) at ../prov/sockets/src/sock_progress.c:2472
#2  0x00007ffff7257b1c in sock_pe_progress_thread (data=0x7ffff7faf010) at ../prov/sockets/src/sock_progress.c:2648
#3  0x00007ffff6ddf6ba in start_thread (arg=0x7ffff6bd3700) at pthread_create.c:333
#4  0x00007ffff75af3dd in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:109
```
Note that this does *not* occur on libfabric ~v1.5.1 and earlier, and it has so far only manifested in Linux with the sockets provider.  I think the problem arises from somewhere around [this commit in libfabric](https://github.com/ofiwg/libfabric/commit/74571f240a6b989ac71808ab148fdae362d297f8).

davidozog:
libfabric issue [#3557](https://github.com/ofiwg/libfabric/issues/3557)

davidozog:
Appears to be fixed in libfabric PR [#3584](https://github.com/ofiwg/libfabric/pull/3584)

jdinan:
Hooray!  Let's leave this ticket open as a reminder to move Travis back to libfabric master once the above PR is merged.

PR #540, opened 2017-12-08
[WIP] More efficient STX sharing/control

davidozog:
This is a WIP that allocates an array of STXs, with the goal of sharing them based on the thread model and OpenSHMEM context options.  It introduces parameters for the maximum array size, as well as the heuristic/algorithm for sharing STXs.  Infrastructure is in place for further optimization (based on tid and a private/shared context), and I would appreciate any feedback you can offer before going further.  Thanks!


src/transport_ofi.c
```diff
@@ -271,6 +272,31 @@ typedef enum{
...
 uint64_t shmem_transport_ofi_max_poll = (1ULL<<30);
 
+
+/* STX control variables */
+enum stx_share_alg_t {
+    ROUNDROBIN = 0,
+    RANDOM
+};
+typedef enum stx_share_alg_t stx_share_alg_t;
+stx_share_alg_t shmem_transport_stx_share_alg;
```
jdinan: static?

src/transport_ofi.c
```diff
@@ -271,6 +272,31 @@ typedef enum{
...
+
+long shmem_transport_ofi_stx_max;
```
jdinan: static?

src/transport_ofi.c
```diff
@@ -271,6 +272,31 @@ typedef enum{
...
+    long              ref_cnt;
+    bool              private;   /* true if private, false if shared */
+#ifdef __APPLE__
+    uint64_t          owner_tid;
+#else
+    pid_t             owner_tid;
+#endif
+};
+typedef struct shmem_transport_stx_t shmem_transport_stx_t;
+shmem_transport_stx_t* shmem_transport_ofi_stx_pool;
```
jdinan: static?

src/transport_ofi.c
jdinan: Naming: we use ```shmem_transport_*``` for things that are part of the generic transport API (i.e. provided by both Portals and OFI).  For OFI-only things, we should name them ```shmem_transport_ofi_*```.

src/transport_ofi.c
```diff
@@ -312,14 +338,17 @@ int bind_enable_cq_ep_resources(shmem_transport_ctx_t *ctx)
...
 
 static inline
-int bind_enable_cntr_ep_resources(shmem_transport_ctx_t *ctx)
+int bind_enable_cntr_ep_resources(shmem_transport_ctx_t *ctx, uint32_t stx_idx)
 {
     int ret = 0;
 
     /* Attach the shared context */
-    ret = fi_ep_bind(ctx->cntr_ep, &ctx->stx->fid, 0);
+    ret = fi_ep_bind(ctx->cntr_ep, &shmem_transport_ofi_stx_pool[stx_idx].stx->fid, 0);
```
jdinan: Can you get ```stx_idx``` from the ```ctx``` argument instead of passing through the function args?

src/transport_ofi.c
```diff
@@ -1057,23 +1086,41 @@ static int shmem_transport_ofi_ctx_init(shmem_transport_ctx_t *ctx, int id)
...
+            case (RANDOM):
+                stx_idx = rand() % shmem_transport_ofi_stx_max;
+                break;
+        }
+    }
+
+#ifdef __APPLE__
+    pthread_threadid_np(NULL, &shmem_transport_ofi_stx_pool[stx_idx].owner_tid);
+#else
+    shmem_transport_ofi_stx_pool[stx_idx].owner_tid = getid();
```
jdinan: gettid()

src/transport_ofi.c
```diff
@@ -1137,6 +1185,35 @@ int shmem_transport_init(void)
...
+    shmem_transport_ofi_stx_max = shmem_internal_params.OFI_STX_MAX;
+
+    /* STX sharing settings */
+    char *type = shmem_internal_params.OFI_STX_SHARE_ALGORITHM;
+    if (0 == strcmp(type, "round-robin")) {
+        shmem_transport_stx_share_alg = ROUNDROBIN;
+    } else if (0 == strcmp(type, "random")) {
+        shmem_transport_stx_share_alg = RANDOM;
+    } else {
+        RAISE_WARN_MSG("Ignoring bad STX share algorithm '%s'\n", type);
```
jdinan: What is ```shmem_transport_stx_share_alg``` set to in the else case?

src/transport_ofi.c
```diff
@@ -1137,6 +1185,35 @@ int shmem_transport_init(void)
...
+    }
+
+    if (shmem_internal_thread_level == SHMEM_THREAD_SINGLE &&
+        shmem_transport_ofi_stx_max > 1) {
+        /* We need only 1 STX per PE with SHMEM_THREAD_SINGLE */
+        RAISE_WARN_MSG("Ignoring invalid STX max setting '%ld' w/THREAD_SINGLE, using 1\n", 
+                       shmem_transport_ofi_stx_max);
+        shmem_transport_ofi_stx_max = 1;
```
jdinan: If the default is more than 1, I think we will get this warning with every THREAD_SINGLE run.  You can look at ```shmem_internal_params.OFI_STX_SHARE_ALGORITHM_provided``` to see if the user provided a value versus if the variable was set to the default value.

I think we also want to land in this case for FUNNELED.  Maybe SERIALIZED, although, I'm not sure..

src/transport_ofi.h
```diff
@@ -242,20 +242,21 @@ struct shmem_transport_ctx_t {
...
   struct fid_ep*                  cntr_ep;
   struct fid_ep*                  cq_ep;
-  struct fid_stx*                 stx;
   struct fid_cntr*                put_cntr;
   struct fid_cntr*                get_cntr;
   struct fid_cq*                  cq;
   shmem_internal_atomic_uint64_t  pending_put_cntr;
   shmem_internal_atomic_uint64_t  pending_get_cntr;
   shmem_free_list_t              *bounce_buffers;
+  int                             stx_idx;
```
jdinan: int here and uint32_t elsewhere?

jdinan:
Closes #523

src/transport_ofi.c
```diff
@@ -86,6 +87,16 @@ static char                     myephostname[EPHOSTNAMELEN];
...
 #ifdef ENABLE_THREADS
 shmem_internal_mutex_t          shmem_transport_ofi_lock;
+/* Need a syscall to gettid() because glibc doesn't provide a wrapper
+ * (see gettid manpage in the NOTES section): */
+    #ifndef __APPLE__
+        #include <sys/syscall.h>
+        #ifdef SYS_gettid
+        #define gettid() syscall(SYS_gettid)
+        #else
+        #warning "SYS_gettid not available"
```
jdinan: This should probably be ```#error``` since the build will fail anyway when ```gettid()``` is undefined.  Also, ```#error``` is standard C, but ```#warning``` is an extension.

src/transport_ofi.c
```diff
@@ -86,6 +87,16 @@ static char                     myephostname[EPHOSTNAMELEN];
 #endif
 #ifdef ENABLE_THREADS
 shmem_internal_mutex_t          shmem_transport_ofi_lock;
+/* Need a syscall to gettid() because glibc doesn't provide a wrapper
+ * (see gettid manpage in the NOTES section): */
```
jdinan: Should we add a configury check to see if ```gettid()``` already exists?  Otherwise, this will cause potential errors since ```sys/types.h``` is included above, which pulls in ```gettid()``` but it's redefined below.

davidozog: Would it help to _not_ redefine `gettid()`, and name this something like `shmem_transport_ofi_gettid()`?

Otherwise, yeah, a configury check would be fine.

jdinan: You can take a look at how this is handled for CMA.  It's pretty much the same situation.

jdinan:
I broke this PR.  Reopened as #551

PR #550, opened 2017-12-20
Cleanup Travis warnings

jdinan:

PR #551, opened 2017-12-20
More efficient STX control/sharing

jdinan:
This is a WIP that allocates an array of STXs, with the goal of sharing them based on the thread model and OpenSHMEM context options. It introduces parameters for the maximum array size, as well as the heuristic/algorithm for sharing STXs. Infrastructure is in place for further optimization (based on tid and a private/shared context), and I would appreciate any feedback you can offer before going further. Thanks!

### DIFF
--- a/README
+++ b/README
@@ -208,7 +208,7 @@ options.
         polling (i.e. there is no polling limit).  The default behavior is to
         call fi_cntr_wait without polling.
 
-    SHMEM_OFI_STX_MAX (default: 128)
+    SHMEM_OFI_STX_MAX (default: 16)
         Sets the maximum number of sharable transmit contexts (STXs) per PE.
         After reaching the maximum, PEs share the available STXs using the
         SHMEM_OFI_STX_SHARE_HEURISTIC parameter.
--- a/README
+++ b/README
@@ -211,7 +211,8 @@ options.
     SHMEM_OFI_STX_MAX (default: 16)
         Sets the maximum number of sharable transmit contexts (STXs) per PE.
         STXs are the underlying transmit resources that are allocated to
-        OpenSHMEM contexts, according to the SHMEM_OFI_STX_ALLOCATOR parameter.
+        OpenSHMEM contexts and they are allocated using the algorithm specified
+        by the SHMEM_OFI_STX_ALLOCATOR parameter.
 
     SHMEM_OFI_STX_ALLOCATOR (default: round-robin)
         Algorithm for allocating STX resources to OpenSHMEM contexts.  In
--- a/README
+++ b/README
@@ -210,13 +210,13 @@ options.
 
     SHMEM_OFI_STX_MAX (default: 16)
         Sets the maximum number of sharable transmit contexts (STXs) per PE.
-        After reaching the maximum, PEs share the available STXs using the
-        SHMEM_OFI_STX_SHARE_ALGORITHM parameter.
+        STXs are the underlying transmit resources that are allocated to
+        OpenSHMEM contexts, according to the SHMEM_OFI_STX_ALLOCATOR parameter.
 
-    SHMEM_OFI_STX_SHARE_ALGORITHM (default: round-robin)
-        Algorithm to share the available STX contexts after reaching the value
-        of SHMEM_OFI_STX_SHARE_MAX.  The default option is round-robin, which
-        uses STXs in increasing circular order.  Options are: round-robin,
+    SHMEM_OFI_STX_ALLOCATOR (default: round-robin)
+        Algorithm for allocating STX resources to OpenSHMEM contexts.  In
+        particular, the algorithm determines how resources are shared by
+        contexts once all STXs have been allocated.  Options are: round-robin,
         random.
 
   Debugging Environment variables:
--- a/README
+++ b/README
@@ -208,6 +208,17 @@ options.
         polling (i.e. there is no polling limit).  The default behavior is to
         call fi_cntr_wait without polling.
 
+    SHMEM_OFI_STX_MAX (default: 256)
+        Sets the maximum number of sharable transmit contexts (STXs) per PE.
+        After reaching the maximum, PEs share the available STXs using the
+        SHMEM_OFI_STX_SHARE_HEURISTIC parameter.
+
+    SHMEM_OFI_STX_SHARE_ALGORITHM (default: round-robin)
+        Algorithm to share the available STX contexts after reaching the value
+        of SHMEM_OFI_STX_SHARE_MAX.  The default option is round-robin, which
+        uses STXs in increasing circular order.  Options are: round-robin,
+        random.
+
   Debugging Environment variables:
 
     SHMEM_DEBUG (default: off)
--- a/README
+++ b/README
@@ -211,7 +211,7 @@ options.
     SHMEM_OFI_STX_MAX (default: 16)
         Sets the maximum number of sharable transmit contexts (STXs) per PE.
         After reaching the maximum, PEs share the available STXs using the
-        SHMEM_OFI_STX_SHARE_HEURISTIC parameter.
+        SHMEM_OFI_STX_SHARE_ALGORITHM parameter.
 
     SHMEM_OFI_STX_SHARE_ALGORITHM (default: round-robin)
         Algorithm to share the available STX contexts after reaching the value
--- a/README
+++ b/README
@@ -208,7 +208,7 @@ options.
         polling (i.e. there is no polling limit).  The default behavior is to
         call fi_cntr_wait without polling.
 
-    SHMEM_OFI_STX_MAX (default: 256)
+    SHMEM_OFI_STX_MAX (default: 128)
         Sets the maximum number of sharable transmit contexts (STXs) per PE.
         After reaching the maximum, PEs share the available STXs using the
         SHMEM_OFI_STX_SHARE_HEURISTIC parameter.
--- a/config/check_cma.m4
+++ b/config/check_cma.m4
@@ -14,7 +14,7 @@ AC_DEFUN([CHECK_CMA], [
              AC_LANG_PUSH([C])
              AC_COMPILE_IFELSE([
                 AC_LANG_SOURCE([[#include <sys/syscall.h>]],
-                               [[int cma=__NR_process_vm_readv;]])],
+                               [[long cma=__NR_process_vm_readv;]])],
                 [cma_happy="yes"],
                 [cma_happy="no"])
              AC_LANG_POP([C])
--- a/config/check_gettid.m4
+++ b/config/check_gettid.m4
@@ -1,0 +1,15 @@
+#CHECK_GETTID([action-if-found], [action-if-not-found])
+# --------------------------------------------------------
+# check if SYS_gettid is available.
+AC_DEFUN([CHECK_GETTID], [
+    AC_MSG_CHECKING([gettid() syscall definitions])
+    AC_LANG_PUSH([C])
+    AC_COMPILE_IFELSE([
+       AC_LANG_SOURCE([[#include <sys/syscall.h>]],
+                      [[int tid = SYS_gettid;]])],
+       [gettid_happy="yes"],
+       [gettid_happy="no"])
+    AC_LANG_POP([C])
+    AC_MSG_RESULT([$gettid_happy])
+    AS_IF([test "$gettid_happy" = "yes"], [$1], [$2])
+])
--- a/config/check_gettid.m4
+++ b/config/check_gettid.m4
@@ -6,7 +6,7 @@ AC_DEFUN([CHECK_GETTID], [
     AC_LANG_PUSH([C])
     AC_COMPILE_IFELSE([
        AC_LANG_SOURCE([[#include <sys/syscall.h>]],
-                      [[int tid = SYS_gettid;]])],
+                      [[long tid = SYS_gettid;]])],
        [gettid_happy="yes"],
        [gettid_happy="no"])
     AC_LANG_POP([C])
--- a/config/check_gettid.m4
+++ b/config/check_gettid.m4
@@ -2,10 +2,11 @@
 # --------------------------------------------------------
 # check if SYS_gettid is available.
 AC_DEFUN([CHECK_GETTID], [
-    AC_MSG_CHECKING([gettid() syscall definitions])
+    AC_MSG_CHECKING([SYS_gettid syscall definitions])
     AC_LANG_PUSH([C])
     AC_COMPILE_IFELSE([
        AC_LANG_SOURCE([[#include <sys/syscall.h>]],
+                      [[#define _GNU_SOURCE]],
                       [[long tid = SYS_gettid;]])],
        [gettid_happy="yes"],
        [gettid_happy="no"])
--- a/configure.ac
+++ b/configure.ac
@@ -108,8 +108,7 @@ AS_IF([test "$enable_threads" != "no"], [
        AC_DEFINE([ENABLE_THREADS], [1], [Enable threads])
        AC_CHECK_HEADERS([stdatomic.h])
        CHECK_GETTID(
-           [AC_DEFINE([_GNU_SOURCE], [1], [Use of syscall from libc requires _GNU_SOURCE feature macro])
-            AC_DEFINE([HAVE_SYS_GETTID], [1], [Have support for SYS_gettid syscall])],
+           [AC_DEFINE([HAVE_SYS_GETTID], [1], [Have support for SYS_gettid syscall])],
            [AC_MSG_WARN([No support for SYS_gettid syscall])])
        AS_IF([test "$enable_thread_completion" = "yes"],
              [AC_DEFINE([USE_CTX_LOCK], [1], [Enable per-context locks])
--- a/configure.ac
+++ b/configure.ac
@@ -108,7 +108,8 @@ AS_IF([test "$enable_threads" != "no"], [
        AC_DEFINE([ENABLE_THREADS], [1], [Enable threads])
        AC_CHECK_HEADERS([stdatomic.h])
        CHECK_GETTID(
-           [AC_DEFINE([HAVE_SYS_GETTID], [1], [Have support for SYS_gettid syscall])],
+           [AC_DEFINE([_GNU_SOURCE], [1], [Use of syscall from libc requires _GNU_SOURCE feature macro])
+            AC_DEFINE([HAVE_SYS_GETTID], [1], [Have support for SYS_gettid syscall])],
            [AC_MSG_WARN([No support for SYS_gettid syscall])])
        AS_IF([test "$enable_thread_completion" = "yes"],
              [AC_DEFINE([USE_CTX_LOCK], [1], [Enable per-context locks])
--- a/configure.ac
+++ b/configure.ac
@@ -107,6 +107,9 @@ AS_IF([test "$enable_threads" != "no"], [
        fi
        AC_DEFINE([ENABLE_THREADS], [1], [Enable threads])
        AC_CHECK_HEADERS([stdatomic.h])
+       CHECK_GETTID(
+           [AC_DEFINE([HAVE_SYS_GETTID], [1], ["Have SYS_gettid"])],
+           [AC_MSG_WARN([Do not have SYS_gettid])])
        ])
 AM_CONDITIONAL([HAVE_PTHREADS], [test "$HAVE_POSIX_THREADS" = "1"])
 
--- a/src/transport_ofi.c
+++ b/src/transport_ofi.c
@@ -35,6 +35,7 @@
 #include "transport_ofi.h"
 #include <unistd.h>
 #include "runtime.h"
+#include "uthash.h"
 
 struct fabric_info {
     struct fi_info *fabrics;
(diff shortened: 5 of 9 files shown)


src/transport_ofi.c
```diff
@@ -1297,6 +1445,12 @@ int shmem_transport_fini(void)
     if (shmem_transport_ofi_contexts) free(shmem_transport_ofi_contexts);
     shmem_transport_ctx_destroy(&shmem_transport_ctx_default);
 
+    for(i = 0; i < shmem_transport_ofi_stx_max; ++i) {
+        ret = fi_close(&shmem_transport_ofi_stx_pool[i].stx->fid);
+        OFI_CHECK_ERROR_MSG(ret, "STX context close failed (%s)\n", fi_strerror(errno));
+    }
+    free(shmem_transport_ofi_stx_pool);
```
jdinan: You should check that the hash table is empty and, if not, issue an error.  The user is required to free all private contexts before calling finalize.

src/transport_ofi.c
```diff
@@ -1057,19 +1137,41 @@ static int shmem_transport_ofi_ctx_init(shmem_transport_ctx_t *ctx, int id)
...
+    if (shmem_internal_thread_level > SHMEM_THREAD_FUNNELED) {
+        stx_idx_prev = stx_idx;
+        switch (shmem_transport_ofi_stx_share_alg) {
+            case (ROUNDROBIN):
+                /* TODO: After reaching the limit, could search for elements with ref_cnt==0. */
+                stx_idx = (stx_idx + 1) % shmem_transport_ofi_stx_max;
+                break;
+            case (RANDOM):
+                /* TODO: Assure fair sharing w/ a periodic PRNG. Could also search for ref_cnt==0. */
+                stx_idx = rand() % shmem_transport_ofi_stx_max;
```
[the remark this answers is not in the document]
jdinan: Also, suggest moving the code to allocate an STX from the pool into a function, where you can tell the function whether you want a shared or private STX.

src/transport_ofi.c
```diff
@@ -87,7 +88,20 @@ static char                     myephostname[EPHOSTNAMELEN];
 #endif
 #ifdef ENABLE_THREADS
 shmem_internal_mutex_t          shmem_transport_ofi_lock;
-#endif
+/* Need a syscall to gettid() because glibc doesn't provide a wrapper
+ * (see gettid manpage in the NOTES section): */
+#ifndef __APPLE__
+#ifdef HAVE_SYS_GETTID
+static inline
+TID_TYPE shmem_transport_ofi_gettid(void)
```
jdinan: Is TID_TYPE correct here?  I think the Linux syscall always returns the same type.

Also, why not have a case here that defines ```shmem_transport_ofi_gettid()``` for __APPLE__?  This should simplify things further down and also reduce the total number of ifdefs in this file.

src/transport_ofi.c
```diff
@@ -315,9 +358,46 @@ int bind_enable_cntr_ep_resources(shmem_transport_ctx_t *ctx)
...
 
-    /* Attach the shared context */
-    ret = fi_ep_bind(ctx->cntr_ep, &ctx->stx->fid, 0);
-    OFI_CHECK_RETURN_STR(ret, "fi_ep_bind STX to CNTR endpoint failed");
+    /* STX contexts are shared by SHMEM contexts that are private to the same
+     * thread (i.e. have SHMEM_CTX_PRIVATE option set and same thread ID).  */
+    if (ctx->options & SHMEM_CTX_PRIVATE) {
+        shmem_transport_ofi_stx_kvs_t *e = malloc(sizeof(shmem_transport_ofi_stx_kvs_t));
+        e->tid = ctx->tid;
+        e->stxs = malloc(sizeof(shmem_transport_ofi_stx_t));
```
jdinan: I think you want e->stxs to be a pointer rather than a struct, so you can capture a reference to the STX rather than a deep copy of the struct (you aren't going to succeed at deep copying the OFI stx object, so better not to try).

src/transport_ofi.c
```diff
@@ -315,9 +358,46 @@ int bind_enable_cntr_ep_resources(shmem_transport_ctx_t *ctx)
...
+
+        shmem_transport_ofi_stx_kvs_t *f;
+        HASH_FIND_INT(shmem_transport_ofi_stx_kvs, &ctx->tid, f);
```
jdinan: I think it would make more sense the check the hash table for an entry for the given TID before constructing a new entry.  If there is an existing entry, you can just increment its ref_cnt rather than replacing it with a new entry.

configure.ac
```diff
@@ -107,7 +107,9 @@ AS_IF([test "$enable_threads" != "no"], [
        fi
        AC_DEFINE([ENABLE_THREADS], [1], [Enable threads])
        AC_CHECK_HEADERS([stdatomic.h])
-
+       CHECK_GETTID(
+           [AC_DEFINE([HAVE_SYS_GETTID], [1], [Have support for SYS_gettid syscall])],
+           [AC_MSG_WARN([No support for SYS_gettid syscall])])
```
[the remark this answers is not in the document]
davidozog: We're currently not using `SYS_gettid` on Mac, and fortunately this warning isn't printed on OSX, because it's available (at least on Sierra/HighSierra) .  _Unfortunately_, I see that it's a deprecated symbol at compile time.  Perhaps I could try to clean this up using something like `AC_CANONICAL_HOST` and `$host_os`?

README
```diff
@@ -208,6 +208,17 @@ options.
         polling (i.e. there is no polling limit).  The default behavior is to
         call fi_cntr_wait without polling.
 
+    SHMEM_OFI_STX_MAX (default: 128)
```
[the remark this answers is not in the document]
davidozog: Created issue #552

src/transport_ofi.c
davidozog: I'm not 100% sure on this one.  It seems that nowadays syscall usually returns `long` (but sometimes `int`).  I think it's ok for the return type to be `pid_t`, which is nowadays often 64-bit (sometimes 32, sometimes signed, sometimes unsigned), but I'm ok setting this to `long`, or other ideas.

src/transport_ofi.c
```diff
@@ -87,7 +88,20 @@ static char                     myephostname[EPHOSTNAMELEN];
...
+ * (see gettid manpage in the NOTES section): */
+#ifndef __APPLE__
+#ifdef HAVE_SYS_GETTID
+static inline
+TID_TYPE shmem_transport_ofi_gettid(void)
+{
+    return syscall(SYS_gettid);
+}
+#else
+#error "SYS_gettid not available"
```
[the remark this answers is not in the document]
davidozog: It seems to work fine (both Mac and Linux), but I'm kinda surprised you don't need an atomic increment.  Maybe our multi-threaded tests are hammering it hard enough, or maybe the compiler takes care of it.  I threw it in as above for now.

src/transport_ofi.c
```diff
@@ -270,6 +284,35 @@ typedef enum{
...
+
+struct shmem_transport_ofi_stx_t {
+    struct fid_stx*   stx;
+    long              ref_cnt;
+};
+typedef struct shmem_transport_ofi_stx_t shmem_transport_ofi_stx_t;
+static shmem_transport_ofi_stx_t* shmem_transport_ofi_stx_pool;
+
+struct shmem_transport_ofi_stx_kvs_t {
+    struct shmem_transport_ofi_stx_t* stxs;
```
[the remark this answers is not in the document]
davidozog: Just my sanity, since it's a different struct than `fid_stx`.  So 's' is "struct", I guess.  I removed it.

src/transport_ofi.c
davidozog: +1

src/transport_ofi.c
davidozog: 👍

src/transport_ofi.c
```diff
@@ -315,9 +358,46 @@ int bind_enable_cntr_ep_resources(shmem_transport_ctx_t *ctx)
...
+            e->stxs->ref_cnt = f->stxs->ref_cnt + 1;
+            stx_idx = stx_idx_prev;
+            e->stx_idx = f->stx_idx;
+            ctx->stx_idx = f->stx_idx;
+            e->stxs->stx = f->stxs->stx;
+
+            ret = fi_ep_bind(ctx->cntr_ep, &e->stxs->stx->fid, 0);
+            OFI_CHECK_RETURN_STR(ret, "fi_ep_bind existing STX private to CNTR endpoint failed");
+
+            HASH_REPLACE_INT(shmem_transport_ofi_stx_kvs, tid, e, f);
```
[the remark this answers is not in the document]
davidozog: 👍

src/transport_ofi.c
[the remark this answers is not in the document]
davidozog: Yeah, good catch.  I think I fixed this in the latest push.

src/transport_ofi.c
```diff
@@ -1132,6 +1235,41 @@ int shmem_transport_init(void)
...
+        shmem_transport_ofi_stx_share_alg = RANDOM;
+    } else {
+        RAISE_WARN_MSG("Ignoring bad STX share algorithm '%s', using 'round-robin'\n", type);
+        shmem_transport_ofi_stx_share_alg = ROUNDROBIN;
+    }
+
+    if ((shmem_internal_thread_level == SHMEM_THREAD_SINGLE ||
+         shmem_internal_thread_level == SHMEM_THREAD_FUNNELED ) &&
+         shmem_internal_params.OFI_STX_MAX_provided &&
+         shmem_internal_params.OFI_STX_MAX > 1) {
```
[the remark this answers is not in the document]
davidozog: 👍

src/transport_ofi.c
davidozog: 👍

src/transport_ofi.h
```diff
@@ -29,6 +29,13 @@
 #include "shmem_free_list.h"
 #include "shmem_internal.h"
 #include "shmem_atomic.h"
+#include <sys/types.h>
+
+#ifdef HAVE_SYS_GETTID
+#define _GNU_SOURCE
```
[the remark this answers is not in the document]
davidozog: I went with exactly the approach in #550 for the headers.  There is also an additional `AC_DEFINE([_GNU_SOURCE], ...)` in `configure.ac`, which appears to work out.

src/transport_ofi.c
```diff
@@ -87,7 +95,39 @@ static char                     myephostname[EPHOSTNAMELEN];
...
+static inline
+pid_t shmem_transport_ofi_gettid(void)
+{
+    return syscall(SYS_gettid);
+}
+#else
+static inline
+uint64_t shmem_transport_ofi_gettid(void)
+{
+    static uint64_t tid = 0;
```
jdinan: Add a comment here to the effect that we can't query TID, so we assume each query corresponds to a unique thread.

src/transport_ofi.c
```diff
@@ -87,7 +95,39 @@ static char                     myephostname[EPHOSTNAMELEN];
...
+uint64_t shmem_transport_ofi_gettid(void)
+{
+    int ret;
+    uint64_t tid;
+    ret = pthread_threadid_np(NULL, &tid);
+    if (ret == 0) {
+      return tid;
+    } else {
+        RAISE_ERROR_MSG("Error geting thread ID: %s \n", strerror(ret));
+        return -1;
```
jdinan: Do you want to return -1 or abort?  ```RAISE_ERROR_*``` will abort (and annotated with noreturn), so the return should be unreachable.  Also, you can't return -1 in a uint.

src/transport_ofi.c
```diff
@@ -271,6 +311,181 @@ typedef enum{
...
+typedef enum stx_share_alg_t stx_share_alg_t;
+static stx_share_alg_t shmem_transport_ofi_stx_share_alg;
+static int stx_start_idx = 0;
+
+static long shmem_transport_ofi_stx_max;
+
+struct shmem_transport_ofi_stx_t {
+    struct fid_stx*   stx;
+    long              ref_cnt;
+    bool              is_private;
```
jdinan: I believe you need to ```#include <stdbool.h>``` for the ```bool``` type and ```true``` and ```false``` macros.

src/transport_ofi.c
```diff
@@ -271,6 +311,181 @@ typedef enum{
...
+ * unselected subset of the STX pool.  After depleting the pool, we start over
+ * again on the (now random) array of pool indices, and so on.*/
+static inline
+void shmem_transport_ofi_stx_rand_init() {
+    rand_pool_indices = malloc(shmem_transport_ofi_stx_max*sizeof(uint32_t));
+    if (rand_pool_indices == NULL) {
+        RAISE_ERROR_STR("Error: out of memory when allocating OFI random STX indices");
+    }
+
+    for(int i=0; i<shmem_transport_ofi_stx_max; i++) {
```
jdinan: I think we require C99 support to build SOS now, right?

src/transport_ofi.c
```diff
@@ -271,6 +311,181 @@ typedef enum{
...
+    return;
+}
+
+static inline
+int fair_rand() {
+    /* Choose an STX index from the unselected subset */
+    int choice = rand() % rand_pool_top_idx;
+
+    /* Swap the value at the chosen index with the value at the top index */
+    int tmp = rand_pool_indices[choice];
```
jdinan: Should tmp be uint32_t?

src/transport_ofi.c
```diff
@@ -271,6 +311,181 @@ typedef enum{
...
+                   !shmem_transport_ofi_stx_pool[i].is_private ) {
+                    ctx->stx_idx = i;
+                    break;
+                } else {
+                    i = fair_rand();
+                    count++;
+                }
+            }
+            if (count == shmem_transport_ofi_stx_max) {
+                /* No free STX is available, so share with shmem_ctx_default */
```
jdinan: Once all of the STXs have been allocated, we could share with any STX that is shared.  You could store these in a second array and use the same algorithm (RR, Random, etc.) to select from the array of shared STXs.

But, let's see if we can merge this and then layer on the above afterward.

src/transport_ofi.c
```diff
@@ -1133,6 +1359,47 @@ int shmem_transport_init(void)
     ret = allocate_fabric_resources(&shmem_transport_ofi_info);
     if (ret != 0) return ret;
 
+    /* STX max settings */
+    if ((shmem_internal_thread_level == SHMEM_THREAD_SINGLE ||
+         shmem_internal_thread_level == SHMEM_THREAD_FUNNELED ) &&
+         shmem_internal_params.OFI_STX_MAX > 1) {
+        if (shmem_internal_params.OFI_STX_MAX_provided) {
+            /* We need only 1 STX per PE with SHMEM_THREAD_SINGLE or SHMEM_THREAD_FUNNELED */
+            RAISE_WARN_MSG("Ignoring invalid STX max setting '%ld' w/THREAD_SINGLE, using 1\n",
```
jdinan: This applies to both single and funneled.  Suggest updating the message to: "Ignoring STX max setting '%ld'; using 1 STX in single-threaded mode".

src/transport_ofi.c
```diff
@@ -1252,8 +1519,23 @@ void shmem_transport_ctx_destroy(shmem_transport_ctx_t *ctx)
...
-    OFI_CHECK_ERROR_MSG(ret, "STX context close failed (%s)\n", fi_strerror(errno));
+    if (ctx->options & SHMEM_CTX_PRIVATE) {
+        shmem_transport_ofi_stx_kvs_t *e;
+        HASH_FIND_INT(shmem_transport_ofi_stx_kvs, &ctx->tid, e);
+        if (e) {
+            e->stx->ref_cnt--;
+            if (e->stx->ref_cnt == 0) {
+                HASH_DEL(shmem_transport_ofi_stx_kvs, e);
+                free(e);
+                shmem_transport_ofi_stx_pool[ctx->stx_idx].is_private = false;
```
jdinan: Will this get re-used after it's freed?

src/transport_ofi.c
```diff
@@ -271,6 +311,181 @@ typedef enum{
...
+        shmem_transport_ofi_stx_kvs_t *f;
+        HASH_FIND_INT(shmem_transport_ofi_stx_kvs, &ctx->tid, f);
+        if (f) {
+            /* A thread-private STX already exists for this thread id, so increment the
+             * reference count and rollback to the last index into the STX resource pool:  */
+            f->stx->ref_cnt++;
+            ctx->stx_idx = f->stx_idx;
+
+        } else {
+            shmem_transport_ofi_stx_pool_search(ctx);
```
jdinan: What if the STX we get back is shared?  It looks like we'll go ahead and register it to this thread, which would be bad.

src/transport_ofi.c
```diff
@@ -271,6 +311,181 @@ typedef enum{
...
+            e->stx->ref_cnt++;
+            e->stx->is_private = true;
+
+            HASH_ADD_INT(shmem_transport_ofi_stx_kvs, tid, e);
+        }
+
+
+    } else {
+        shmem_transport_ofi_stx_pool_search(ctx);
+        shmem_transport_ofi_stx_pool[ctx->stx_idx].is_private = false;
```
jdinan: This looks like the a risky place to reset is_private.

src/transport_ofi.c
davidozog: Yeah, I just want to abort.

src/transport_ofi.c
```diff
@@ -271,6 +311,181 @@ typedef enum{
...
 uint64_t shmem_transport_ofi_max_poll = (1ULL<<30);
 
+
+/* STX control variables */
+enum stx_share_alg_t {
+    ROUNDROBIN = 0,
+    RANDOM
+};
+typedef enum stx_share_alg_t stx_share_alg_t;
+static stx_share_alg_t shmem_transport_ofi_stx_share_alg;
```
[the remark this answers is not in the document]
davidozog: 👍

src/transport_ofi.c
```diff
@@ -271,6 +311,181 @@ typedef enum{
...
+static int stx_start_idx = 0;
```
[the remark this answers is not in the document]
davidozog: 👍

src/transport_ofi.c
davidozog: 🌮

src/transport_ofi.c
```diff
@@ -271,6 +311,181 @@ typedef enum{
...
+                }
+            }
+            if (count == shmem_transport_ofi_stx_max) {
+                /* No free STX is available, so share with shmem_ctx_default */
+                ctx->stx_idx = 0;
+                if (ctx->options & SHMEM_CTX_PRIVATE) {
+                    RAISE_WARN_STR("No OFI STX available, so sharing with the default context.");
+                }
+            }
+            break;
```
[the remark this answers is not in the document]
davidozog: 👍

src/transport_ofi.c
```diff
@@ -271,6 +311,181 @@ typedef enum{
...
+static inline
+void shmem_transport_ofi_stx_pool_search(shmem_transport_ctx_t *ctx)
+{
+    int i, count = 0;
+    switch (shmem_transport_ofi_stx_share_alg) {
+        case (ROUNDROBIN):
+            i = stx_start_idx;
+            while (count < shmem_transport_ofi_stx_max) {
+                if( shmem_transport_ofi_stx_pool[i].ref_cnt == 0 &&
+                   !shmem_transport_ofi_stx_pool[i].is_private ) {
```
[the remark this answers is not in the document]
davidozog: I think they are in the latest version, right?

src/transport_ofi.c
davidozog: I think this is addressed in the latest round of commits.

src/transport_ofi.c
davidozog: I think this is addressed in the latest round of commits.

README
```diff
@@ -208,6 +208,18 @@ options.
...
+        Sets the maximum number of sharable transmit contexts (STXs) per PE.
+        STXs are the underlying transmit resources that are allocated to
+        OpenSHMEM contexts and they are allocated using the algorithm specified
+        by the SHMEM_OFI_STX_ALLOCATOR parameter.
+
+    SHMEM_OFI_STX_ALLOCATOR (default: round-robin)
+        Algorithm for allocating STX resources to OpenSHMEM contexts.  In
+        particular, the algorithm determines how resources are shared by
+        contexts once all STXs have been allocated.  Options are: round-robin,
+        random.
```
jdinan: Add ```SHMEM_OFI_STX_THRESHOLD```.

config/check_gettid.m4
```diff
@@ -0,0 +1,16 @@
+#CHECK_GETTID([action-if-found], [action-if-not-found])
+# --------------------------------------------------------
+# check if SYS_gettid is available.
+AC_DEFUN([CHECK_GETTID], [
+    AC_MSG_CHECKING([SYS_gettid syscall definitions])
+    AC_LANG_PUSH([C])
+    AC_COMPILE_IFELSE([
+       AC_LANG_SOURCE([[#include <sys/syscall.h>]],
+                      [[#define _GNU_SOURCE]],
+                      [[long tid = SYS_gettid;]])],
```
jdinan: I think AC_LANG_SOURCE only takes one argument?  I suspect the other arguments are getting ignored.

It looks like ```_GNU_SOURCE``` needs to move up so it's defined before including ```sys/syscall.h```.  You could put an ```#ifndef __APPLE__``` in there to force this test to always pass on Mac.

config/check_cma.m4
```diff
@@ -15,7 +15,7 @@ AC_DEFUN([CHECK_CMA], [
              AC_LANG_PUSH([C])
              AC_COMPILE_IFELSE([
                 AC_LANG_SOURCE([[#include <sys/syscall.h>]],
-                               [[int cma=__NR_process_vm_readv;]])],
+                               [[long cma=__NR_process_vm_readv;]])],
```
jdinan: See AC_LANG_SOURCE comment below.

jdinan: Also, does this need _GNU_SOURCE?

config/check_gettid.m4
jdinan: I verified that the second and third arguments are getting ignored, so this should be updated to:
```
        AC_LANG_SOURCE([[
#ifndef __APPLE__
#define _GNU_SOURCE
#include <sys/syscall.h>
long tid = SYS_gettid;
#endif
]])
```

jdinan:
@davidozog Please review the changes I added in 02c5dd5 to resolve the merge conflict with master.

jdinan:
Ready to merge.  👍

davidozog:
@jdinan, did you mean commit https://github.com/Sandia-OpenSHMEM/SOS/pull/551/commits/3cf6537db5e349275356df783df67085329fc8e3, https://github.com/Sandia-OpenSHMEM/SOS/commit/02c5dd559b673d8e8b7f5cabc8ffa247776c3461, or both?   Regardless, I'm actually just wondering if there's anything new in addition to the `stx_idx = -1` initialization and the check that it's `>=0` in `ctx_destroy`, which looks good.

jdinan:
Those are the only new changes.

PR #560, opened 2017-12-22
Make data_start/end weak symbols

jdinan:
Fix LD_PRELOAD error encountered by tools folks.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -596,19 +596,24 @@ AC_TRY_RUN([
 #else
 #pragma weak __data_start
 #pragma weak _end
-extern char __data_start;
-extern char _end;
+extern int __data_start;
+extern int _end;
 #endif
+
+int data;
 
 int main(void) {
     void *base;
-    unsigned long length;
+    long length;
 #ifdef __APPLE__
     base = (void*) get_etext();
     length = get_end() - get_etext();
 #else
-    if (&__data_start == 0 || &_end == 0) return 1;
+    if (&__data_start == (int*) 0 || &_end == (int*) 0) return 1;
+    base = (void*) &__data_start;
+    length = (long) ((char*) &_end - (char*) &__data_start);
 #endif
+    if (!(base != (void*)0 && length > 0)) return 2;
     return 0;
 }],
     AC_MSG_RESULT([found]),
--- a/configure.ac
+++ b/configure.ac
@@ -594,8 +594,10 @@ AC_TRY_RUN([
 #ifdef __APPLE__
 #include <mach-o/getsect.h>
 #else
-extern char data_start;
-extern char end;
+#pragma weak __data_start
+#pragma weak _end
+extern char __data_start;
+extern char _end;
 #endif
 
 int main(void) {
@@ -605,8 +607,7 @@ int main(void) {
     base = (void*) get_etext();
     length = get_end() - get_etext();
 #else
-    base = &data_start;
-    length = (unsigned long) &end  - (unsigned long) &data_start;
+    if (&__data_start == 0 || &_end == 0) return 1;
 #endif
     return 0;
 }],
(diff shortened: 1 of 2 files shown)


src/init.c
```diff
@@ -209,8 +213,12 @@ shmem_internal_init(int tl_requested, int *tl_provided)
     shmem_internal_data_base = (void*) get_etext();
     shmem_internal_data_length = get_end() - get_etext();
 #else
-    shmem_internal_data_base = &data_start;
-    shmem_internal_data_length = (unsigned long) &end  - (unsigned long) &data_start;
+    if (&__data_start == 0 || &_end == 0)
```
jdinan: Should probably cast 0 to (int*).

Also, need a comment to record that if the weak symbol isn't bound, its address will be 0.

jdinan:
@davidozog Ready for review

configure.ac
```diff
@@ -594,20 +594,26 @@ AC_TRY_RUN([
...
     base = (void*) get_etext();
     length = get_end() - get_etext();
 #else
-    base = &data_start;
-    length = (unsigned long) &end  - (unsigned long) &data_start;
+    if (&__data_start == (int*) 0 || &_end == (int*) 0) return 1;
+    base = (void*) &__data_start;
+    length = (long) ((char*) &_end - (char*) &__data_start);
 #endif
+    if (!(base != (void*)0 && length > 0)) return 2;
```
davidozog: The `length>0` seems like a good check.  Do we also want it in `src/init.c`?  Or is it checked further along?

configure.ac
```diff
@@ -594,20 +594,26 @@ AC_TRY_RUN([
...
+    long length;
 #ifdef __APPLE__
     base = (void*) get_etext();
     length = get_end() - get_etext();
 #else
-    base = &data_start;
-    length = (unsigned long) &end  - (unsigned long) &data_start;
+    if (&__data_start == (int*) 0 || &_end == (int*) 0) return 1;
+    base = (void*) &__data_start;
+    length = (long) ((char*) &_end - (char*) &__data_start);
```
davidozog: The `(char*)` cast is slightly perplexing to me, but it looks like it works out...  Do we need it?  E.g., would `(long)&end - (long)&__data_start` not work?

src/init.c
```diff
@@ -209,8 +213,26 @@ shmem_internal_init(int tl_requested, int *tl_provided)
...
+     * assign them an lvalue of 0.  Here, we check that the binary that
+     * initializes OpenSHMEM has exposed these symbols, enabling the library to
+     * locate its symmetric data segment. */
+
+    if (&__data_start == (int*) 0 || &_end == (int*) 0)
+        RETURN_ERROR_MSG("Unable to locate symmetric data segment (%p, %p)\n",
+                         (void*) &__data_start, (void*) &_end);
+
+    shmem_internal_data_base = (void*) &__data_start;
+    shmem_internal_data_length = (long) ((char*) &_end - (char*) &__data_start);
```
davidozog: Same question as above regarding the `(char*)` cast.

configure.ac
jdinan: Casting a pointer to an integer is bad (unless it's intptr_t); it's better to do the arithmetic on the pointer type.  If we do the arithmetic on int*, we need to multiply the result by sizeof(int).  Casting to char* tells us directly how many bytes, which is the number we need.

configure.ac
jdinan: A zero-length data segment is technically allowed (although I'm not sure we've ever tested that this works).  The configury test puts something in the data segment, so we expect the size to be nonzero.

configure.ac
davidozog: Ah yes, agreed.

PR #569, opened 2018-01-08
Destroy context upon failure

davidozog:
This adds the return value to `shmem_ctx_create`, and calls `shmem_transport_ctx_destroy` if a non-zero return value is encountered during the OFI initializations.

The `many-ctx` unit test attempts to initialize as many contexts as possible.  When it does, it destroys some contexts, freeing enough resources to recover for some tests of the atomics (this is needed for the sockets provider, at least).

Note: the `many-ctx` test seems to work fine in Linux, but I've just noticed a seg. fault in OSX that I'm investigating now...

### DIFF
--- a/src/transport_ofi.c
+++ b/src/transport_ofi.c
@@ -1044,8 +1044,14 @@ static int shmem_transport_ofi_ctx_init(shmem_transport_ctx_t *ctx, int id)
 #ifdef USE_CTX_LOCK
     SHMEM_MUTEX_INIT(ctx->lock);
 #endif
-    shmem_internal_atomic_write(&ctx->pending_put_cntr, 0);
-    shmem_internal_atomic_write(&ctx->pending_get_cntr, 0);
+    ctx->cntr_ep = NULL;
+    ctx->cq_ep = NULL;
+    ctx->stx = NULL;
+    ctx->put_cntr = NULL;
+    ctx->get_cntr = NULL;
+    ctx->cq = NULL;
+    ctx->pending_put_cntr =  0;
+    ctx->pending_get_cntr = 0;
 
     ret = fi_cntr_open(shmem_transport_ofi_domainfd, &cntr_put_attr,
                        &ctx->put_cntr, NULL);
--- a/src/transport_ofi.c
+++ b/src/transport_ofi.c
@@ -1044,14 +1044,6 @@ static int shmem_transport_ofi_ctx_init(shmem_transport_ctx_t *ctx, int id)
 #ifdef USE_CTX_LOCK
     SHMEM_MUTEX_INIT(ctx->lock);
 #endif
-    ctx->cntr_ep = NULL;
-    ctx->cq_ep = NULL;
-    ctx->stx = NULL;
-    ctx->put_cntr = NULL;
-    ctx->get_cntr = NULL;
-    ctx->cq = NULL;
-    ctx->pending_put_cntr =  0;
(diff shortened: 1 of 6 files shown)


src/transport_ofi.c
```diff
@@ -1252,17 +1251,25 @@ void shmem_transport_ctx_destroy(shmem_transport_ctx_t *ctx)
...
-    ret = fi_close(&ctx->get_cntr->fid);
-    OFI_CHECK_ERROR_MSG(ret, "Context get CNTR close failed (%s)\n", fi_strerror(errno));
+    if (ctx->get_cntr) {
+        ret = fi_close(&ctx->get_cntr->fid);
+        OFI_CHECK_ERROR_MSG(ret, "Context get CNTR close failed (%s)\n", fi_strerror(errno));
+    }
 
-    ret = fi_close(&ctx->cq->fid);
-    OFI_CHECK_ERROR_MSG(ret, "Context CQ close failed (%s)\n", fi_strerror(errno));
+    if (ctx->cq) {
```
davidozog: On OSX (sockets provider), this pointer is non-null after the `FI_EMFILE` error above, and trying to call `fi_close` on it results in a bad access seg. fault.  Seems like this approach might not work in general...

davidozog: I think @kseager called it!  We need to initialize the `shmem_transport_ctx_t` field pointers to NULL for this to work correctly.  My latest commit resolves the OSX issue.

src/transport_ofi.c
```diff
@@ -1044,8 +1044,14 @@ static int shmem_transport_ofi_ctx_init(shmem_transport_ctx_t *ctx, int id)
...
-    shmem_internal_atomic_write(&ctx->pending_put_cntr, 0);
-    shmem_internal_atomic_write(&ctx->pending_get_cntr, 0);
+    ctx->cntr_ep = NULL;
+    ctx->cq_ep = NULL;
+    ctx->stx = NULL;
+    ctx->put_cntr = NULL;
+    ctx->get_cntr = NULL;
+    ctx->cq = NULL;
+    ctx->pending_put_cntr =  0;
+    ctx->pending_get_cntr = 0;
```
jdinan: Might need to use the atomic_write routines to keep the C11 atomics model happy.

jdinan: Also, you could use ```memset(ctx, 0, sizeof(shmem_transport_ctx_t))``` here.  If we add any fields in the future, that would ensure they also get zeroed out.

src/transport_ofi.c
```diff
@@ -1056,6 +1062,9 @@ static int shmem_transport_ofi_ctx_init(shmem_transport_ctx_t *ctx, int id)
     OFI_CHECK_RETURN_MSG(ret, "get_cntr creation failed (%s)\n", fi_strerror(errno));
 
     ret = fi_cq_open(shmem_transport_ofi_domainfd, &cq_attr, &ctx->cq, NULL);
+    if (ret && errno == FI_EMFILE) {
+        DEBUG_STR("Consider increasing the open file limit with 'ulimit' command");
```
jdinan: Need more context in the debug message, e.g.:
```Context creation failed because of open files limit, consider increasing with 'ulimit' command```.

jdinan: We may want to consider updating the "CHECK_RETURN_MSG" output since these are no longer considered to be errors.  Not sure I have a great suggestion, just an observation.

PR #575, opened 2018-01-19
Fix several communication/tracking completion races in Portals

jdinan:
Fix several communication tracking/completion races in ```SHMEM_THREAD_MULTIPLE``` mode in the Portals transport.


jdinan:
@davidozog Note that the Portals code gets built, but not tested by Travis (since the Portals sockets build doesn't work correctly in the Travis environment).

davidozog:
Looks like there might be a problem with the `shmem_internal_atomics` inclusions.

jdinan:
Argh... Travis!!  ⚡️ ✊ 🌩  (shakes fist at sky)

Issue #579, opened 2018-01-24
Private contexts and the contexts array


jdinan:
## Issue

Private contexts are assigned an ID and entered into the contexts array.  They are subsequently freed during shutdown, without warning.

## Proposed fix

Generate a warning when freeing a private context (spec requires the user to do this).

PR #582, opened 2018-01-25
Updates to OFI and Portals code

jdinan:


src/transport_ofi.c
```diff
@@ -1624,6 +1624,8 @@ int shmem_transport_fini(void)
 
     for (i = 0; i < shmem_transport_ofi_contexts_len; ++i) {
         if (shmem_transport_ofi_contexts[i]) {
+            if (shmem_transport_ofi_contexts[i]->options & SHMEM_CTX_PRIVATE)
+                RAISE_WARN_MSG("Shutting down with unfreed private context (%zd)\n", i);
```
jdinan: @davidozog Note that we are entering private contexts into the contexts array, so this loop will free them.  As a result the check below on line 1646 should really only trip if something went wrong with STX management.  I added this warning message to ensure the user gets a warning about unfreed private contexts.  Not sure if you want to make any broader changes with regard to how private contexts are handled.  At a minimum, maybe we want to update the warning message below?

jdinan:
Wonder if we can get @bwbarrett to take a look at 8b88758 and make sure I didn't miss something?

bwbarrett:
@jdinan Wow, you've caused all kinds of flashbacks with that code.

I'm 99% sure your patch is correct.  The implementations of Portals4 I had at the time would always be ordered for small messages, so we never would have caught the mistake in testing.  And there was just me writing code, so no one was going to catch that in a code review...

src/transport_ofi.c
davidozog: Yup, this looks good to me.  And I agree that 1646 should never happen unless something strange occurs.  Should we just remove it?  I'm not sure what we would print other than something like "unexpected context(s) was not freed".

davidozog:
Can you please review the warning message changes I made, @jdinan?  Thanks!

src/transport_ofi.c
jdinan: Looks like past Jim also filed an issue for this: #579

jdinan:
Thanks, @bwbarrett!

Issue #576, opened 2018-01-19
Portals: Per-context CT


jdinan:
Add support for per-context event counters in the Portals transport.

Possibly also add support for per-context bounce buffering and separate EQs per context.  It might be sufficient (or at least a sufficient first step) to just make bounce buffering a per-context option and use a single common bounce buffer pool among all contexts.

jdinan:
Closed by #587

PR #587, opened 2018-01-26
Portals: Update contexts implementation

jdinan:
* Add support for per-context CTs
 * Make bounce buffering and blocking put isolation per-context options

### DIFF
--- a/src/transport_portals4.c
+++ b/src/transport_portals4.c
@@ -731,7 +731,7 @@ shmem_transport_startup(void)
     }
 
     ret = shmem_transport_ctx_init((shmem_transport_ctx_t*)SHMEM_CTX_DEFAULT,
-                                   SHMEMX_CTX_BOUNCE_BUFFER,
+                                   SHMEMX_CTX_BOUNCE_BUFFER | SHMEMX_CTX_ISOLATE_PUTS,
                                    SHMEM_TRANSPORT_CTX_DEFAULT_ID);
 
  cleanup:
@@ -754,7 +754,7 @@ shmem_transport_fini(void)
     for (i = 0; i < shmem_transport_portals4_contexts_len; ++i) {
         if (shmem_transport_portals4_contexts[i]) {
             if (shmem_transport_portals4_contexts[i]->options & SHMEM_CTX_PRIVATE)
-                RAISE_WARN_MSG("Shutting down with unfreed private context (%zd)\n", i);
+                RAISE_WARN_MSG("Shutting down with unfreed private context (%d)\n", i);
             shmem_transport_quiet(shmem_transport_portals4_contexts[i]);
             shmem_transport_ctx_destroy(shmem_transport_portals4_contexts[i]);
         }
--- a/src/transport_portals4.c
+++ b/src/transport_portals4.c
@@ -711,7 +711,7 @@ shmem_transport_startup(void)
     ret = PtlCTAlloc(shmem_transport_portals4_ni_h,
                      &shmem_transport_portals4_put_event_ct_h);
     if (PTL_OK != ret) {
-        RAISE_WARN_MSG("Put CT allocation failed: %d\n", ret);
+        RAISE_WARN_MSG("Put event CT allocation failed: %d\n", ret);
         goto cleanup;
     }
 
@@ -726,7 +726,7 @@ shmem_transport_startup(void)
     ret = PtlMDBind(shmem_transport_portals4_ni_h, &md,
                     &shmem_transport_portals4_put_event_md_h);
     if (PTL_OK != ret) {
-        RETURN_ERROR_MSG("PtlMDBind of put MD failed: %d\n", ret);
+        RETURN_ERROR_MSG("PtlMDBind of put event MD failed: %d\n", ret);
         goto cleanup;
     }
 
@@ -748,8 +748,8 @@ shmem_transport_fini(void)
     /* synchronize the atomic cache, if there is one */
     PtlAtomicSync();
 
-    /* Free all shareable contexts.  This performs a quiet on each context,
-     * ensuring all operations have completed before proceeding with shutdown. */
+    /* Free all contexts.  This performs a quiet on each context, ensuring all
+     * operations have completed before proceeding with shutdown. */
 
     for (i = 0; i < shmem_transport_portals4_contexts_len; ++i) {
         if (shmem_transport_portals4_contexts[i]) {
--- a/src/transport_portals4.c
+++ b/src/transport_portals4.c
@@ -711,7 +711,7 @@ shmem_transport_startup(void)
     ret = PtlCTAlloc(shmem_transport_portals4_ni_h,
                      &shmem_transport_portals4_put_event_ct_h);
     if (PTL_OK != ret) {
-        RAISE_WARN_MSG("Put event CT allocation failed: %d\n", ret);
+        RAISE_ERROR_MSG("Put event CT allocation failed: %d\n", ret);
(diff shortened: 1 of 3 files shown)


jdinan:
@davidozog This cleared testing with the Portals4 IB implementation on the internal cluster.  Ready for review.

src/transport_portals4.c
```diff
@@ -120,25 +115,184 @@ static ptl_pt_index_t heap_pt = PTL_PT_ANY;
...
+    ctx->options = options;
+    ctx->put_md = PTL_INVALID_HANDLE;
+    ctx->get_md = PTL_INVALID_HANDLE;
+    ctx->put_ct = PTL_INVALID_HANDLE;
+    ctx->get_ct = PTL_INVALID_HANDLE;
+    shmem_internal_atomic_write(&ctx->pending_put_cntr, 0);
+    shmem_internal_atomic_write(&ctx->pending_get_cntr, 0);
+
+    /* Allocate put completion tracking resources */
+    ret = PtlCTAlloc(shmem_transport_portals4_ni_h, &ctx->put_ct);
```
jdinan: Quick note -- Portals 4 indicates that this operation (and ```PtlMDBind``` below) should thread safe, and the reference implementation also appears to support this semantic.  /crosses fingers/

No such luck in OFI.  We wrap a mutex around the context resource allocation..

jdinan: @regrant Is this correct?  Are ```PtlCTAlloc``` and ```PtlMDBind``` thread safe?

regrant: @jdinan PtlMDBind and PtlCTAlloc should both be thread safe, and are such in the implementation. You should be good.

src/transport_portals4.c
```diff
@@ -120,25 +115,184 @@ static ptl_pt_index_t heap_pt = PTL_PT_ANY;
...
+    md.options = PTL_MD_EVENT_CT_ACK | PTL_MD_EVENT_SUCCESS_DISABLE;
+    if (1 != PORTALS4_TOTAL_DATA_ORDERING) {
+        md.options |= PTL_MD_UNORDERED;
+    }
+    md.eq_handle = shmem_transport_portals4_eq_h;
+    md.ct_handle = ctx->put_ct;
+    ret = PtlMDBind(shmem_transport_portals4_ni_h, &md, &ctx->put_md);
+    if (PTL_OK != ret) {
+        RETURN_ERROR_MSG("PtlMDBind of put MD failed: %d\n", ret);
+        goto cleanup;
```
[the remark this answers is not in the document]
jdinan: This is a fairly common idiom for this type of "try - catch" in C.  What's nice about structuring the code in this way, is it allows us to do cleanup for resources whose references are only in-scope within this function's body.  It's not the case right now, but could be in the future.  It also keeps all of the code within the same function -- if we pulled this out into a cleanup routine, it would be callable from elsewhere, but probably shouldn't be used except for as cleanup for this routine.

I'm mostly ambivalent, but I think I slightly prefer the goto approach since it is more self-contained and results in less code.  We do have precedent for this type of structure in the Portals ```shmem_transport_startup```.  Anyway, happy to hear arguments in favor of the function approach.

src/transport_portals4.c
```diff
@@ -120,25 +115,184 @@ static ptl_pt_index_t heap_pt = PTL_PT_ANY;
...
+{
+    int ret, id;
+
+    SHMEM_MUTEX_LOCK(shmem_internal_mutex_ptl4_ctx);
+
+    /* Look for an open slot in the contexts array */
+    for (id = 0; id < shmem_transport_portals4_contexts_len; id++)
+        if (shmem_transport_portals4_contexts[id] == NULL) break;
+
+    /* If none found, grow the array */
```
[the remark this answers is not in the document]
jdinan: I had started writing a "growable array" utility library, but once you make the code type agnostic, the amount of code ends up being more than the combined 30 lines from having this pasted into both transport implementations.  We could use utarray (https://troydhanson.github.io/uthash/utarray.html), but we'll end up needing to rewrite all of the array accesses to use this other interface, which doesn't seem to be worth the complexity tradeoff (i.e. to eliminate 15 lines of code).  FWIW, I did cringe when I copy/pasted this code, but subsequently meh'd when I considered the above.

src/transport_portals4.c
```diff
@@ -120,25 +115,184 @@ static ptl_pt_index_t heap_pt = PTL_PT_ANY;
...
+
+        for ( ; i < shmem_transport_portals4_contexts_len; i++)
+            shmem_transport_portals4_contexts[i] = NULL;
+
+        if (shmem_transport_portals4_contexts == NULL) {
+            RAISE_ERROR_STR("Error: out of memory when allocating ctx array");
+        }
+    }
+
+    *ctx = malloc(sizeof(shmem_transport_ctx_t));
```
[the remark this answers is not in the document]
jdinan: Well, the advantage to doing it the way it was done in the OFI code is not writing the bug in 505fd477.  😬

src/transport_portals4.c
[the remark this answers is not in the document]
davidozog: I'm mostly ambivalent as well, just thought I'd mention it since `goto` is so infamous.  I also think this pattern is actually a good use-case for `goto`.

src/transport_portals4.c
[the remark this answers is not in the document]
davidozog: Ha, sounds good.  It is definitely more trouble than it's worth to write a utility, at least at this point in time.

src/transport_portals4.c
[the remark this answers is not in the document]
davidozog: Indeed.  If we ever change that, we'll have to remember to free correctly.

jdinan:
Closes #576

src/transport_portals4.h
```diff
@@ -650,6 +652,30 @@ shmem_transport_put_nbi(shmem_transport_ctx_t* ctx, void *target, const void *so
...
+        shmem_transport_portals4_put_nb_internal(ctx, target, source, len, pe,
+                                                 completion,
+                                                 shmem_transport_portals4_data_pt,
+                                                 shmem_transport_portals4_heap_pt);
+#endif
+    } else {
+        shmem_transport_put_nbi(ctx, target, source, len, pe);
+    }
+}
+
```
davidozog: So if I understand correctly, only the default context can do bounce buffering and the "frag" stuff.  In other words, it won't be optional for user-created contexts to do bounce buffering.  It's no problem, just making sure that's the design choice.

jdinan: Right.  The user can enable the ```SHMEMX_CTX_ISOLATE_PUT``` and ```SHMEMX_CTX_BOUNCE_BUFFER``` on a given context if they want, but they are disabled by default.  We should probably have a unit test for this, come to think of it.

davidozog: If a user only sets `SHMEMX_CTX_BOUNCE_BUFFER` (and not `ISOLATE_PUT`), it won't do bounce buffering, right?  Should there be a warning or something for that case?

jdinan: Oh!  I didn't notice that ```shmem_transport_put_nbi``` doesn't support bounce buffering.  In that case, maybe I shouldn't try to separate these features -- in Portals, ```SHMEMX_CTX_BOUNCE_BUFFER``` enables both optimizations.  Sound ok?

davidozog: Yup!

jdinan: Done.

PR #590, opened 2018-01-29
OFI: Eliminate atomics with FI thread completion

### LINKED ISSUE #546: OFI: Disable atomics when USE_CTX_LOCK
Atomics are not needed when USE_CTX_LOCK is enabled. Wrap these atomics in macros (e.g. 
CNTR_INC/DEC/READ) to allow disabling atomics when using per-context locks.

jdinan:
In FI_THREAD_COMPLETION mode, accesses to context structure fields are
guarded by a mutex.  Thus, atomics on the pending put/get counters are
an unnecessary overhead.

Closes #546

### DIFF
--- a/src/transport_ofi.c
+++ b/src/transport_ofi.c
@@ -1524,8 +1524,11 @@ int shmem_transport_ctx_create(long options, shmem_transport_ctx_t **ctx)
     }
 
     memset(ctxp, 0, sizeof(shmem_transport_ctx_t));
+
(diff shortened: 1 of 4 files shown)


davidozog:
I like it!  A lot!  Consider it approved if it passes testing.

jdinan:
@davidozog Thanks!  I added a simple unit test to ensure that quiet in parallel with one-sided operations performs correctly.  Could you also give this a look?

PR #594, opened 2018-01-30
Multi-threaded Throughput Benchmarks 

wrrobin:


lanl-ompi:

test/performance/shmem_perf_suite/bi_dir.h
```diff
@@ -38,9 +38,13 @@ void static inline bi_bw(int len, perf_metrics_t *metric_info)
             if(i == metric_info->warmup)
                 start = perf_shmemx_wtime();
 
-            for(j = 0; j < metric_info->window_size; j++)
+            for(j = 0; j < metric_info->window_size; j++) {
+#ifdef USE_NONBLOCKING_API
```
kseager: you have putmem/putmem_nbi switched

test/performance/shmem_perf_suite/bi_dir.h
```diff
@@ -49,10 +53,55 @@ void static inline bi_bw(int len, perf_metrics_t *metric_info)
...
+
+    shmem_barrier_all();
+
+    if (streaming_node(*metric_info)) {
+        for (i = 0; i < metric_info->trials + metric_info->warmup; i++) {
+            if(i == metric_info->warmup)
+                start = perf_shmemx_wtime();
 
+            for(j = 0; j < metric_info->window_size; j++) {
+#ifdef USE_NONBLOCKING_API
```
kseager: need to switch getmem/getmem_nbi as well

wrrobin: Thanks @kseager. These are corrected in my local version. Will check them in through my next commit.

test/performance/shmem_perf_suite/bi_dir.h
```diff
@@ -49,10 +53,55 @@ void static inline bi_bw(int len, perf_metrics_t *metric_info)
...
+    } else {
+        for (i = 0; i < metric_info->trials + metric_info->warmup; i++) {
+            for(j = 0; j < metric_info->window_size; j++) {
+#ifdef USE_NONBLOCKING_API
+                shmem_getmem(metric_info->dest, metric_info->src, len, dest);
+#else
+                shmem_getmem_nbi(metric_info->dest, metric_info->src, len, dest);
+#endif
+            }
             shmem_quiet();
```
kseager: previously for get blocking I wasn't using a quiet

test/performance/shmem_perf_suite/bw_common.h
```diff
@@ -278,11 +354,10 @@ void static command_line_arg_check(int argc, char *argv[],
                     " + warmup * sizes (8/4KB) <= max length \n"
                     "[-r number of nodes at target, use only with -t] \n"
                     "[-l number of nodes at initiator, use only with -t, "
-                    "l/r cannot be used together] \n");
+                    "l/r cannot be used together] \n"
+                    "[-C thread-safety-config: SINGLE, FUNNELED, SERIALIZED, or MULTIPLE] \n"
+                    "[-T num-threads] \n");
         }
-#ifndef VERSION_1_0
-        shmem_finalize();
```
kseager: what's the reason for taking out finalize?

kseager:
overall this looks good to me, I focused on the areas that pertained to the refactoring. I agree with the ifdef usage rework that's a much better choice. Thanks for the work

test/performance/shmem_perf_suite/bi_dir.h
wrrobin: Ok. Will keep the quiet call inside the #ifdef block.

test/performance/shmem_perf_suite/bw_common.h
wrrobin: This came from one of the initial patches. Will dig up to find any reason.

test/performance/shmem_perf_suite/bi_dir_ctx.h
```diff
@@ -0,0 +1,98 @@
...
+            }
+#pragma omp barrier
+#pragma omp master 
+            {
+                end = perf_shmemx_wtime();
+            }
+            shmem_ctx_destroy(ctx);
+        }
+        calc_and_print_results((end - start), len, *metric_info);
+    } else {
```
jdinan: This is the `!streaming_node` case -- should it be generating puts?  IIRC, the non-streaming node should be a passive target of RMA operations.

test/performance/shmem_perf_suite/bw_common.h
```diff
@@ -303,6 +378,35 @@ void static inline only_even_PEs_check(int my_node, int num_pes) {
...
+        return "SERIALIZED";
+    } else if (thread_safety == SHMEM_THREAD_MULTIPLE) {
+        return "MULTIPLE";
+    } else {
+        fprintf(stderr, "Unexpected thread safety value: %d\n", thread_safety);
+        exit(1);
+    }
+}
+
+static void inline thread_safety_validation_check(perf_metrics_t *metric_info) {
```
jdinan: 👍

test/performance/shmem_perf_suite/bw_common.h
```diff
@@ -415,8 +521,14 @@ void static inline calc_and_print_results(double total_t, int len,
     PE_set_used_adjustments(&nPEs, &stride, &start_pe, metric_info);
 
     if (total_t > 0 ) {
+
+#ifdef ENABLE_OPENMP
+        bw = (len / 1e6 * metric_info.window_size * metric_info.trials *
+                (double)metric_info.nthreads) / (total_t / 1e6);
```
jdinan: I forget C type promotion rules, but I wonder if there could be loss of precision here since this will be evaluated left-to-right (i.e. `len/1e6` generating an integer result).  I would tend to err on the safe side and change those `1e6` literals to `1.0e6` or throw in an explicit type cast on the left-hand side of the expression.

jdinan: (same comment in the `#else` branch of this code)

test/performance/shmem_perf_suite/bi_dir_ctx.h
wrrobin: @jdinan, For bi-directional bandwidth, shouldn't both nodes inject traffic? This is followed in case of single-threaded version (```bi_dir.h```) as well.

test/performance/shmem_perf_suite/bw_common.h
wrrobin: @jdinan Thanks, good suggestion. Will update it.

test/performance/shmem_perf_suite/bi_dir_ctx.h
jdinan: Yup, you're right.  I got turned around flipping through the code and thought this was a uni directional case.  Please ignore this comment.

wrrobin:
Thanks @davidozog. Good point. I have not yet looked at README. But, we definitely need to add all the flags' info before releasing. Will add it soon.

wrrobin:
@kseager, Made some changes in the ```bw_common.h``` while adding the shmem_finalize. Can you please check and let me know if these changes look ok?

test/performance/shmem_perf_suite/uni_dir.h
```diff
@@ -56,3 +60,38 @@ void inline uni_dir_bw(int len, perf_metrics_t *metric_info)
...
+
+    if (snode) {
+        for (i = 0; i < metric_info->trials + metric_info->warmup; i++) {
+            if(i == metric_info->warmup)
+                start = perf_shmemx_wtime();
+
+            for(j = 0; j < metric_info->window_size; j++) {
+#ifdef USE_NONBLOCKING_API
+                shmem_getmem_nbi(metric_info->dest, metric_info->src, len, dest);
+                shmem_quiet();
```
davidozog: Do we need a `quiet` for each `getmem_nbi`, or can we get away with one after the loop?  Looks to me like blocking and non-blocking will achieve the exact same performance with all the quiets.  However, each get has the same destination... so maybe this is intended?

jdinan: This may be an artifact of the contexts prototype.  I think we only prototyped the nonblocking API, so to get blocking behavior in the benchmark we had to do this.  So, the quiet should probably be deleted.

wrrobin: @jdinan Do you mean that for non-blocking also, quiet is not needed?

jdinan: We should not need to invoke quiet after each operation.  It doesn't look like we're consistent right now about where it's done (per operation, per window, etc.).  Technically what we're doing is undefined by the OpenSHMEM spec (both blocking and nonblocking cases), since each put/get operation is writing to the same location.  But, I think we chose to cross our fingers and ignore that a while back (a comment to document this wouldn't hurt).

wrrobin: Yes, the fact that we are writing in the same location every iteration is semantically different what a real application would do with these APIs. However, performance benchmarking has always been this way, I guess. Alternatively, should we create a series of buffers and write in different locations in each iteration instead? That may require some changes in all the benchmarks but will reflect real application behavior.

jdinan: For now, I think a comment should suffice.

wrrobin: ok, thanks

jdinan: The quiet in the latency test is probably correct.

wrrobin:
Thanks everyone for the feedback. The PR is ready to be merged.

Issue #578, opened 2018-01-23
Gettid hook


jdinan:
Should we add a ```shmemx_register_gettid(uint64_t* (*gettid_fn)(void))``` to allow the user to register a TID query routine?

davidozog:
👍

PR #605, opened 2018-02-12
Pr/gettid register

davidozog:
Addresses issue #578:

* Adds the `shmemx_register_gettid()` function, which allows user to pass a function pointer for setting thread IDs.
* Removes the `TID_T` definition in favor of the `shmem_internal_pid_t` structure, which currently supports either `pid_t` (from gettid syscall) or `uint64_t` TID values.
* Adds a simple unit test, `gettid_register` as a smoke test for creating/using private contexts with a custom gettid routine.

### DIFF
--- a/src/transport_ofi.h
+++ b/src/transport_ofi.h
@@ -233,11 +233,16 @@ typedef struct shmem_transport_ofi_bounce_buffer_t shmem_transport_ofi_bounce_bu
 
 typedef int shmem_transport_ct_t;
 
-#ifdef __APPLE__
-#define TID_TYPE uint64_t
-#else
-#define TID_TYPE pid_t
-#endif
+enum shmem_internal_tid_t { PID_T, UINT64_T };
+struct shmem_internal_tid
+{
+  enum shmem_internal_tid_t tid_t;
+  union
+  {
+    pid_t shmem_internal_pid_t;
+    uint64_t shmem_internal_uint64_t;
+  };
+};
 
 struct shmem_transport_ctx_t {
     int                             id;
@@ -260,7 +265,7 @@ struct shmem_transport_ctx_t {
 #endif
     shmem_free_list_t              *bounce_buffers;
     int                             stx_idx;
-    TID_TYPE                        tid;
+    struct shmem_internal_tid       tid;
 };
 
 typedef struct shmem_transport_ctx_t shmem_transport_ctx_t;
--- a/src/transport_ofi.h
+++ b/src/transport_ofi.h
@@ -239,8 +239,8 @@ struct shmem_internal_tid
   enum shmem_internal_tid_t tid_t;
   union
   {
-    pid_t shmem_internal_pid_t;
-    uint64_t shmem_internal_uint64_t;
+    pid_t pid_val;
+    uint64_t uint64_val;
   };
 };
 
--- a/test/unit/Makefile.am
+++ b/test/unit/Makefile.am
@@ -145,6 +145,7 @@ if HAVE_PTHREADS
 check_PROGRAMS += \
     mt_a2a \
     mt_contention \
+    gettid_register \
     threading \
     web \
     thread_wait
--- a/test/unit/Makefile.am
+++ b/test/unit/Makefile.am
@@ -183,6 +183,10 @@ mt_contention_LDFLAGS = $(PTHREAD_LIBS)
 mt_contention_CFLAGS = -I$(top_srcdir)/test/unit $(PTHREAD_CFLAGS)
 mt_contention_LDADD = $(LDADD) $(PTHREAD_CFLAGS)
 
+gettid_register_LDFLAGS = $(PTHREAD_LIBS)
+gettid_register_CFLAGS = -I$(top_srcdir)/test/unit $(PTHREAD_CFLAGS)
+gettid_register_LDADD = $(LDADD) $(PTHREAD_CFLAGS)
+
 threading_LDFLAGS = $(PTHREAD_LIBS)
 threading_CFLAGS = -I$(top_srcdir)/test/unit $(PTHREAD_CFLAGS)
 threading_LDADD = $(LDADD) $(PTHREAD_CFLAGS)
--- a/test/unit/gettid_register.c
+++ b/test/unit/gettid_register.c
@@ -1,0 +1,118 @@
+/*
+ *  Copyright (c) 2018 Intel Corporation. All rights reserved.
+ *  This software is available to you under the BSD license below:
+ *
+ *      Redistribution and use in source and binary forms, with or
+ *      without modification, are permitted provided that the following
+ *      conditions are met:
+ *
+ *      - Redistributions of source code must retain the above
+ *        copyright notice, this list of conditions and the following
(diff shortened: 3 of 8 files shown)


davidozog:
Should be ready to review, @jdinan and/or @wrrobin.  Thanks!

test/unit/gettid_register.c
```diff
@@ -0,0 +1,121 @@
...
+int dest;
+
+int me, npes;
+int errors = 0;
+
+
+uint64_t my_gettid(void) {
+    static uint64_t tid_val = 100;
+    return tid_val++;
+}
```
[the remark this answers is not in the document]
davidozog: I think there would need to be a type-cast from `pthread_t` to `uint64_t` at the moment.  But if that's ok, it would be a more interesting example.

src/transport_ofi.h
```diff
@@ -233,11 +233,16 @@ typedef struct shmem_transport_ofi_bounce_buffer_t shmem_transport_ofi_bounce_bu
 
 typedef int shmem_transport_ct_t;
 
-#ifdef __APPLE__
-#define TID_TYPE uint64_t
-#else
-#define TID_TYPE pid_t
-#endif
+enum shmem_internal_tid_t { PID_T, UINT64_T };
```
jdinan: Seems possible these names could clash since they differ from the types only in case.  What about something like `tid_is_pid_t` and `tid_is_uint64_t`?

src/transport_ofi.h
```diff
@@ -233,11 +233,16 @@ typedef struct shmem_transport_ofi_bounce_buffer_t shmem_transport_ofi_bounce_bu
...
+struct shmem_internal_tid
+{
+  enum shmem_internal_tid_t tid_t;
+  union
+  {
+    pid_t pid_val;
+    uint64_t uint64_val;
+  };
```
jdinan: 4 spaces per indent, please.  :)

jdinan: `:set shiftwidth=4`

test/unit/Makefile.am
```diff
@@ -182,6 +183,10 @@ mt_contention_LDFLAGS = $(PTHREAD_LIBS)
 mt_contention_CFLAGS = -I$(top_srcdir)/test/unit $(PTHREAD_CFLAGS)
 mt_contention_LDADD = $(LDADD) $(PTHREAD_CFLAGS)
 
+gettid_register_LDFLAGS = $(PTHREAD_LIBS)
```
jdinan: Add `$(AM_LDFLAGS)` per #606.

test/unit/Makefile.am
```diff
@@ -182,6 +183,10 @@ mt_contention_LDFLAGS = $(PTHREAD_LIBS)
...
+gettid_register_CFLAGS = -I$(top_srcdir)/test/unit $(PTHREAD_CFLAGS)
```
jdinan: You can safely remove `-I$(top_srcdir)/test/unit`.  This is only needed if the unit test includes the `pthread_barrier` shim header.

test/unit/gettid_register.c
```diff
@@ -0,0 +1,121 @@
...
+#define T 8
+
+int dest;
+
+int me, npes;
+int errors = 0;
+
+
+uint64_t my_gettid(void) {
+    uint64_t tid_val = (uint64_t)pthread_self();
```
jdinan: You may want to add a sanity check in `main` to the effect:
```
if (sizeof(pthread_t) < sizeof(uint64_t)) {
  printf("Can't run this test, sorrynotsorry\n");
  shmem_finalize();
  return 0;
}
```

test/unit/gettid_register.c
```diff
@@ -0,0 +1,121 @@
...
+    return tid_val;
+}
+
+
+static void * thread_main(void *arg) {
+    int tid = * (int *) arg;
+    int i;
+
+    shmem_ctx_t ctx;
+    int ret = shmem_ctx_create(SHMEM_CTX_PRIVATE, &ctx);
```
jdinan: Oh, this is a bold move.  I sure hope those tids are unique.  🙈 🙉 🙊

test/unit/gettid_register.c
davidozog: Should that be `>`?  The concern is losing bits from the cast, right?

Also, it looks like a POSIX `pthread_t` could be a struct:
"IEEE Std 1003.1-2001/Cor 2-2004, item XBD/TC2/D6/26 is applied, adding pthread_t to the list of types that are not required to be arithmetic types, thus allowing pthread_t to be defined as a structure."

Maybe this cast will come back to bite us?

davidozog:
Note: 2 rows in Travis failed from that same "write error" I mentioned in #613.  These errors always occur during the tests-uh compilations.

test/unit/gettid_register.c
jdinan: It probably will (see comment below -- if it's a struct, you're likely to get non-unique TIDs by just looking at the first 8 bytes, which will break private STX assignment).  We do pass a `tid` as an argument to `thread_main`.  Why don't you move that into a global `_Thread_local int tid` variable and return that value from `my_gettid`?  `_Thread_local` is a C11-ism, so you should wrap this in an preprocessor check for C11 support and just return success if the compiler doesn't support C11.  Alternatively (and probably easier from a portability perspective), you can use `pthread_set/get_specific`.  Just create the key before you spawn threads to avoid needing the synchronize on key creation.

davidozog: 👍 I'll look into the `pthread_setspecific` option.

test/unit/gettid_register.c
```diff
@@ -41,19 +42,26 @@ int dest;
...
     return tid_val;
 }
 
 
 static void * thread_main(void *arg) {
-    int tid = * (int *) arg;
+    uint64_t tid = * (uint64_t *) arg;
     int i;
 
+    int ret = pthread_setspecific(key, &tid);
```
jdinan: If you don't need `tid`, you can put `arg` into the KVS here.

test/unit/gettid_register.c
```diff
@@ -41,19 +42,26 @@ int dest;
 int me, npes;
 int errors = 0;
 
+pthread_key_t key;
 
 uint64_t my_gettid(void) {
-    uint64_t tid_val = (uint64_t)pthread_self();
+    uint64_t tid_val = 0;
+
+    memcpy(&tid_val, pthread_getspecific(key), sizeof(uint64_t));
```
jdinan: I think you can simplify this to:
```
tid_val = * (uint64_t*) pthread_getspecific(key);
```

wrrobin:
@davidozog Getting an error in unit tests as below. Let me know if I missed anything.
```
test/unit/gettid_register.c:84: undefined reference to `shmemx_register_gettid'
```

davidozog:
Hmm... that's strange.  You might have to make clean then make install...?  You're in Ubuntu?

wrrobin:
Thanks. Cleaning resolved the issue though I thought I did clean before. Seeing some new warnings in transport_ofi.c but may not be relevant to this.

PR #607, opened 2018-02-13
Unit test for memory barrier

wrrobin:
Supports unit test for shmem_fence(), wait_until(), and lock APIs.


lanl-ompi:

wrrobin:
@jdinan, @davidozog, please review and let me know any concerns as well as what other tests we can add here.

test/unit/mt_membar.c
```diff
@@ -0,0 +1,189 @@
+#include <stdio.h>
```
jdinan: Please add the copyright statement.

test/unit/mt_membar.c
```diff
@@ -0,0 +1,189 @@
...
+#endif
+
+int shared_dest_1 = 0, shared_dest_2 = 0, result = 0;
+int me, npes, errors = 0, sum_error = 0;
+long lock = 0;
+
+pthread_barrier_t fencebar;
+
+long pSync[_SHMEM_REDUCE_SYNC_SIZE];
+int pWrk[MAX(1, _SHMEM_REDUCE_MIN_WRKDATA_SIZE)];
```
jdinan: The `_SHMEM*` constants are deprecated, please use the `SHMEM*` name for these constants.

test/unit/mt_membar.c
```diff
@@ -0,0 +1,189 @@
...
+        } else {
+            return ret;
+        }
+    }
+
+    me = shmem_my_pe();
+    npes = shmem_n_pes();
+
+    for (i = 0; i < _SHMEM_REDUCE_SYNC_SIZE; i++) {
+        pSync[i] = _SHMEM_SYNC_VALUE;
```
jdinan: More deprecated constant names.

test/unit/mt_membar.c
```diff
@@ -0,0 +1,189 @@
...
+        assert(0 == err);
+    }
+
+    for (i = 0; i < T; i++) {
+        int err;
+        err = pthread_join(threads[i], NULL);
+        assert(0 == err);
+    }
+
+    shmem_sync_all();
```
jdinan: Is this sync necessary?  The SHMEM barrier below also appears to be extraneous, but is probably safe programming practice since there isn't a barrier in `main` between initializing `psync` and performing the reduction.  Thus, a change in `thread_main` could introduce a race on `psync`.

test/unit/mt_membar.c
```diff
@@ -0,0 +1,189 @@
...
+        if (result != T * ITER) {
+            printf("ERROR in WAIT test from %d : result = %d, expected = %d\n", 
+                    me, result, T * ITER);
+        }
+        result = 0;
+	shared_dest_1 = 0;
+    }
+
+    pthread_barrier_wait(&fencebar);
+    if (0 == tid) shmem_barrier_all();
```
jdinan: Is this SHMEM barrier necessary?  If all communication is to self, there shouldn't be a need to synchronize with other PEs?  Although, maybe the quiet is needed (I don't know if pthread barrier implies a full memory barrier or just thread synchronization).

test/unit/mt_membar.c
```diff
@@ -0,0 +1,189 @@
...
+    int tid = *(int *) arg;
+    int one = 1, zero = 0;
+    int i;
+
+    /* TEST WAIT */
+    for (i = 0; i < ITER; i++) {
+        if (tid == 0) {
+            shared_dest_1 = 1;
+            shmem_int_wait_until(&shared_dest_1, SHMEM_CMP_EQ, zero);
+            shmem_int_atomic_add(&result, one, me);
```
jdinan: Can we replace this with:
```
result++;
shmem_fence();
```
IIUC, we should have atomicity from the "token" that is being passed around via `shared_dest_1`.  We just need to flush writes before passing the token, which is what `shmem_fence` is supposed to handle.

jdinan: Note this substitution is for both threads 1 and 2.

test/unit/mt_membar.c
```diff
@@ -0,0 +1,189 @@
...
+            shmem_fence();
+        }
+    }
+
+    pthread_barrier_wait(&fencebar);
+    if (0 == tid) shmem_barrier_all();
+    pthread_barrier_wait(&fencebar);
+
+    if (tid == 0) {
+        errors += result == T * ITER ? 0 : 1;
```
jdinan: I _think_ operator precedence should get you this grouping `errors += ((result == (T * ITER)) ? 0 : 1)` but the parens would be super helpful for those of us that haven't memorized the C operator precedence table.  🤕

test/unit/mt_membar.c
```diff
@@ -0,0 +1,189 @@
...
+        }
+
+        if (tid == 1) {
+            shmem_int_wait_until(&shared_dest_2, SHMEM_CMP_EQ, one);
+            shmem_int_atomic_add(&result, shared_dest_1 + shared_dest_2, me);
+            shared_dest_2 = 0;
+            shared_dest_1 = 0;
+            shmem_fence();
+        }
+    }
```
jdinan: Nice.  👍

test/unit/mt_membar.c
```diff
@@ -0,0 +1,189 @@
...
+
+    if (tid == 0) {
+        errors += result == T * ITER ? 0 : 1;
+        if (result != T * ITER) {
+            printf("ERROR in FENCE test from %d : result = %d, expected = %d\n",
+                    me, result, T * ITER);
+        }
+        result = 0;
+	shared_dest_1 = 0;
+	shared_dest_2 = 0;
```
jdinan: Are there hard tabs in here?  If you use vim, you can get it to highlight hard tabs by adding this to your .vimrc:
```
" Show special representation for hard tabs
set listchars=tab:>-
set list
```

test/unit/mt_membar.c
```diff
@@ -0,0 +1,189 @@
...
+            shmem_set_lock(&lock);
+            shared_dest_1 = 1;
+            shmem_clear_lock(&lock);
+            shared_dest_2 = 1;
+            shmem_int_wait_until(&shared_dest_2, SHMEM_CMP_EQ, zero);
+        }
+
+        if (tid == 1) {
+            shmem_int_wait_until(&shared_dest_2, SHMEM_CMP_EQ, one);
+            shmem_int_atomic_add(&result, shared_dest_1, me);
```
jdinan: Does this need to be an atomic?

test/unit/mt_membar.c
wrrobin: It is true that each PE can proceed not depending on others as the communications are with self. But, I just wanted to keep them in sync as they are using the shared variables inside each test. From the description of the pthread_barrier_wait (http://pubs.opengroup.org/onlinepubs/009695399/functions/pthread_barrier_wait.html), it looks like that is it responsible only to synchronize all the threads.

test/unit/mt_membar.c
wrrobin: Thanks for the suggestion. After your comment, I also got confused which operator precedes which and I definitely dont remember the table :P

test/unit/mt_membar.c
wrrobin: No. I changed it to an assignment. Thanks.

test/unit/mt_membar.c
wrrobin: I think the sync is extra. Added it before I wrote anything in the thread_main and then forgot to remove this one. Removing this. Keeping the barrier though just for safety.

wrrobin:
@jdinan, Can you please review the PR again? Also, I have few questions that I left in comments.

test/unit/mt_membar.c
jdinan: Ok -- makes sense.  Do you think this would be potentially useful as another test?
```
shmem_int_wait_until(&shared_dest_1, SHMEM_CMP_EQ, tid);
result++;
shmem_fence();
shared_dest_1 = (tid+1) % T;
```

jdinan:
Looks good to me.  Only remaining question is whether we would increase coverage by adding the `results++` case that I mentioned.  The non-atomic read-modify-write is a bit scary, which is nice.  :)

wrrobin:
Thanks @jdinan. I will add that test. Do you think of any other test that we can add (i.e. I did not add quiet, barrier since the membar functions for those have already been tested by the above)?

jdinan:
This looks like a good checkpoint to merge, but you're welcome to make a hobby of adding cases to this test.  :neckbeard:

PR #610, opened 2018-02-15
Bug fix: sync before start and stop timer for multiple PE BW experiments

wrrobin:


lanl-ompi:

wrrobin:
@jdinan Is the "Resource temporarily unavailable" error in Travis related to insufficient space in disk? Seems like one test failed because of this and the other one timed out in ISx.

wrrobin:
@jdinan Looks like the restarted jobs passed. This PR is ready for review. 
@kseager, @davidozog please also review and let me know any thoughts you might have.

test/performance/shmem_perf_suite/bi_dir.h
```diff
@@ -47,12 +44,34 @@ void static inline bi_bw_put(int len, perf_metrics_t *metric_info)
...
+            for(j = 0; j < metric_info->window_size; j++) {
+#ifdef USE_NONBLOCKING_API
+                shmem_putmem_nbi(metric_info->dest, metric_info->src, len, dest);
+#else
+                shmem_putmem(metric_info->dest, metric_info->src, len, dest);
+#endif
+            }
+            shmem_quiet();
+        }
+    }
```
jdinan: Are the if and else cases identical?  If so ... why have the conditional?

wrrobin: oh, yes. Sorry that I missed this. Earlier the if and else cases was different because of the timers. But now since we are handling this outside this scope, we can code this better. This change will only impact the bi-dir case.

wrrobin:
@kseager, @jdinan This PR is ready for re-review.

jdinan:
@kseager Could you give this a final look before we merge?  Thanks!

test/performance/shmem_perf_suite/bi_dir.h
kseager: I agree for the warm-up the cases are doing the same thing, and this is also true for the second section where we are doing timing and all we really need is 
 if (streaming_node(*metric_info)) {
        start = perf_shmemx_wtime();

kseager:
thanks for the fix, I agree with these changes. Since we are on this topic, what about target_put.h? 
should we move the calc_and_print from the target_node side in target_data_uni_bw to be after the shmem_barrier?

 if(snode || target_node(metric_info))
        calc_and_print_results((end - start), len, metric_info);

wrrobin:
@kseager thanks for raising this point. I was looking at the target_put.h first time and looks like the time calculated includes both the warmup and trial puts. Also, both get and put bw routines invoke the same target bw routine which measured put bw only. Should we also make changes to these?

wrrobin:
@kseager Please review the changes and suggest any update necessary.

kseager:
this looks good thanks!

wrrobin:
Thanks @kseager. 
@jdinan, @davidozog if you guys have any comments, let me know. Otherwise, this is ready to be merged.

PR #613, opened 2018-02-20
Use DEFAULT_POLL_LIMIT in shmem_env_defs.h

davidozog:
This must have been an oversight when breaking apart the TX and RX poll limits.  We need to set the poll limit value to either 0 or -1, depending on whether the user passes `--enable-completion-polling` or not.
  
### DIFF
--- a/src/shmem_env_defs.h
+++ b/src/shmem_env_defs.h
@@ -13,6 +13,10 @@
  *
  */
 
+/* This header depends on 'config.h', which cannot be included here. Please
+ * include 'config.h' before including this file, 'shmem_env_defs.h'
+ */
+
 /* SOS_ENV( name, kind, default, category, short description )
  *
  * Kinds: long, size, bool, string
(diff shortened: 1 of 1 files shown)


jdinan:
Observation -- This adds a dependence on `config.h` but we can't include it in this header because of how/where it gets included (could add a comment to this effect).  The alternative is to generate the env-defs header from a .h4.  Although I'm not sure that is a great option (more moving parts on something that's already pretty complicated).

davidozog:
Yup, I understand.  I also agree that a comment is better here until we have a better motivation to generate the env-defs.  Are you thinking a comment at the top of `shmem_env_defs.h`?  Something like, "This header depends on 'config.h', which cannot be included here.  Please include config.h before including shmem_env_defs.h".

davidozog:
I've seen a strange "write error" a couple of times so far in Travis for this PR.  This issue appears to be tracked here:
https://github.com/travis-ci/travis-ci/issues/4704

I restarted those tests to see if it's temporary.

Issue #614, opened 2018-02-21
Make lock API thread-friendly


jdinan:
See: https://github.com/openshmem-org/specification/issues/205

jdinan:
This is trending as wontfix, per OpenSHMEM threading WG.

davidozog:
Closed by #658

PR #616, opened 2018-02-22
Add threading support to SHMEM lock API

jdinan:
Add support for multiple threads to invoke SHMEM lock operations concurrently.  This works by introducing a local queuing lock in front of every SHMEM lock.  Threads must first obtain the local "guard" lock before requesting the SHMEM lock.  One thread per PE participates in the SHMEM lock algorithm (MCS queueing lock) at a time.  This is necessary because SHMEM's distributed lock structure doesn't provide enough space to enqueue NPES*T (where T is threads per PE) requests.

In 9ebe353 the guard lock is a PThread mutex.  Commit b7a1630 updates the guard lock to a queueing lock to provide FCFS prioritization among the threads in a given PE (SHMEM locks require FCFS prioritization across PEs).  This is nearly, but not quite, FCFS across all threads in a given SHMEM job.


jdinan:
Closes #614

jdinan:
Open questions:
* Do we want the local lock to spin instead of blocking on a cond variable?
* Do we want a local lock that uses atomics instead of PThread mutexes (e.g. CLH lock)?
* Do we want to forward the lock locally before releasing it (e.g. add an env var to forward N times before releasing the SHMEM lock)?

src/shmem_lock.c
```diff
@@ -0,0 +1,173 @@
...
+static inline void shmem_internal_qlock_lock(shmem_internal_lock_guard_t *g) {
+    int ret;
+    char errmsg[256];
+    uint64_t my_ticket;
+
+    ret = pthread_mutex_lock(&g->mutex);
+    if (ret) RAISE_ERROR_MSG("pthread_mutex_lock failed: %s\n",
+                             shmem_util_strerror(ret, errmsg, 256));
+
+    my_ticket = g->next_ticket++;
```
davidozog: Does this (and others below) need to be an atomic read and inc?

jdinan: Nope, it's guarded by `g->mutex`.

davidozog: Doesn't the `pthread_cond_wait` release the mutex?

jdinan: `pthread_cond_wait` also reacquires the mutex before it returns.

jdinan:
Update based on recent discussion in the OpenSHMEM threads working group:

The committee is leaning toward requiring the locking provided in this PR to be implemented at the application layer.  We may end up moving this code to a "companion library" so that users can easily incorporate it in their code.  With that in mind, I'll move 9ebe3539 to its own PR to cherry pick merge the `strerror_r` fixes.

Issue #147, opened 2016-03-16
Tests as separate sub-module


naveen-rn:
Unit test cases in this repository looks very comprehensive and also tests OpenSHMEM-1.3 features. If it is possible, can we make this as separate repository and add as a submodule in this repo. This would allow to use these tests in other SHMEM implementations.

jdinan:
I think we're open to this, but we don't have the developer resources to tackle it at the moment.  I'm not sure that a git submodule is the best long-term approach, since we're using automake to run tests and others would presumably want to use a different harness.

Do you have any suggestions in terms of a test harness or an example that you think was done well?  It looks like the test harness used by [openshmem-org/tests-cray](https://github.com/openshmem-org/tests-cray) are roughly equivalent to what automake does for us currently.  I've also invested a fair amount of effort into improving the [openshmem-org/tests-uh](https://github.com/openshmem-org/tests-uh) harness; that could be another option.

If it's useful enough in your work, I'd be happy to talk about working together on this.

sayantansur:
+1 to ONE community maintained testing harness and tests.

jdinan:
This issue is now closed -- https://github.com/openshmem-org/tests-sos :tada:

PR #619, opened 2018-02-25
Allow tests to be run external to SOS repository

jdinan:
Update the test suite to enable building/running external to the SOS repository.

Ref #147


jdinan:
@davidozog Ready for your perusal.  Once this is merged, we can publish tests-sos.

One note for review -- I removed the `--disable-openmp` option since this is already covered by `AC_OPENMP` and was causing the option to appear twice when you run `configure --help`.

test/unit/Makefile.am
```diff
@@ -145,11 +145,15 @@ if HAVE_PTHREADS
...
     mt_contention \
-    gettid_register \
     mt_membar \
     threading \
     web \
     thread_wait
+
+if SHMEMX_TESTS
+check_PROGRAMS += \
+    gettid_register
```
davidozog: Do we also need `pcontrol` and `shmem_ct` here?

jdinan: It's a little sloppy, but those are each handled through separate paths.  The `shmem_ct` test is effectively disabled externally by setting `USE_PORTALS4` to `false` in the tests-sos `configure.ac`.  The `pcontrol` test is enabled/disabled (disabled by default) with a separate `--enable-profiling` option that sets the AM conditional `ENABLE_PROFILING`.

PR #628, opened 2018-03-07
Add probe routine for making manual progress

davidozog:
Adds a static inline `shmem_transport_probe()` routine for making manual-progress and the `--enable-manual-progress` configure option to enable intermittently calling the routine.  In the OFI transport, the probe reads the target (RX) counter.  In other transports, a probe does nothing.

If enabling manual progress, the OFI probe routine is called in "put_quiet", and "get_wait" and the various "shmem_wait" routines.  We might want to think about placing the probe elsewhere as well, especially for the case with multiple STX's on the libfabric PSM2 provider.

### DIFF
--- a/src/transport_ofi.h
+++ b/src/transport_ofi.h
@@ -1170,39 +1170,29 @@ static inline
 uint64_t shmem_transport_received_cntr_get(void)
 {
 #ifndef ENABLE_HARD_POLLING
-#ifdef USE_THREAD_COMPLETION
-    uint64_t cnt = 0;
-    if (0 == pthread_mutex_trylock(&shmem_transport_ofi_progress_lock)) {
-        cnt = fi_cntr_read(shmem_transport_ofi_target_cntrfd);
-        pthread_mutex_unlock(&shmem_transport_ofi_progress_lock);
-    }
-    return cnt;
-#else
+    shmem_internal_assert(shmem_internal_thread_level == SHMEM_THREAD_SINGLE);
+    /* NOTE-MT: This is only reachable in single-threaded runs, otherwise
+     * we would need a mutex to support FI_THREAD_COMPLETION builds. */
     return fi_cntr_read(shmem_transport_ofi_target_cntrfd);
-#endif /* USE_THREAD_COMPLETION */
 #else
     RAISE_ERROR_STR("OFI transport configured for hard polling");
-#endif /* ENABLE_HARD_POLLING */
+    return 0;
+#endif
 }
 
 static inline
 void shmem_transport_received_cntr_wait(uint64_t ge_val)
 {
 #ifndef ENABLE_HARD_POLLING
-#ifdef USE_THREAD_COMPLETION
-    if (0 == pthread_mutex_trylock(&shmem_transport_ofi_progress_lock)) {
-        int ret = fi_cntr_wait(shmem_transport_ofi_target_cntrfd, ge_val, -1);
-        OFI_CHECK_ERROR(ret);
-        pthread_mutex_unlock(&shmem_transport_ofi_progress_lock);
-    }
-#else
+    shmem_internal_assert(shmem_internal_thread_level == SHMEM_THREAD_SINGLE);
+    /* NOTE-MT: This is only reachable in single-threaded runs, otherwise
+     * we would need a mutex to support FI_THREAD_COMPLETION builds. */
     int ret = fi_cntr_wait(shmem_transport_ofi_target_cntrfd, ge_val, -1);
+
     OFI_CHECK_ERROR(ret);
-#endif /* USE_THREAD_COMPLETION */
 #else
     RAISE_ERROR_STR("OFI transport configured for hard polling");
-#endif /* ENABLE_HARD_POLLING */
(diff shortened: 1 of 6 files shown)


src/transport_ofi.h
```diff
@@ -308,6 +310,17 @@ extern struct fid_ep* shmem_transport_ofi_target_ep;
             shmem_free_list_unlock(ctx->bounce_buffers);                        \
     } while (0)
 
+static inline
+void shmem_transport_probe(void) {
+#if defined(ENABLE_MANUAL_PROGRESS) && !defined(HARD_POLLING)
+    if (0 == pthread_mutex_trylock(&shmem_transport_ofi_progress_lock)) {
```
jdinan: The lock is only required when `USE_THREAD_COMPLETION` is defined.  Not sure if there is any benefit from having a single thread poke OFI when `FI_THREAD_SAFE` is used.

src/transport_ofi.h
```diff
@@ -308,6 +310,17 @@ extern struct fid_ep* shmem_transport_ofi_target_ep;
             shmem_free_list_unlock(ctx->bounce_buffers);                        \
     } while (0)
 
+static inline
+void shmem_transport_probe(void) {
+#if defined(ENABLE_MANUAL_PROGRESS) && !defined(HARD_POLLING)
```
jdinan: I think `HARD_POLLING` here might be a looming bug.  This gets switched on automatically when XPMEM is enabled, so enabling XPMEM could have the second order effect of disabling manual progress.

src/transport_ofi.h
```diff
@@ -1153,29 +1170,39 @@ static inline
 uint64_t shmem_transport_received_cntr_get(void)
```
jdinan: The changes to the received_cntr_get/wait functions seem weird, since these functions already generate progress.  IIRC, these changes were part of a hack to use these routines as stand-ins for a proper "probe" routine during earlier exploratory work.  We do need to respect the received_cntr lock here, but we should be using "probe" to generate progress instead of using these routines.

src/transport_ofi.h
jdinan: IIRC, the `target_cntr` is disabled when hard polling is enabled.  We might need to update that to hard polling enabled and manual progress disabled.

src/transport_ofi.h
davidozog: Oh yeah - it looks like I didn't see that using these routines earlier was just a hack for probe.  I'll just back out of these changes.

src/transport_ofi.h
davidozog: Not sure I fully understand.  Do you mean that we should consider changing instances of
`#ifndef HARD_POLLING`
to
`#if !defined(HARD_POLLING) || defined(ENABLE_MANUAL_PROGRESS)`
which guard `target_cntr`?  If so, I'm not sure I completely appreciate the consequences of enabling hard polling and manual progress.

jdinan: The hard polling code was fairly well isolated before threading support was added, but it's less so now.  The `HARD_POLLING` parameter was originally meant to influence how shmem_wait operations are implemented -- do we poll memory or block on a transport-layer event (see shmem_synchronization.c:120).  Presently, when you enable hard polling, we don't need the `target_cntr`, so it gets compiled out (which makes the GNI provider happy, so it's important we continue to support this mode).  However, this PR adds a second case where we need the `target_cntr` -- manual progress.

Anyway, long winded way to say "yes" I think we need to update the `target_cntr` preprocessor magic so we get the cntr when needed to support either non-hard polling (which we still support in single-threaded runs without XPMEM) or manual progress.

davidozog: Ok, I just pushed commit 4d1b444 to try to address this.  I'm still not quite sure if things get screwy when  simultaneously enabling both hard polling and manual progress.  In that case, does `query_for_fabric()` look ok with the `RMA_EVENT` and `MR_RMA_EVENT` settings?  Thanks!

That commit also allows the probe even with `FI_THREAD_SAFE` - I figure it could be useful to try out later, but yeah, it's probably less important than with `FI_THREAD_COMPLETION`.

jdinan:
PR looks good.  The `ENABLE_TARGET_CNTR` intermediate is nice, in terms of developer sanity.  Before merging, please test the different build configs by hand (I don't think we have all these combinations covered in CI) and verify that the progress looks good on PSM2 with the BW/Latency tests.

davidozog:
Sounds good, thanks.

davidozog:
This is good to merge.  If `--hard-polling`, the target counter is disabled.  But beware that with both `--hard-polling` and `--enable-manual-progress`, the target counter is enabled.

Interesting side note:  With 1 thread in THREAD_MULTIPLE mode (where we've had serious performance problems before),  I'm now seeing substantial performance improvement on PSM2 with manual progress.  Manual progress (alone) actually looks quite a bit better than with `FI_PSM2_PROG_INTERVAL`=1, 10, 100, or 1000 alone; and perhaps slightly better than both `PROG_INTERVAL` and manual progress simultaneously.

PR #629, opened 2018-03-07
Add pkg-config file

### LINKED ISSUE #585: Provide pkg-config file for Sandia OpenSHMEM
### Feature Request

Provide `pkg-config` file(s) for Sandia OpenSHMEM.

jdinan:
Closes #585


jdinan:
Heads-up that this patch also changes the distribution tarball name from "sandia-openshmem-X.Y.Z.tar.gz" to "sos-X.Y.Z.tar.gz" (we weren't setting the tarname previously, but need to set it now in order to set the package URL in AC_INIT).  It's also easy to keep the old name, if anyone prefers that.

jdinan:
Oh, good, Travis' autoconf is too old to provide `PKG_INSTALLDIR`... 🤦‍♂️

jdinan:
Another note: `PKG_INSTALLDIR` is provided by `pkg-config`, so this patch also adds a build system dependence on having pkg-config installed.

jdinan:
Looks like Travis is having this problem: https://stackoverflow.com/questions/40427879/graph-tool-autoconf-configure-ac39-error-possibly-undefined-macro-pkg-inst

The `PKG_INSTALLDIR` macro is not provided prior to v0.27.

jdinan:
Looks like OSHMEM also has oshmem-c.pc, oshmem-cxx.pc, and oshmem-fort.pc [1].  Should probably look into the same for SOS.

[1] https://github.com/open-mpi/ompi/issues/4730

jdinan:
The other way to provide C++/Fortran flags is to define variables that can be queried via `pkg-config --variable=cxxflags sos`.  This looks a bit easier to support, in terms of build system tomfoolery/copy-paste tradeoffs.  Here's an example from MPICH: https://github.com/pmodels/mpich/blob/master/src/packaging/pkgconfig/mpich.pc.in

@nspark do you have a preference between pkg-config variables versus additional packages (sos-cxx, sos-fort, etc) for providing C++/Fortran flags?

nspark:
I don't really have a preference between the two approaches. I think either one will work for us.

FWIW, we having a growing set of tools and applications that rely on CMake. The `--variable` interface for `pkg-config` only seems to be supported in [CMake >=3.4](https://cmake.org/cmake/help/v3.4/module/FindPkgConfig.html). That should still be for us, as we generally use newer versions of CMake and are almost exclusively C developers anyway.

jdinan:
Apparently there is a bug in older versions of pkg-config that interferes with the rpath arguments that SOS wants to store in libs:
```
 line>Libs: -L${libdir}   -L/home/travis/travis/install/libfabric//lib    -Wl,-rpath -Wl,/home/travis/travis/install/libfabric//lib -Wl,-rpath -Wl,${libdir} -Wl,--enable-new-dtags -lsma   -lpmi_simple
...
 removing duplicate "-Wl,-rpath"
...
-Wl,-rpath -Wl,/home/travis/travis/install/libfabric//lib -Wl,/home/travis/travis/install/sandia-shmem-ofi/lib -Wl,--enable-new-dtags -L/home/travis/travis/install/sandia-shmem-ofi/lib -L/home/travis/travis/install/libfabric//lib -lsma -lpmi_simple
```
Notice that the second `-Wl,-rpath` is removed, resulting in a linker error.  Thanks, pkg-config!!  🤣

This appears to have been fixed in pkg-config 0.28, released 1/24/2013:
```
pkg-config 0.28
===
- Fixed a pair of long-standing and intertwined bugs involving unwanted
  removal of flags. The first is that other Libs flags like -Wl are now
  kept in context order with -l flags. The second is that aggressive
  removal of all duplicate arguments has been scaled back so that just
  consecutive duplicate arguments are removed. One result of this change
  is that some flags could be repeated in the final output, especially
  flags from non-pkg-config packages like -lm. Since pkg-config rarely
  has enough knowledge here about the right thing to do, we throw the
  duplicate arguments at the compiler/linker and trust it will do the
  right thing.
```

jdinan:
Anyway, I'm giving up on using pkg-config in Travis, since it's working with pkg-config 0.26.

jdinan:
This PR is a mess.  I'm going to rebase/fix the branch and post a fresh PR with less aggregate pkg-sadness.

nspark:
Wow, I'm surprised such small thing has caused so much trouble. Carry on, @jdinan!

jdinan:
There is no easy build system work.  😅

PR #630, opened 2018-03-08
Removing duplicated ctx destroy

shawone:
shmem_transport_ctx_destroy() is already called from cleanup_handles()

### DIFF
--- a/src/transport_portals4.c
+++ b/src/transport_portals4.c
@@ -321,8 +321,6 @@ init_long_frag(shmem_free_list_item_t *item)
 static void
 cleanup_handles(void)
 {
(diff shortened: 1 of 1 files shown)


shawone:
I have no preference, but if it's possible to remove one of them it's could be great.
Please let me know If you prefere issue in github nex time. I can update the PR if you want.
Thanks

jdinan:
Hey @shawone, thanks for the PR!  👍 

I think I would prefer to free all of the contexts together in `shmem_transport_fini` and delete the older code that cleans up only `SHMEM_CTX_DEFAULT` in `cleanup_handles`.  There shouldn't be a need for the default context to stay active longer than any of the others and keeping all of this code in one place reduces the possibility for another error like this in the future.

jdinan:
Thanks!

PR #633, opened 2018-03-09
Use HASH_FIND/ADD, not HASH_FIND/ADD_INT for tids

davidozog:
When I added support for multiple tid types, I forgot to update the UT_hash table to use the struct as a key instead of the int.  This led to problems when allocating OFI STX's for private contexts.

### DIFF
--- a/src/transport_ofi.c
+++ b/src/transport_ofi.c
@@ -104,6 +104,7 @@ static inline
 struct shmem_internal_tid shmem_transport_ofi_gettid(void)
 {
     struct shmem_internal_tid tid;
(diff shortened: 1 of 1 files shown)


src/transport_ofi.c
```diff
@@ -104,6 +104,7 @@ static inline
 struct shmem_internal_tid shmem_transport_ofi_gettid(void)
 {
     struct shmem_internal_tid tid;
+    memset(&tid, 0, sizeof(struct shmem_internal_tid));
```
jdinan: Nice.  💯

jdinan:
Looks good to me!  Feel free to merge whenever you're ready.

PR #634, opened 2018-03-12
Add env var to disable private contexts in OFI

davidozog:
This adds an environment variable that disallows private contexts from claiming exclusive access to an STX in the OFI layer.  This may improve performance for multi-threaded apps that use more exclusive thread-private contexts than available STXs.

### DIFF
--- a/README
+++ b/README
@@ -231,6 +231,12 @@ options.
        available for private use (i.e., with contexts that enable the
        SHMEM_CTX_PRIVATE option).
 
+    SHMEM_OFI_STX_DISABLE_PRIVATE (default: 0)
+       Disallow private contexts from having exclusive STX access.  Setting this
+       to 1 may improve performance for applications using many private
+       contexts, especially if the number of threads that use private contexts
+       is greater than SHMEM_OFI_STX_MAX.
+
   Debugging Environment variables:
 
     SHMEM_DEBUG (default: off)
--- a/README
+++ b/README
@@ -231,11 +231,10 @@ options.
        available for private use (i.e., with contexts that enable the
        SHMEM_CTX_PRIVATE option).
 
-    SHMEM_OFI_STX_DISABLE_PRIVATE (default: 0)
-       Disallow private contexts from having exclusive STX access.  Setting this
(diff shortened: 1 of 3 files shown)


davidozog:
@jdinan, this is ready for a review when you are.

src/transport_ofi.c
```diff
@@ -1432,6 +1443,12 @@ int shmem_transport_init(void)
         shmem_transport_ofi_stx_allocator = ROUNDROBIN;
     }
 
+    if (shmem_internal_params.OFI_STX_DISABLE_PRIVATE) {
+        shmem_transport_ofi_disable_private = 1;
+    } else {
+        shmem_transport_ofi_disable_private = 0;
+    }
```
jdinan: Unless the OFI transport layer overrides the param set by the user, it's preferred to use the param directly rather than duplicating it into `shmem_transport_ofi_disable_private`.  Can you just check `shmem_internal_params.OFI_STX_DISABLE_PRIVATE` within `shmem_transport_ofi_is_private` and remove the second arg to this function?

README
```diff
@@ -231,6 +231,12 @@ options.
        available for private use (i.e., with contexts that enable the
        SHMEM_CTX_PRIVATE option).
 
+    SHMEM_OFI_STX_DISABLE_PRIVATE (default: 0)
+       Disallow private contexts from having exclusive STX access.  Setting this
+       to 1 may improve performance for applications using many private
+       contexts, especially if the number of threads that use private contexts
+       is greater than SHMEM_OFI_STX_MAX.
```
jdinan: Default should be "off".  For boolean env vars, setting them to _any_ value (including the empty string) enables them.  In terms of verbiage, here's a suggestion:

Disable STX privatization.  Enabling this may improve load balance across transmit resources, especially in scenarios where the number of contexts exceeds the number of STXs.

src/transport_ofi.c
davidozog: 👍

README
davidozog: 👍

PR #641, opened 2018-03-15
Use constant number of threads in web

jdinan:


jdinan:
Looks like this edit is working -- @davidozog could you check my work and make sure I didn't break this test?  Otherwise, this should be ready to merge (and should also improve Travis testing throughput, since it reduce the number of threads per PE in the web test from 8 to 4).

PR #642, opened 2018-03-20
Remove NB fence prototype from Portals transport

jdinan:
NB fence is an interesting optimization that automatically delays fence operations to improve overlap.  However, the implementation has poor threading performance, is unmaintained, and is poorly tested.  Users can also achieve this optimization at the application layer (and arguably, this is preferred anyway since the optimization becomes portable across SHMEM libraries) by delaying fences as follows:
```
shmem_put(...);
shmem_fence();
compute(...);
shmem_put(...);
```
Into this:
```
shmem_put(...);
compute(...);
shmem_fence();
shmem_put(...);
```

### DIFF
--- a/src/transport_portals4.h
+++ b/src/transport_portals4.h
@@ -105,10 +105,6 @@ extern int shmem_transport_portals4_total_data_ordering;
 extern int shmem_transport_portals4_long_pending;
 #endif
 
-#ifdef ENABLE_NONBLOCKING_FENCE
-extern int shmem_transport_portals4_fence_pending;
-#endif
-
 #ifdef ENABLE_THREADS
 extern shmem_internal_mutex_t shmem_internal_mutex_ptl4_ctx;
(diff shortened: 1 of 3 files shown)


jdinan:
@wrrobin This cleared Travis build testing and testing on the local IB cluster.  Ready for your review!

src/transport_portals4.h
```diff
@@ -105,10 +105,6 @@ extern int shmem_transport_portals4_total_data_ordering;
 extern int shmem_transport_portals4_long_pending;
 #endif
 
-#ifdef ENABLE_NONBLOCKING_FENCE
-extern int shmem_transport_portals4_fence_pending;
-#endif
-
 #ifdef ENABLE_THREADS
```
wrrobin: We can safely remove the mutex for nb fence as well (```shmem_internal_mutex_ptl4_nb_fence```).

PR #651, opened 2018-03-21
NIC memory flush

wrrobin:


lanl-ompi:

jdinan:
@lanl-ompi ok to test

jdinan:
@lanl-ompi add to whitelist

jdinan:
@hppritcha Might need your help getting @lanl-ompi to cooperate.  🆘

wrrobin:
@jdinan This is ready for review.

hppritcha:
okay should be suffice to say

okay to test

hppritcha:
add to whitelist

src/shmem_synchronization.h
```diff
@@ -42,6 +43,7 @@ shmem_internal_fence(shmem_ctx_t ctx)
     if (0 != ret) { RAISE_ERROR(ret); }
 
     shmem_internal_membar_store();
+    shmem_transport_syncmem();
```
jdinan: I think this call may not be needed since fence doesn't provide any guarantees about what's visible in memory.

src/shmem_synchronization.h
```diff
@@ -30,6 +30,7 @@ shmem_internal_quiet(shmem_ctx_t ctx)
     if (0 != ret) { RAISE_ERROR(ret); }
 
     shmem_internal_membar();
+    shmem_transport_syncmem();
```
jdinan: Where possible, comments capturing rationale may be helpful.  In this case the syncmem is needed to make visible (i.e. to subsequent coherent load operations performed either by the local PE or another PE on the same node via the `shmem_ptr` API) the result of atomics that targeted the local process and were performed through the transport layer API.

jdinan: I think the rationale is the same for set/test lock, except that it should also provide visibility of operations performed within a previous critical section.

wrrobin: Thanks @jdinan. Can you please review the comments I have added based on your suggestions? Let me know if these look ok.

Issue #552, opened 2017-12-20
Evenly divide STXs into the PEs on each node


davidozog:
We should change the `SHMEM_OFI_STX_MAX` environment variable to `SHMEM_OFI_STX_PER_NODE`; ideally detecting the number of PEs/node, and evenly dividing STXs into PEs.

davidozog:
See PR #653

jdinan:
@davidozog Would it be reasonable to pull this feature into 1.4.2?

davidozog:
@jdinan Yes, I think so.

davidozog:
Note: #653 moved to #828

jdinan:
Closerd by #828

PR #653, opened 2018-03-27
Auto partition STXs with "shmem_node" utility

davidozog:
Addresses #552 by introducing the following features:

- If `SHMEM_OFI_STX_AUTO` is set, then detect the number of available HFI TX resources via `tx_ctx_cnt`, and partition these evenly across PEs that share the same hostname.
- Use `SHMEM_OFI_STX_NODE_MAX` with `SHMEM_OFI_STX_AUTO` to override reading the `tx_ctx_cnt` value.  
- Change `decode()` in `pmi-runtime.c` to allow for reading values from the KVS with length longer than what was written, which is useful when storing/loading hostnames.
- Simplify the hostname length variables with `SHMEM_INTERNAL_MAX_HOSTNAME_LEN`
- Avoid doing hostname storage into the PMI KVS twice if `STX_AUTO` and `ON_NODE_COMMS` are both enabled.
- Add `shmem_util_node` routines to help modularize the node-local discovery code.


src/runtime-pmi.c
```diff
@@ -77,11 +77,13 @@ decode(const char *inval, void *outval, int outvallen)
     int i;
     char *ret = (char*) outval;
 
-    if (outvallen != strlen(inval) / 2) {
+    if (outvallen < strlen(inval) / 2) {
         return 1;
     }
 
-    for (i = 0 ; i < outvallen ; ++i) {
+    size_t outlen = strlen(inval) / 2;
```
jdinan: Could you move this up and change the conditional to `if (outvallen < outlen)`?  Also, I'm assuming that the NULL termination covered during this horrifying hex->ASCII conversion?

src/transport_ofi.c
```diff
@@ -1064,18 +1051,18 @@ int populate_av(void)
...
-            SHMEM_SET_RANK_SAME_NODE(i, num_on_node++);
-            if (num_on_node > 255) {
-                RAISE_WARN_STR("Number of local ranks exceeds limit of 255");
-                return 1;
-            }
+        num_on_node = shmem_node_util_count_local_pes();
+        if (num_on_node <= 0) {
+            RAISE_WARN_STR("Failed to find any node-local PEs");
+            return 1;
+        } else if (num_on_node > 255) {
```
jdinan: It would be good to create a constant for this instead of hard coding `255`.

src/shmem_node_util.h
```diff
@@ -0,0 +1,54 @@
...
+#include <limits.h>
+#include <sys/param.h>
+
+#ifdef MAXHOSTNAMELEN
+#define SHMEM_INTERNAL_MAX_HOSTNAME_LEN MAXHOSTNAMELEN
+#else
+#define SHMEM_INTERNAL_MAX_HOSTNAME_LEN HOST_NAME_MAX
+#endif
+
+extern char *shmem_internal_location_array;
```
jdinan: Given the the usage model is signed (0/-1), it would be safer for the type of this array to be `signed char`.  From the C99 spec:
> If the value of an object of type char is treated as a signed integer when used in an
expression, the value of CHAR_MIN shall be the same as that of SCHAR_MIN and the
value of CHAR_MAX shall be the same as that of SCHAR_MAX. Otherwise, the value of
CHAR_MIN shall be 0 and the value of CHAR_MAX shall be the same as that of
UCHAR_MAX.

src/shmem_node_util.h
```diff
@@ -0,0 +1,54 @@
...
+
+int shmem_node_util_is_initialized(void);
+
+int shmem_node_util_init(void);
+
+int shmem_node_util_get_rank_same_node(int pe);
+
+void shmem_node_util_set_node_rank(int pe, int node_rank);
+
+int shmem_node_util_pe_on_same_node(int pe);
```
jdinan: Do these functions duplicate the functionality of the macros above?  If so, it would be better to have one API.

src/shmem_node_util.c
```diff
@@ -0,0 +1,92 @@
...
+char *shmem_internal_location_array = NULL;
+
+static int node_util_is_initialized = 0;
+
+
+int shmem_node_util_init(void) {
+    int ret;
+
+    if (!node_util_is_initialized) {
+        shmem_internal_location_array = malloc(sizeof(char) * shmem_internal_num_pes);
```
jdinan: Leaked during finalize?

src/shmem_node_util.c
```diff
@@ -0,0 +1,92 @@
...
+        if (NULL == shmem_internal_location_array) {
+            RAISE_ERROR_STR("Error: out of memory when allocating node_util location array");
+            return 1;
```
jdinan: The return is unreachable since `RAISE_ERROR_*` will abort.  Do you want `RETURN_ERROR_*` instead?  If not, should this routine be define to return a value?

src/shmem_node_util.c
```diff
@@ -0,0 +1,92 @@
...
+inline
+int shmem_node_util_get_rank_same_node(int pe) {
+
+#ifdef USE_ON_NODE_COMMS
+    return shmem_internal_location_array[pe];
+#elif defined(USE_MEMCPY)
+    return pe == shmem_internal_my_pe ? 0 : -1;
+#else
+    return -1;
+#endif
```
jdinan: Copy/paste from `GET_RANK_SAME_NODE` == 😿

src/shmem_node_util.c
```diff
@@ -0,0 +1,92 @@
...
+int shmem_node_util_pe_on_same_node(int pe) {
+    char nodename[SHMEM_INTERNAL_MAX_HOSTNAME_LEN];
+
+    shmem_runtime_get(pe, "nodename", nodename, SHMEM_INTERNAL_MAX_HOSTNAME_LEN);
+    if (strncmp(shmem_internal_nodename(), nodename, strlen(shmem_internal_nodename())) == 0) {
+        return 1;
+    } else {
+        return 0;
+    }
+}
```
jdinan: This function is never used?

src/shmem_node_util.c
```diff
@@ -0,0 +1,92 @@
...
+    int ret;
+
+    if (!node_util_is_initialized) {
+        shmem_internal_location_array = malloc(sizeof(char) * shmem_internal_num_pes);
+        if (NULL == shmem_internal_location_array) {
+            RAISE_ERROR_STR("Error: out of memory when allocating node_util location array");
+            return 1;
+        }
+
+        memset(shmem_internal_location_array, -1, shmem_internal_num_pes);
```
jdinan: Where is the array actually set?

src/shmem_node_util.c
```diff
@@ -0,0 +1,92 @@
...
+        if (strncmp(shmem_internal_nodename(), nodename, strlen(shmem_internal_nodename())) == 0) {
+            shmem_node_util_set_node_rank(i, num_on_node++);
+#ifdef USE_ON_NODE_COMMS
+            if (num_on_node > 255) {
+                RAISE_WARN_STR("Number of local ranks exceeds limit of 255");
+                return -1;
+            }
+#endif
+        }
+    }
```
jdinan: I would have expected this code to iterate over the location array.  Is there some reason why we would need to iterate over the KVS (much slower) instead?

src/shmem_node_util.c
```diff
@@ -0,0 +1,92 @@
...
+
+    return 0;
+}
+
+
+inline
+void shmem_node_util_set_node_rank(int pe, int node_rank) {
+
+    shmem_internal_location_array[pe] = node_rank;
```
jdinan: Is this implicitly converting `node_rank` from `int` to `char`?  Seems very bad.

src/shmem_node_util.c
jdinan: Oh, is this also the function that initializes the location array?  If so, it would be better to split this into two parts: one to populate the array and another to query the number of local PEs.

src/transport_ofi.c
```diff
@@ -1354,32 +1347,39 @@ static int shmem_transport_ofi_ctx_init(shmem_transport_ctx_t *ctx, int id)
         shmem_transport_ofi_is_private(ctx->options)) {
             ctx->tid = shmem_transport_ofi_gettid();
     }
-    shmem_transport_ofi_stx_allocate(ctx);
 
-    ret = bind_enable_cntr_ep_resources(ctx);
-    OFI_CHECK_RETURN_MSG(ret, "context bind/enable CNTR endpoint failed (%s)\n", fi_strerror(errno));
+    /* STX allocation deferred to shmem_transport_startup for default context */
+    if (ctx->id != SHMEM_TRANSPORT_CTX_DEFAULT_ID) {
```
jdinan: Can we delay creating the default context to startup instead of adding this special case?

src/shmem_node_util.c
davidozog: Yikes, that does seem bad.  This was inherited from the [macro above](https://github.com/Sandia-OpenSHMEM/SOS/pull/653/files#diff-599d81d6db092bca8aafa073bfd4f93bL32).

src/shmem_node_util.c
davidozog: Haha, sad kitty indeed.  I hesitated to use this function instead of the macro, because it's called in all the `shmem_internal` RMAs and AMOs in `shmem_comm.h`, on the critical path.  Do you think it's fine to call this inlined version throughout?  I'd guess it would be fine, but I'm glad it's come up in code review cause I'm not 100% sure.

src/shmem_node_util.c
davidozog: Whoops, yup.  Looks like it has always leaked...

src/transport_ofi.c
```diff
@@ -1500,6 +1488,83 @@ int shmem_transport_init(void)
...
     int ret;
+    int i;
+    int  num_on_node = 0;
+
+    if (shmem_internal_params.OFI_STX_AUTO) {
+
+        num_on_node = shmem_node_util_count_local_pes();
+        if (num_on_node <= 0) {
+            RAISE_ERROR_STR("Failed to find any node-local PEs");
+            return 1;
```
davidozog: Looks like another instance of the "return 1" not actually happening?

src/shmem_node_util.c
jdinan: Yeah, this seems like the source of the 255 constant, although 255 is wrong since we set entries to -1 (also 255) for PE's not on the same node.  It would be much safer for `shmem_internal_location_array` to have the same type (`int`) as a regular SHMEM PE.  Or I suspect this will bite us in the future.

src/shmem_node_util.c
jdinan: Sure, just move it to the header so it's inlined.

src/transport_ofi.c
jdinan: Yeah, either we should return on error or abort.  Currently the function signature indicates it will return, but instead aborts.

davidozog:
Sorry about all the commits - I kept running into testing failures when unifying the node_util API.  I can squash the commits if that's preferable.

Another thing I just noticed is that the portals4 code populates the node-local PE array using a `shmem_runtime_put/get` of `PtlGetPhysId`.  While that's certainly interesting, it's really not necessary with this latest rendition of the node_util.  Any thoughts on removing all that `shmem_runtime_put/get` and "USE_ON_NODE_COMMS" code from the portals4 transport?

src/shmem_node_util.c
```diff
@@ -0,0 +1,97 @@
...
+static char my_hostname[SHMEM_INTERNAL_MAX_HOSTNAME_LEN];
+static int n_local_pes = 0;
+
+int shmem_node_util_init(void)
+{
+    int ret;
+    char errmsg[256];
+
+    if (!node_util_is_initialized) {
+        if (gethostname(my_hostname, SHMEM_INTERNAL_MAX_HOSTNAME_LEN)) {
```
jdinan: I think what we might want to do is replace `gethostname` with `shmem_transport_gethostname` and then allow the transport layer to supply this info.  That should allow us to build a single table of hostnames in the PMI KVS (e.g. reuse the addresses we gather to build the AV/Portals map).

jdinan: The `gethostname` code would move to transport-none.  Hopefully for Portals and OFI we can use the addresses they already gather to populate address maps.

src/shmem_node_util.c
```diff
@@ -0,0 +1,97 @@
...
+            RETURN_ERROR_MSG("gethostname failed '%s'", shmem_util_strerror(errno, errmsg, 256));
+            return errno;
+        }
+
+        shmem_internal_location_array = malloc(sizeof(int) * shmem_internal_num_pes);
+        if (NULL == shmem_internal_location_array) {
+            RETURN_ERROR_STR("Error: out of memory when allocating node_util location array");
```
jdinan: No need to include "Error" in the message; the macro does this for you.

src/shmem_node_util.h
```diff
@@ -0,0 +1,63 @@
...
+#include <limits.h>
+#include <sys/param.h>
+
+#ifdef MAXHOSTNAMELEN
+#define SHMEM_INTERNAL_MAX_HOSTNAME_LEN MAXHOSTNAMELEN
+#else
+#define SHMEM_INTERNAL_MAX_HOSTNAME_LEN HOST_NAME_MAX
+#endif
+
+#define SHMEM_INTERNAL_MAX_NPES_PER_NODE 255
```
jdinan: Should this now be `sizeof(int)`?

src/shmem_node_util.c
```diff
@@ -0,0 +1,97 @@
...
+    int ret, i;
+    char nodename[SHMEM_INTERNAL_MAX_HOSTNAME_LEN];
+
+    for (i = 0; i < shmem_internal_num_pes; i++) {
+        ret = shmem_runtime_get(i, "nodename", nodename, SHMEM_INTERNAL_MAX_HOSTNAME_LEN);
+        if (ret != 0) {
+            RETURN_ERROR_MSG("shmem_node_util_startup failed during nodename read from KVS: (%d)", ret);
+            return ret;
+        }
+        if (strncmp(shmem_node_util_nodename(), nodename, strlen(shmem_node_util_nodename())) == 0) {
```
jdinan: Using `strlen` to get the length to pass to `strncmp` defeats the purpose of using the `n` variant of this function.  I think you mean `SHMEM_INTERNAL_MAX_HOSTNAME_LEN`.  :neckbeard:

src/shmem_node_util.c
```diff
@@ -0,0 +1,97 @@
...
+            shmem_node_util_set_local_rank(i, n_local_pes++);
+        }
+    }
+#ifdef USE_ON_NODE_COMMS
```
jdinan: This `ifdef` seems out of place here.  Don't we always want to check that this limit is not exceeded?  Also, should the limit no longer be 255 since we're upgraded the array from `char` to `int`?

src/transport_ofi.c
```diff
@@ -1500,6 +1463,60 @@ int shmem_transport_init(void)
...
+    int i;
+    int  num_on_node = 0;
+
+    if (shmem_internal_params.OFI_STX_AUTO) {
+
+        ret = shmem_node_util_startup();
+        if (ret != 0) return ret;
+
+        num_on_node = shmem_node_util_n_local_pes();
+        if (num_on_node <= 0) {
```
jdinan: This check appears to be redundant, since this is already checked in `shmem_node_util_startup`.

src/shmem_node_util.h
jdinan: Do these constants need to be exposed in the header, or can the be moved to the .c file?

src/transport_ofi.c
```diff
@@ -1500,6 +1463,60 @@ int shmem_transport_init(void)
...
+        OFI_CHECK_RETURN_MSG(ret, "STX context creation failed (%s)\n", fi_strerror(ret));
+        shmem_transport_ofi_stx_pool[i].ref_cnt = 0;
+        shmem_transport_ofi_stx_pool[i].is_private = 0;
+    }
+
+    /* Allocate STX from the pool for the default context */
+    if (shmem_internal_thread_level > SHMEM_THREAD_FUNNELED &&
+        shmem_transport_ofi_is_private(shmem_transport_ctx_default.options)) {
+            shmem_transport_ctx_default.tid = shmem_transport_ofi_gettid();
+    }
```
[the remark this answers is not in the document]
davidozog: Whoops!  Also, this check is done in `shmem_transport_ofi_ctx_init` anyway, so I'm not sure why this is here.

src/shmem_node_util.c
davidozog: Yeah, I see a few in `transport_ofi.c` as well that I'll fix.

src/transport_ofi.c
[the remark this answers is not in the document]
jdinan: I'm no opposed to turning this into a `shmem_internal_assert` if you want to keep the check.

src/shmem_node_util.c
```diff
@@ -0,0 +1,97 @@
...
+static int node_util_is_initialized = 0;
+static char my_hostname[SHMEM_INTERNAL_MAX_HOSTNAME_LEN];
+static int n_local_pes = 0;
+
+int shmem_node_util_init(void)
+{
+    int ret;
+    char errmsg[256];
+
+    if (!node_util_is_initialized) {
```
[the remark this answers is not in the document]
davidozog: I allow init to be called multiple times, for example, because `STX_AUTO` and `ON_NODE_COMMS` both require init, but init only needs to be called once.  Alternatively, I could  first check `if (!shmem_node_util_is_initialized())` and throw an error if init is called more than once like you describe.  Is that what you prefer?

I think startup should probably have a similar behavior, right?  I must have missed that.

src/shmem_node_util.c
davidozog: This check was in the original code, and I'm not 100% positive the 255 limit corresponds to the `char` size - I was thinking there could be something about CMA or XPMEM that imposes that limit...  I have no problem removing the `ifdef` though if you think it's fine.

src/shmem_node_util.h
davidozog: Do you mean `2**sizeof(int)`?  I would hope we could run more than 4 PEs per node... ;)  But then again, I don't think 4294967295 would ever work in practice.

Yes, I think that constant can go in the .c file, but I'm not sure if `MAX_HOSTNAME_LEN` can now that we have the `shmem_transport_addr_t` type which may consist of a string this length.

src/shmem_node_util.c
davidozog: Ha!  Yeah, looks like that originates from [here](https://github.com/Sandia-OpenSHMEM/SOS/blob/master/src/transport_ofi.c#L1070).  I won't tell you what git blame says about that... It appears to be from a whitespace cleanup though  ;p

src/shmem_node_util.c
jdinan: Hmm ... there could be a limit based on the number of open files or some other resource.  I suggest filing an issue to look into this.  We can go back and add a limit check, if needed.

davidozog:
Yes, I believe so.  Looking back, it is not as fresh as it once was, but a review might help revive it.

src/transport_cma.c
```diff
@@ -62,24 +63,13 @@ shmem_transport_cma_startup(void)
     int i, ret, peer_num, num_on_node = 0;
     pmi_cma_data_t cma_data;
 
-    for (i = 0 ; i < shmem_internal_num_pes; ++i) {
-        if (-1 != SHMEM_GET_RANK_SAME_NODE(i)) {
-            num_on_node++;
-        }
-    }
```
jdinan: I think you need to replace this with:
```
num_on_node = shmem_node_util_n_local_pes;
```

jdinan: Also, remove the (now) redundant initialization to 0 above.

src/transport_cma.c
```diff
@@ -62,24 +63,13 @@ shmem_transport_cma_startup(void)
...
-
-    if (num_on_node > 255) {
-        RETURN_ERROR_STR("Too many local ranks for CMA transport");
```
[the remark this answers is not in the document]
jdinan: Maybe we can try running more than 255 processes on a Neptune node to see what happens?

src/transport_none.h
```diff
@@ -87,6 +89,14 @@ shmem_transport_probe(void)
     return;
 }
 
+static inline
+shmem_transport_addr_t shmem_transport_get_local_addr(void)
+{
+    shmem_transport_addr_t addr = {0};
```
jdinan: This will cause all PEs to appear to be on the same node, which might be wrong.  A better solution might be to define the addr type as:
```
typedef int shmem_transport_addr_t;
```
So, here you can answer this query as:
```
return shmem_internal_my_pe;
```
This says all PEs are one different nodes.  Although, you could imagine a user wanting only shared memory and not being interested in the network, in which case, you could return the Unix hostname here.  That might be better, if it's not too hard to implement.

src/transport_ofi.c
```diff
@@ -555,6 +553,19 @@ void shmem_transport_ofi_stx_allocate(shmem_transport_ctx_t *ctx)
     return;
 }
 
+
+shmem_transport_addr_t shmem_transport_get_local_addr(void)
+{
+    shmem_transport_addr_t addr;
+
+    void *addr_ptr = shmem_transport_ofi_info.fabrics->src_addr;
```
jdinan: OFI says "a provider shall return an address that corresponds to the indicated fabric, domain, node, and/or service fields".  If this can correspond to the specific domain, then I'm not sure it will serve as a node name.

@sayantansur Is this the right way to get a node name from OFI?

src/transport_ofi.c
```diff
@@ -1440,6 +1433,9 @@ int shmem_transport_init(void)
                            shmem_internal_params.OFI_STX_MAX);
         }
         shmem_transport_ofi_stx_max = 1;
+    } else if (shmem_internal_params.OFI_STX_AUTO) {
+        ret = shmem_node_util_init();
```
jdinan: Should node utils be initialized by the transport or as part of the top-level init?  The latter gives cleaner separation between these components.

src/transport_ofi.c
```diff
@@ -1490,18 +1472,12 @@ int shmem_transport_init(void)
...
-    if (ret != 0) return ret;
-
     ret = shmem_transport_ofi_target_ep_init();
     if (ret != 0) return ret;
 
     ret = publish_mr_info();
     if (ret != 0) return ret;
 
-    ret = atomic_limitations_check();
-    if (ret != 0) return ret;
```
jdinan: Why was this moved from init to startup?

src/transport_ofi.c
```diff
@@ -1511,6 +1487,54 @@ int shmem_transport_init(void)
 int shmem_transport_startup(void)
 {
     int ret;
+    int i;
+    int  num_on_node = 0;
+
+    if (shmem_internal_params.OFI_STX_AUTO) {
+
+        ret = shmem_node_util_startup();
+        if (ret != 0) return ret;
```
jdinan: Same comment as earlier, I think I would prefer to startup the node at the top level before transport startup.

src/transport_ofi.c
```diff
@@ -1260,6 +1246,12 @@ int query_for_fabric(struct fabric_info *info)
     shmem_transport_ofi_mr_rma_event = (info->p_info->domain_attr->mr_mode & FI_MR_RMA_EVENT) != 0;
 #endif
 
+    if (shmem_internal_params.OFI_STX_NODE_MAX == -1) {
+        shmem_transport_ofi_tx_ctx_cnt = info->fabrics->domain_attr->tx_ctx_cnt;
+    } else {
+        shmem_transport_ofi_tx_ctx_cnt = shmem_internal_params.OFI_STX_NODE_MAX;
+    }
```
jdinan: Can we move this down to transport_init to avoid carrying the tx_ctx_cnt state across functions?

src/transport_ofi.c
jdinan: Oh, I see, it needs the default context.

src/transport_portals4.h
```diff
@@ -150,6 +150,11 @@ struct shmem_transport_ct_t {
 };
 typedef struct shmem_transport_ct_t shmem_transport_ct_t;
 
+struct shmem_transport_addr_t {
+    ptl_process_t ptl_addr;
+};
+typedef struct shmem_transport_addr_t shmem_transport_addr_t;
```
jdinan: Why wrap a struct around `ptl_process_t`?

src/transport_portals4.h
```diff
@@ -269,6 +274,18 @@ static inline void shmem_transport_probe(void) {
...
+static inline
+shmem_transport_addr_t shmem_transport_get_local_addr(void)
+{
+    shmem_transport_addr_t addr;
+    int ret = PtlGetPhysId(shmem_transport_portals4_ni_h, &addr.ptl_addr);
+    if (PTL_OK != ret) {
+        RAISE_ERROR_MSG("PtlGetPhysId failed: %d\n", ret);
+    }
+    return addr;
+}
```
jdinan: The physical ID isn't quite enough to determine which processes share a node.  You need to look at the `nid` in:
```
typedef union {
  struct {
    ptl_nid_t nid;
    ptl_pid_t pid;
  } phys;
  ptl_rank_t rank;
} ptl_process_t;
```
However, you do want to store the whole `ptl_process_t` to avoid putting duplicate info into the KVS.  I think this can be remedied by adding a `shmem_transport_same_node(const shmem_transport_addr_t *a1, const shmem_transport_addr_t *a2)` comparison operation to the transport interface.

jdinan: See line 558 in transport_portals4.c for the nid usage.

src/transport_portals4.c
```diff
@@ -536,41 +535,29 @@ shmem_transport_startup(void)
...
-    ret = PtlGetPhysId(shmem_transport_portals4_ni_h, &my_id);
-    if (PTL_OK != ret) {
-        RETURN_ERROR_MSG("PtlGetPhysId failed: %d\n", ret);
-        goto cleanup;
-    }
-
     for (i = 0 ; i < shmem_internal_num_pes; ++i) {
+#ifdef USE_ON_NODE_COMMS
+        shmem_transport_addr_t addr = shmem_node_util_get_addr(i);
+        memcpy(&pe_map[i], &addr.ptl_addr, sizeof(ptl_process_t));
```
jdinan: I think you can get rid of this memcpy if you have `shmem_node_util_get_addr` take a pointer to the buffer where you want to store the addr.

src/shmem_node_util.c
```diff
@@ -0,0 +1,117 @@
...
+        ret = shmem_runtime_put("nodename", &addr, sizeof(addr));
+        if (ret !=0) {
+            RETURN_ERROR_MSG("shmem_node_util_init failed during nodename store to KVS: (%d)", ret);
+            return ret;
+        }
+
+        node_util_is_initialized = 1;
+    } else {
+        if (shmem_internal_params.DEBUG) {
+            DEBUG_STR("Initialized shmem_node_util more than once");
```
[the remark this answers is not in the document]
davidozog: There are some instances where `shmem_node_util_init()` would be called more than once; for example, when using XPMEM and setting `OFI_STX_AUTO` (or any `ON_NODE_COMMS` + `OFI_STX_AUTO`).

src/shmem_node_util.c
```diff
@@ -0,0 +1,117 @@
...
+            return 1;
+        } else if (n_local_pes <= 0) {
+            RETURN_ERROR_STR("Failed to find any node-local PEs");
+            return 1;
+        }
+#endif
+        node_util_is_started = 1;
+    } else {
+        if (shmem_internal_params.DEBUG) {
+            DEBUG_STR("Called shmem_node_util_startup more than once");
```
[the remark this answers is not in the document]
davidozog: This might also occur with `USE_ON_NODE_COMMS` and `OFI_STX_AUTO`.  

Side note: this way (allowing multiple initializations) seems preferable to me than cluttering the transport code with checks for `USE_ON_NODE_COMMS`, but that could also work.

src/shmem_node_util.c
```diff
@@ -0,0 +1,117 @@
...
+                RETURN_ERROR_MSG("shmem_node_util_startup failed during nodename read from KVS: (%d)", ret);
+                return ret;
+            }
+            if (memcmp(&addr, &addr_cmp, sizeof(shmem_transport_addr_t)) == 0) {
+                shmem_node_util_set_local_rank(i, n_local_pes++);
+            }
+        }
+#ifdef USE_ON_NODE_COMMS
+        if (n_local_pes > SHMEM_INTERNAL_MAX_NPES_PER_NODE) {
+            RETURN_ERROR_MSG("Number of local ranks exceeds limit of %d", SHMEM_INTERNAL_MAX_NPES_PER_NODE);
```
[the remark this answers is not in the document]
davidozog: Not that I know of.  I'm honestly not sure where it comes from, but I was tentatively planning on looking into whether it's valid (empirically).  I'm fine removing it though.

src/transport_cma.c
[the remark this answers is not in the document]
davidozog: Yeah, I can run well over 255 processes with OpenMPI configured to use CMA.

src/transport_ofi.c
davidozog: Do you mean move this check for `OFI_STX_AUTO` to `init.c`?  I'd think we'd prefer to keep OFI specific parameters out of the top-level init.  It might be nice to have `USE_ON_NODE_COMMS` capture `OFI_STX_AUTO`, but since it's set at configure time, I'm not sure how to do that.

src/transport_ofi.c
davidozog: I don't quite understand.  I believe this is inside transport_init right now (`query_for_fabric`).  Do you mean move it to `transport_startup`?  If so, I think that would be fine.

src/transport_ofi.c
jdinan: You could add a query function like `shmem_transport_needs_node_utils()`

src/transport_ofi.c
jdinan: Yes, that's what I meant.  :)  Move resolving this parameter down to where it's consumed, so we don't need to carry the extra state between functions.

davidozog:
@jdinan, this round of changes is ready for review.

You'll see that I ran into a problem with `MAXHOSTNAMELEN` in  OFI's `shmem_transport_addr_t` being too large to support the PMI-1 limit on value size in the KVS.  Now I use an address size (`addrlen`) and corresponding `void *` in the struct, with 2 calls to `runtime_put` and `runtime_get` to stay under the limit...  I'm certainly open to a cleaner fix if you can think of one.

davidozog:
My apologies on those last few commits - I saw the tests were failing, which gradually led to me cleaning things up a bit.

PR #656, opened 2018-03-28
Header files for threaded perf tests and updated tarball name in spec

wrrobin:


wrrobin:
Thanks @jdinan. Done.

PR #658, opened 2018-03-29
Add multithreaded lock unit test

jdinan:
Supersedes #616


jdinan:
@davidozog All the same code you reviewed previously, now in a shiny new form factor.

PR #659, opened 2018-03-29
Travis check for distribution tarball and RPM build

wrrobin:


jdinan:
Cool!  Rather than adding two rows to the testing matrix, can we do both in the same row?

.travis.yml
```diff
@@ -383,6 +386,23 @@ script:
...
+          cd $SOS_SRC
+          ./configure --with-ofi=$TRAVIS_INSTALL/libfabric/ --enable-pmi-simple
+          make dist
+          rpmbuild -ta ./sos-*.tar.gz --define "configargs --with-ofi=$TRAVIS_INSTALL/libfabric/ --enable-pmi-simple"
+          tar zxvf sos-*.tar.gz
+          cd sos-*
+          ./autogen.sh
+          mkdir build
+          cd build
+          ../configure --with-ofi=$TRAVIS_INSTALL/libfabric/ --enable-pmi-simple
```
jdinan: `../configure --with-ofi=$TRAVIS_INSTALL/libfabric/ $SOS_GLOBAL_BUILD_OPTS $SOS_BUILD_OPTS`

.travis.yml
```diff
@@ -383,6 +386,23 @@ script:
...
+          make check
```
jdinan:
```
make $TRAVIS_PAR_MAKE check TESTS=
make VERBOSE=1 TEST_RUNNER="mpiexec.hydra -np 2" check
```

jdinan: This will shorten the build time.

.travis.yml
wrrobin: Oh. I thought those are the only two flags that we want to be tested and similar to what is expected during IFS release. Thanks, will update it.

PR #666, opened 2018-04-04
Hostname check and optional individual bandwidth for throughput perf tests

wrrobin:


test/performance/shmem_perf_suite/bw_common.h
```diff
@@ -541,10 +547,34 @@ void static inline calc_and_print_results(double total_t, int len,
...
+    pe_time = total_t;
+    shmem_barrier(start_pe, stride, nPEs, red_psync);
+    if (nPEs >= 2) {
+        shmem_double_max_to_all(&pe_time_max, &pe_time, nred_elements, start_pe,
                                 stride, nPEs, pwrk,
                                 red_psync);
+    }
+
+    /* calculating bandwidth based on the highest time taken across all PEs */
+    if (pe_time_max > 0 ) {
```
jdinan: What should we do if `pe_time_max` is not greater than 0?

test/performance/shmem_perf_suite/bw_common.h
```diff
@@ -541,10 +547,34 @@ void static inline calc_and_print_results(double total_t, int len,
...
+                (double)metric_info.nthreads) / (pe_time_max / 1.0e6);
+#else
+        bw = (len * metric_info.midpt / 1.0e6 * metric_info.window_size * metric_info.trials) /
+                (pe_time_max / 1.0e6);
+#endif
+    }
+
+    /* 2x as many messages/bytes at once for bi-directional */
+    if(metric_info.type == BI_DIR)
+        bw *= 2.0;
```
jdinan: Would it be better to multiply `len` by 2.0 above since it's twice the data volume for bi-directional?  This should yield the same result, but may be easier to understand.

test/performance/shmem_perf_suite/bw_common.h
```diff
@@ -781,3 +811,58 @@ void static inline uni_dir_bw_main(int argc, char *argv[], bw_style bwstyl) {
...
+    int hostname_status = -1;
+    int hostname_size = (HOST_NAME_MAX % 4 == 0) ? HOST_NAME_MAX : HOST_NAME_MAX + (4 - HOST_NAME_MAX % 4);
+    int i,j;
+
+    static long pSync_collect[SHMEM_COLLECT_SYNC_SIZE];
+    for (i = 0; i < SHMEM_COLLECT_SYNC_SIZE; i++)
+        pSync_collect[i] = SHMEM_SYNC_VALUE;
+
+    char *hostname = (char *) shmem_malloc (hostname_size * sizeof(char));
+    char *dest = (char *) shmem_malloc (my_info.num_pes * hostname_size * sizeof(char));
```
jdinan: Should these buffers be freed before this function returns?

test/performance/shmem_perf_suite/bw_common.h
```diff
@@ -781,3 +811,58 @@ void static inline uni_dir_bw_main(int argc, char *argv[], bw_style bwstyl) {
...
+
+    hostname_status = gethostname(hostname, hostname_size);
+    if (hostname_status == -1) {
+        fprintf(stderr, "gethostname returned -1. Exiting.\n");
+        return -1;
```
jdinan: Better to use `RETURN_ERROR_STR` here.  These macros will add extra debugging output including source file/line and PE id.

jdinan: Also, it would be better to check `hostname_status != 0` and print the exit code: `RETURN_ERROR_MSG("gethostname failed (%d)\n", hostname_status);`

test/performance/shmem_perf_suite/bw_common.h
```diff
@@ -781,3 +811,58 @@ void static inline uni_dir_bw_main(int argc, char *argv[], bw_style bwstyl) {
...
+    }
+    shmem_barrier_all();
+
+    shmem_fcollect32(dest, hostname, hostname_size/4, 0, 0, my_info.num_pes, pSync_collect);
+
+    char *streaming_node = malloc (hostname_size * sizeof(char));
+    char *target_node = malloc (hostname_size * sizeof(char));
```
jdinan: All of the malloc'd buffers appear to be leaked.

test/performance/shmem_perf_suite/bw_common.h
```diff
@@ -781,3 +811,58 @@ void static inline uni_dir_bw_main(int argc, char *argv[], bw_style bwstyl) {
...
+    hostname_status = gethostname(hostname, hostname_size);
+    if (hostname_status == -1) {
+        fprintf(stderr, "gethostname returned -1. Exiting.\n");
+        return -1;
+    }
+    shmem_barrier_all();
+
+    shmem_fcollect32(dest, hostname, hostname_size/4, 0, 0, my_info.num_pes, pSync_collect);
+
+    char *streaming_node = malloc (hostname_size * sizeof(char));
```
jdinan: Isn't there already a function named `streaming node`?

test/performance/shmem_perf_suite/bw_common.h
```diff
@@ -781,3 +811,58 @@ void static inline uni_dir_bw_main(int argc, char *argv[], bw_style bwstyl) {
...
+        } else if (i > my_info.midpt && i < my_info.num_pes) {
+            if (strcmp(target_node, temp) != 0) {
+                fprintf(stderr, "PE %d on %s is a target node but not placed on %s\n", my_info.my_node, temp, target_node);
+                return -1;
+            }
+        } else {
+           fprintf(stderr, "Wrong PE %d and node %s combination\n", my_info.my_node, temp);
+           return -1;
+        }
+    }
```
jdinan: The memory allocation and string copying shouldn't be necessary.  I think something like this ought to work (you will need to add the function `streaming_node_peer`):
```
int errors = 0;
char *snode_name = NULL;
char *tnode_name = NULL;
for (i = 0; i < my_info.num_pes; i++) {
    char *cur_name = &dest[i*hostname_size];

    if (streaming_node_peer(my_info, i)) {
        if (snode_name == NULL)
            snode_name = cur_name;

        if (strcmp(snode_name, cur_name) != 0) {
            fprintf(stderr, "PE %d on %s is a streaming node but not placed on %s\n",
                    i, cur_name, snode_name);
            errors++;
        }
    }
    else {
        if (tnode_name == NULL)
            tnode_name = cur_name;
        else {
            if (strcmp(tnode_name, cur_name) != 0) {
                fprintf(stderr, "PE %d on %s is a target node but not placed on %s\n",
                        i, cur_name, snode_name);
                errors++;
            }
        }
    }
}

if (strcmp(snode_name, tnode_name) == 0) {
    fprintf(stderr, "Warning: senders and receivers are running on the same node %s\n", snode_name);
    errors++;
}

return errors;
```

test/performance/shmem_perf_suite/common.h
```diff
@@ -36,6 +36,8 @@
 #include <time.h>
 #include <stdint.h>
 
+#include <bits/local_lim.h>
```
jdinan: What's this for?  IIRC including headers directly from `bits/` is bad for portability.

test/performance/shmem_perf_suite/bw_common.h
wrrobin: `bw` is already assigned 0, so in terms of calculation, there is nothing to be done. But, we can definitely add an error message like the impossible case has happened.

test/performance/shmem_perf_suite/bw_common.h
wrrobin: Actually, I had it. But then removed it because of a crash caused by multiple free of a single location. Thanks for catching this. Would check that back in.

test/performance/shmem_perf_suite/bw_common.h
wrrobin: Yes. Thanks.

test/performance/shmem_perf_suite/common.h
wrrobin: This is for `HOST_NAME_MAX`. Initially, I added `limits.h`, but changed to this one. Should I keep `limits.h` instead?

wrrobin:
@jdinan, After adding the bandwidth measurements like you suggested yesterday, I did not see any over line rate issue while using the handshaking for each comm pair. Just need to see whether we get same behavior on the user end.

test/performance/shmem_perf_suite/bw_common.h
wrrobin: The last condition outside the loop should not have `errors` incremented, right? Since this is a valid way to run.

test/performance/shmem_perf_suite/bw_common.h
wrrobin: Should we just use `fprintf` to report the error? Since `RETURN_ERROR_MSG` brings the dependency on `shmem_internal.h` and it also does not get all the internal parameters (e.g. `shmem_internal_my_pe`) automatically.

jdinan: Sorry, I forgot this is in the unit tests.  You're right, we don't want to pull in shmem_internal.h here.

test/performance/shmem_perf_suite/common.h
jdinan: I think you want something similar to the solution in #653 
```
#include <limits.h>
#include <sys/param.h>

#ifdef MAXHOSTNAMELEN
#define SHMEM_INTERNAL_MAX_HOSTNAME_LEN MAXHOSTNAMELEN
#else
#define SHMEM_INTERNAL_MAX_HOSTNAME_LEN HOST_NAME_MAX
#endif
```

test/performance/shmem_perf_suite/bw_common.h
```diff
@@ -206,6 +206,16 @@ int static inline streaming_node(perf_metrics_t my_info)
     }
 }
 
+static int inline is_streaming_node(perf_metrics_t my_info, int node)
+{
+    if(my_info.cstyle == COMM_PAIRWISE) {
+        return (node < my_info.szinitiator);
+    } else {
+        assert(my_info.cstyle == COMM_INCAST);
+        return true;
```
jdinan: Is that right?  All PEs are streaming in `COMM_INCAST`?

wrrobin: Yes, thats what I got from the function above.

test/performance/shmem_perf_suite/bi_dir.h
```diff
@@ -30,6 +30,20 @@ void static inline bi_bw_put(int len, perf_metrics_t *metric_info)
     double start = 0.0, end = 0.0;
     int dest = partner_node(*metric_info);
     int i = 0, j = 0;
+    static int check_once = 0;
+    static int fin = -1;
+
+    if (!check_once) {
+        /* check to see whether sender and receiver are the same process */
+        if (dest == metric_info->my_node) {
+            fprintf(stderr, "Sender and receiver are the same process (%d)\n", dest);
```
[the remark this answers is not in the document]
wrrobin: I treated this as an unwanted case and thus returned without running the instance. If we add a warning then we can run the benchmark after the warning is printed.

test/performance/shmem_perf_suite/bw_common.h
```diff
@@ -512,39 +528,77 @@ void static inline PE_set_used_adjustments(int *nPEs, int *stride, int *start_pe
...
     static double pwrk[SHMEM_REDUCE_MIN_WRKDATA_SIZE];
+    static double pe_time_start, pe_time_end, end_time_max = 0.0, start_time_min = 0.0;
+    double total_t = 0.0, total_t_max = 0.0;
 
     PE_set_used_adjustments(&nPEs, &stride, &start_pe, metric_info);
 
-    if (total_t > 0 ) {
+    /* 2x as many messages at once for bi-directional */
+    if(metric_info.type == BI_DIR)
+        len *= 2.0;
```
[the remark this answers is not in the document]
wrrobin: Thanks @jdinan. I have changed it to `2`. Since `len` is limited by `max_len` and that is a user input, we can add a check for invalid input there.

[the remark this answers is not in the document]
wrrobin: @jdinan Added some changes to handle this. Let me know if the new changes look ok.

wrrobin:
Thanks @jdinan for all the suggestions.

PR #668, opened 2018-04-17
Fix build warnings

jdinan:
@davidozog Requesting your review of the OFI transport changes
@wrrobin Requesting your review of perf suite changes

### DIFF
--- a/src/transport_ofi.h
+++ b/src/transport_ofi.h
@@ -32,7 +32,11 @@
 #include <sys/types.h>
 
 
-#define ENABLE_TARGET_CNTR (!defined(ENABLE_HARD_POLLING) || defined(ENABLE_MANUAL_PROGRESS))
+#if !defined(ENABLE_HARD_POLLING) || defined(ENABLE_MANUAL_PROGRESS)
+#define ENABLE_TARGET_CNTR 1
+#else
+#define ENABLE_TARGET_CNTR 0
+#endif
 
 #if ENABLE_TARGET_CNTR
 extern struct fid_cntr*                 shmem_transport_ofi_target_cntrfd;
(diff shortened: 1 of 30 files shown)


src/transport_ofi.h
```diff
@@ -381,7 +385,9 @@ shmem_transport_ofi_bounce_buffer_t * create_bounce_buffer(shmem_transport_ctx_t
 
     SHMEM_TRANSPORT_OFI_CTX_BB_LOCK(ctx);
 
-    while (ctx->bounce_buffers->nalloc >= shmem_transport_ofi_max_bounce_buffers) {
+    shmem_internal_assert(shmem_transport_ofi_max_bounce_size > 0);
```
davidozog: Looks like a typo.  Should be `shmem_transport_ofi_max_bounce_buffers`, right?

test/performance/shmem_perf_suite/latency_common.h
```diff
@@ -153,7 +153,7 @@ extern void int_element_latency(perf_metrics_t data);
 extern void streaming_latency(int len, perf_metrics_t *data);
 
 void static inline  multi_size_latency(perf_metrics_t data, char *argv[]) {
-    int len;
+    unsigned int len;
     int partner_pe = partner_node(data.my_node);
```
wrrobin: Should we make `start_len` and `max_len` in both `latency_common` and `bw_common` `unsigned long int` data type to make it unified?

jdinan: Sure, let's do this after the 1.4.1 release.  Created #669 to track.

jdinan:
If npes is not a positive integer, then something is horribly wrong.  If you want to check for this, better to do it before returning from `shmem_init`.  🤓

test/performance/shmem_perf_suite/latency_common.h
wrrobin: Ok. thanks. Other than this, the perf suite changes look good to me.

davidozog:
Yup, agreed.  It looks like we don't check for this in `shmem_init`, we just take what PMI gives us.  But, I don't think it's all that necessary.  If npes is non-positive, then the whole stack probably deserves what's coming ;)

PR #675, opened 2018-04-18
Minor fixes found during Portals testing

jdinan:


jdinan:
@davidozog FYI, I had a late bugfix to the profiling header in 9aa974d.

PR #682, opened 2018-04-19
Cleanup msgrate test

davidozog:
Call finalize before exit, remove psync addr print.


jdinan:
Tested sucessfully in @davidozog 's fork.

PR #690, opened 2018-05-04
Simple fixes for Klocwork warnings

davidozog:
Check for NULL in a couple cases and initialize some buffers.


bindings/generate_bind_cxx_extras.cpp
```diff
@@ -107,7 +108,11 @@ int main(int argc, char **argv)
     time_t curtime = time(NULL);
     struct tm *t = localtime(&curtime);
 
-    strftime(timestr, sizeof(timestr)-1, "%m-%d-%Y %H:%M", t);
+    if (t != NULL) {
+        strftime(timestr, sizeof(timestr)-1, "%m-%d-%Y %H:%M", t);
```
jdinan: The `-1` to the size should be unnecessary since `strftime` NULL terminates the string.  Also, this is silly because we're also not setting that bonus character to `\0`.

`strftime` can also fail, which we're not checking for.  I suggest just initializing `char timestr[100] = "unknown time";` since it will more easily address all the ways in which this can fail.

jdinan: Also, I realize that the `-1` silliness was totally my doing.  :neckbeard:

PR #698, opened 2018-05-17
Fix for the msg rate issue in atomics test

wrrobin:


wrrobin:
Yes @davidozog , you are right. I was under the impression that the start and stop times exhibit similar trends across both types of bandwidth tests. But, after printing out the times for AMO tests, I find big differences between start and stop times. That is the reason for very small bandwidth using max of start and min of end times. 
For now, this fix is correct. But, we can take a look at the AMO test later to find out whether we need to refine it.

test/performance/shmem_perf_suite/bw_common.h
```diff
@@ -579,46 +579,55 @@ static inline void calc_and_print_results(double end_t, double start_t, int len,
     pe_bw_sum = bw;
 
     if (metric_info.individual_report == 1) {
-        printf("Individual bandwith for PE %6d is %10.2f\n", 
-                metric_info.my_node, pe_bw_sum);
+        printf("Individual bandwith for PE %6d is %10.2f start = %lf, end = %lf\n", 
+                metric_info.my_node, pe_bw_sum, start_t, end_t);
```
jdinan: Did you mean to include this?

wrrobin: Thanks. Removed it.

wrrobin:
@kseager Waiting for your approval. Please let me know if you have any suggestions.

wrrobin:
@jdinan @kseager @davidozog This PR is ready for re-review. Included are: 
1. Fix for the bandwidth and message rate issue in uni-directional atomic test (already reviewed)
2. Merge with PR#693, resolved conflicts.
3. Minor edits to enhance validation reporting.

test/performance/shmem_perf_suite/shmem_bibw_atomics_perf.c
```diff
@@ -169,7 +169,7 @@ static inline void bw_set_metric_info_len(perf_metrics_t *metric_info)
                                         sizeof(long long)};
     int snode = streaming_node(*metric_info);
     atomic_op_type op_type = OP_ADD;
-    metric_info->type = BI_DIR;
+    metric_info->b_type = BI_DIR;
```
jdinan: Should this be set prior to the call to `streaming_node`?

jdinan: Looks like `streaming_node` doesn't check this field, but maybe a good idea to switch the order anyway?  Also, it would be good if `streaming_node` called `is_streaming_node` with `my_info.my_node` to eliminate the copy/paste.  Maybe file an issue for that?

test/performance/shmem_perf_suite/shmem_bibw_atomics_perf.c
```diff
@@ -152,7 +170,6 @@ static inline void bw_set_metric_info_len(perf_metrics_t *metric_info)
     int snode = streaming_node(*metric_info);
     atomic_op_type op_type = OP_ADD;
     metric_info->b_type = BI_DIR;
```
jdinan: In case github lost my other comment.  Should this like be moved up to before the call to `streaming_node`?

jdinan:
Looks good to me.  @kseager do you want to check the new patches?

test/performance/shmem_perf_suite/shmem_bibw_atomics_perf.c
wrrobin: Yes, `streaming_node` does not depend on this. The only metric it depends on is the `comm_style`. I will reverse the order anyway. 
Thanks for the other suggestion. Will fix it in a future PR.

wrrobin:
@kseager Please let me know if the new patches look good to you. Thanks.

kseager:
@wrrobin new patches look good to me

wrrobin:
Thanks @kseager @jdinan 
'Squash and merge'-ing this as these are small patches.

PR #701, opened 2018-05-23
Fix start_pes error check

jdinan:
OpenSHMEM specifies that "Calling start_pes more than once has no
subsequent effect."


jdinan:
Should we also apply this fix to the 1.3.x branch?

PR #702, opened 2018-05-24
Introduction of query APIs for performance counters

wrrobin:


src/transport_none.h
```diff
@@ -311,4 +311,34 @@ void shmem_transport_syncmem(void)
     return;
 }
 
+static inline
+uint64_t shmem_transport_get_pending_put_cntr(shmem_transport_ctx_t *ctx)
+{
+    return -1;
```
jdinan: These functions return an unsigned quantity, so 0 might be a better choice than -1.  In fact, 0 is a good choice since there is no transport and therefore these counters are effectively stuck at 0.

src/transport_ofi.h
```diff
@@ -1213,4 +1213,63 @@ void shmem_transport_syncmem(void)
     // transport routine will be a nop until an API is provided.
 }
 
+static inline
+uint64_t shmem_transport_get_pending_put_cntr(shmem_transport_ctx_t *ctx)
+{
+    uint64_t cnt = 0;
```
jdinan: No need to initialize this to 0 (potentially adds an instruction since this is stack-allocated, although the compiler should see that this is not read before it's written below).

src/transport_ofi.h
```diff
@@ -1213,4 +1213,63 @@ void shmem_transport_syncmem(void)
...
+#  ifdef USE_THREAD_COMPLETION
+    if (0 == pthread_mutex_lock(&shmem_transport_ofi_progress_lock)) {
+#  endif
+        cnt = fi_cntr_read(shmem_transport_ofi_target_cntrfd);
+#  ifdef USE_THREAD_COMPLETION
+        pthread_mutex_unlock(&shmem_transport_ofi_progress_lock);
+    }
+#  endif
+#else
+    cnt = -1;
```
jdinan: This should be 0, not -1 since -1 is not within the range of uint64_t.  If you want to return all bits set, better to use `UINT64_MAX`.

src/transport_portals4.h
```diff
@@ -1314,4 +1314,79 @@ void shmem_transport_syncmem(void)
...
 
+#ifdef ENABLE_THREADS
+#define SHMEM_TRANSPORT_PTL4_CNTR_LOCK(ctx) SHMEM_MUTEX_LOCK(ctx) 
+#define SHMEM_TRANSPORT_PTL4_CNTR_UNLOCK(ctx) SHMEM_MUTEX_UNLOCK(ctx)
+#define SHMEM_TRANSPORT_PTL4_CNTR_READ(cntr) *(cntr) 
+#else 
+#define SHMEM_TRANSPORT_PTL4_CNTR_LOCK(ctx) 
+#define SHMEM_TRANSPORT_PTL4_CNTR_UNLOCK(ctx) 
+#define SHMEM_TRANSPORT_PTL4_CNTR_READ(cntr) shmem_internal_atomic_read(cntr) 
+#endif
```
jdinan: There are no context locks in the Portals transport (similar to FI_THREAD_SAFE build in the OFI transport).  You should just be able to `shmem_internal_atomic_read` the pending counters and `PtlCTGet` the event counters.

src/transport_portals4.h
```diff
@@ -1314,4 +1314,79 @@ void shmem_transport_syncmem(void)
...
+
+static inline
+uint64_t shmem_transport_get_pending_put_cntr(shmem_transport_ctx_t *ctx)
+{
+    uint64_t cnt = 0;
+    SHMEM_TRANSPORT_PTL4_CNTR_LOCK(shmem_internal_mutex_ptl4_ctx);
+    cnt = SHMEM_TRANSPORT_PTL4_CNTR_READ(&shmem_transport_portals4_pending_put_event_cntr);
```
jdinan: Only add this in if `ctx->options & SHMEMX_CTX_BOUNCE_BUFFER`.

src/transport_portals4.h
```diff
@@ -1314,4 +1314,79 @@ void shmem_transport_syncmem(void)
...
+#ifndef ENABLE_HARD_POLLING
+    int ret;
+    ptl_ct_event_t ev;
+
+    ret = PtlCTGet(shmem_transport_portals4_target_ct_h, &ev);
+
+    if (PTL_OK != ret) { RAISE_ERROR(ret); }
+    return (uint64_t) ev.success;
+#else
+    return (uint64_t) -1;
```
jdinan: Return 0

src/transport_portals4.h
```diff
@@ -1314,4 +1314,79 @@ void shmem_transport_syncmem(void)
...
+static inline
+uint64_t shmem_transport_get_fi_put_cntr(shmem_transport_ctx_t *ctx)
+{
+    int ret;
+    ptl_ct_event_t ev;
+
+    SHMEM_TRANSPORT_PTL4_CNTR_LOCK(shmem_internal_mutex_ptl4_ctx);
+    ret = PtlCTGet(ctx->put_ct, &ev);
+    SHMEM_TRANSPORT_PTL4_CNTR_UNLOCK(shmem_internal_mutex_ptl4_ctx);
+    if (PTL_OK != ret) { RAISE_ERROR(ret); }
```
jdinan: Need to account for buffered puts:
```
uint64_t cnt = ev.success;
if (ctx->options & SHMEMX_CTX_BOUNCE_BUFFER) {
    ret = PtlCTGet(shmem_transport_portals4_put_event_ct_h);
    if (PTL_OK != ret) { RAISE_ERROR(ret); }
    cnt += ev.success;
}

return ret;
```

src/transport_portals4.h
```diff
@@ -1314,4 +1314,79 @@ void shmem_transport_syncmem(void)
...
+    cnt += SHMEM_TRANSPORT_PTL4_CNTR_READ(&ctx->pending_put_cntr);
+    SHMEM_TRANSPORT_PTL4_CNTR_UNLOCK(shmem_internal_mutex_ptl4_ctx);
+    return cnt;
+}
+
+static inline
+uint64_t shmem_transport_get_pending_get_cntr(shmem_transport_ctx_t *ctx)
+{
+    uint64_t cnt = 0;
+    SHMEM_TRANSPORT_PTL4_CNTR_LOCK(shmem_internal_mutex_ptl4_ctx);
```
jdinan: No need to lock around calls to PtlCTGet; it's already thread safe.  This lock object is used to guard the context tracking array, so it won't help here.

wrrobin:
Thanks @jdinan for the feedback. Incorporated these changes.

src/perf_counters_c.c
```diff
@@ -0,0 +1,78 @@
...
+    SHMEM_ERR_CHECK_INITIALIZED();
+    *cntr_value = shmem_transport_get_pending_put_cntr((shmem_transport_ctx_t *) ctx);
+    return;
+}
+
+void SHMEM_FUNCTION_ATTRIBUTES 
+shmemx_pcntr_get_pending_get(shmem_ctx_t ctx, uint64_t *cntr_value)
+{
+    SHMEM_ERR_CHECK_INITIALIZED();
+    *cntr_value = shmem_transport_get_pending_get_cntr((shmem_transport_ctx_t *) ctx);
```
jdinan: Could you rename the transport API routines to match the corresponding shmemx routines?  E.g., `shmemx_pcntr_get_pending_get` should map to `shmem_transport_pcntr_get_pending_get`.

wrrobin: Thanks @jdinan. Will do. 
Do you think the APIs for the two new counters you introduced in #704 will be handy?

wrrobin:
Thanks @jdinan. I will incorporate the BB counters and the single API for all the counters separately.

PR #704, opened 2018-05-25
Add bb performance counters

jdinan:
* Add bounce buffer performance counters
 * Add debug message to print diagnostic info upon context destruction
 * Fix type issue with SHMEM_CTX_* constants (int implicitly promoted to long)

### DIFF
--- a/mpp/shmem.h4.in
+++ b/mpp/shmem.h4.in
@@ -87,9 +87,9 @@ typedef struct { int dummy; } * shmem_ctx_t;
 #else
   extern shmem_ctx_t SHMEM_CTX_DEFAULT;
 #endif
-#define SHMEM_CTX_PRIVATE       (1<<0)
-#define SHMEM_CTX_SERIALIZED    (1<<1)
-#define SHMEM_CTX_NOSTORE       (1<<2)
+#define SHMEM_CTX_PRIVATE       (1l<<0)
+#define SHMEM_CTX_SERIALIZED    (1l<<1)
+#define SHMEM_CTX_NOSTORE       (1l<<2)
 
 define(`SHPRE', `')dnl
 include(shmem_c_func.h4)dnl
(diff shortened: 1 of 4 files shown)


jdinan:
Sample output from `test/performance/shmem_perf_suite/shmem_bw_put_perf` (unidirectional bandwidth, measured with blocking puts):
```
[0000] DEBUG: ../../src/transport_ofi.c:1589: shmem_transport_ctx_destroy
[0000]        id = -1, options = 0x80000000, stx_idx = 0
[0000]        pending_put_cntr =    422546, completed_put_cntr =    422546
[0000]        pending_get_cntr =         0, completed_get_cntr =         0
[0000]        pending_bb_cntr  =    140801, completed_bb_cntr  =    140801
[0001] DEBUG: ../../src/transport_ofi.c:1589: shmem_transport_ctx_destroy
[0001]        id = -1, options = 0x80000000, stx_idx = 0
[0001]        pending_put_cntr =       146, completed_put_cntr =       146
[0001]        pending_get_cntr =         0, completed_get_cntr =         0
[0001]        pending_bb_cntr  =         1, completed_bb_cntr  =         1
```

jdinan:
It's interesting that the target PE performed a bounce buffered put.  Wonder where that came from?

wrrobin:
This PR looks good. I would review the changes. Regarding the bb counter values, its surprising. Immediately got the reason why this counter was needed :P Also, the put counter values are high compared to what I would expect. Is it reflecting only this BW test?

jdinan:
It's the full application, including barriers, etc.  The put counter also includes non-fetching AMOs and the get counter includes fetching AMOs.

wrrobin:
Thanks. The full application includes all the message sizes. In that case, the high put counter values are making sense.

jdinan:
FYI, the +1 buffered put came from a call to `shmem_fcollect32` in `check_hostname_validation`.

PR #708, opened 2018-05-30
Incorporation of BB counters in perf counter APIs for OFI

wrrobin:
…n the BB counters


wrrobin:
We can do that. I kept it this way to keep it consistent with the ``quiet``. But, either way, there should not be any deadlock.

jdinan:
Thanks!  It's safer not to nest the mutexes unless we have a good reason for doing so (maybe we should revisit this in the quiet code).  Go ahead and merge when ready!

PR #712, opened 2018-06-12
Travis: Install cargo package for PRK build system

jdinan:


wrrobin:
@jdinan I have already added this in #710.

jdinan:
@wrrobin Can we merge this fix while we wrap up #710?  All Travis builds are currently failing because of this.

wrrobin:
Sure @jdinan. I will remove this fix from #710

PR #713, opened 2018-06-27
Fix FI_MR_RMA_EVENT used with FI_MR_SCALABLE.

matcabral:
FI_MR_SCALABLE is "old" (pre v1.5) OFI mr_mode
(still supported for backward compatibility)
that must not be mixed with new API mr_mode bits.

Signed-off-by: Matias Cabral <matias.a.cabral@intel.com>


lanl-ompi:

jdinan:
Travis failure was Cray finc_swap test, which is known to fail occasionally unrelated to this PR.  I restarted that row in the testing matrix.

matcabral:
> But it looks like the build fails when I try that. 

As you saw, libfabric v < 1.5 does not recognize FI_MR_RMA_EVENT.  I warning message with a graceful failure may not be a bad idea given that distros are slow to update versions.

jdinan:
(shrug) I'm fine with the compile-time failure.  What is kinda weird is that you need to both `--enable-mr-scalable` and `--enable-mr-rma-event` to enable this feature.  Might be nice to write the configury so that you only need `--enable-mr-rma-event`.

PR #731, opened 2018-07-13
Fix for BW mismatch with -t flag in perf suite

wrrobin:


wrrobin:
@jdinan @kseager This PR is ready for review now.

test/performance/shmem_perf_suite/common.h
```diff
@@ -578,8 +573,8 @@ int streaming_node(perf_metrics_t my_info)
 static inline
 int target_node(perf_metrics_t my_info)
 {
-    return (my_info.my_node >= my_info.midpt &&
-        (my_info.my_node < (my_info.midpt + my_info.sztarget)));
+    return (my_info.my_node >= my_info.szinitiator &&
+        (my_info.my_node < (my_info.szinitiator + my_info.sztarget)));
```
kseager: the reason midpt was compared was because we maintain the boundary between the half initiated for the initiator and the other half initiated for the target. Even when we decrease the amount on either side, we still maintain that boundary which defines initiator vs target

wrrobin: Thanks, @kseager. The reason I changed this is that when `l/r` flag is used with a different number other than the `midpt`, this method will return a wrong answer for some of the processes that are actually target. For example, if we use `-l 3` with a total of 16 processes, then this method will provide the wrong answer for PE 3 to 7 with the previous implementation. Let me know if I am getting something wrong here.

kseager: The original idea for this feature was the following:
-note that this assumes a process manager dividing the PEs at the halfway point across nodes
-note that this model wants to you launch the max PEs in the set and then reduce from that set

Thus the idea is to maintain that logical division. Its not the same idea as what I was doing with atomics where every PE is incast/active. In this case you reduce the active set (l/r stood for left and right).  You do however have to track the active set for later book keeping, but in your proposed example 3-7 is inactive and not target.

wrrobin: Oh.. thanks for confirming this. I was thinking that `-l 3` means 3 sources and 13 targets. But, based on what you said, it is 3 sources and 8 targets and 5 unused/inactive. In that case, I see the reasoning why it should be based on `midpt`. 

But, with this model, although there is no need to, it is not possible to run a variable number of sources and targets at the same time. I am guessing there is no need to.

kseager: my thought process was that if someone wanted less PEs on both sides then that person should just launch with less overall PEs.

wrrobin:
@kseager @jdinan Can you please review the new patches? I have reverted some of the changes earlier and also updated the way we are calculating the bandwidth when a target is selected by multiple initiators or vice versa. Please let me know if these changes look ok.

test/performance/shmem_perf_suite/common.h
```diff
@@ -679,32 +669,30 @@ int error_checking_init_target_usage(perf_metrics_t *metric_info) {
...
     assert(metric_info->midpt > 0);
 
-    if(metric_info->sztarget != -1 && metric_info->szinitiator != -1)
-        error = true; /* can't use them together  */
-
-    if(metric_info->sztarget != -1) {
-        if(metric_info->sztarget < 1 ||
-           metric_info->sztarget > metric_info->midpt ||
-           !metric_info->target_data) {
+    if (metric_info->sztarget != -1 && metric_info->szinitiator != -1) {
```
kseager: this if is empty

test/performance/shmem_perf_suite/target_put.h
```diff
@@ -107,34 +121,29 @@ static inline void target_data_uni_bw(int len, perf_metrics_t metric_info)
...
+                }
+                shmem_quiet();
             }
-            shmem_quiet();
             shmem_int_atomic_inc(&completion_signal, my_PE_partners[i]);
         }
     }
 
-    shmem_barrier_all();
+    //shmem_barrier_all();
```
kseager: why is this commented out?

test/performance/shmem_perf_suite/common.h
wrrobin: Yes, kept it intentionally. Earlier, it was an error condition, now both the l and r flags can be used at the same time if user wants. If that looks fine, we will remove this if block.

test/performance/shmem_perf_suite/target_put.h
wrrobin: Yes, I was going to ask this. I am not sure why we need a barrier at this point. Both sender and target are synchronized through the AMO and wait-until, right?

test/performance/shmem_perf_suite/target_put.h
```diff
@@ -107,34 +121,29 @@ static inline void target_data_uni_bw(int len, perf_metrics_t metric_info)
...
     if (snode || target_node(metric_info)) {
         end = perf_shmemx_wtime();
-        calc_and_print_results(end, start, len, metric_info);
+        calc_and_print_results(end, start, len, num_partners, metric_info);
     }
```
wrrobin: @kseager Can you please check this change whether it looks correct for target side calculation? Earlier, we were only sending the `length` not accounting how many of size `length` we are sending or receiving.

test/performance/shmem_perf_suite/bi_dir.h
```diff
@@ -78,7 +78,7 @@ static inline void bi_bw_put(int len, perf_metrics_t *metric_info)
         shmem_int_p(&fin, 1, dest);
         shmem_int_wait_until(&fin, SHMEM_CMP_EQ, 0);
         end = perf_shmemx_wtime();
-        calc_and_print_results(end, start, len, *metric_info);
+        calc_and_print_results(end, start, len, 1, *metric_info);
```
kseager: personally I'd prefer bundling this into metric_info calling it num_partners and setting one as default (override it for target), it would be easier to read

wrrobin: Sure, will do.

test/performance/shmem_perf_suite/bw_common.h
```diff
@@ -202,12 +204,18 @@ void calc_and_print_results(double end_t, double start_t, int len,
...
+                PEs_on_other_side = metric_info.szinitiator;
+            }
 #ifdef ENABLE_OPENMP
-            bw = ((double) len * (double) multiplier * (double) metric_info.midpt / 
+            bw = ((double) len * (double) multiplier * (double) PEs_on_other_side / 
                  1.0e6 * metric_info.window_size * metric_info.trials * 
                  (double) metric_info.nthreads) / (total_t_max / 1.0e6);
 #else
-            bw = ((double) len * (double) multiplier * (double) metric_info.midpt / 
+            bw = ((double) len * (double) multiplier * (double) PEs_on_other_side /
```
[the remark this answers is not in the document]
kseager: I believe the case this calculation is missing is when two initiators are sending to the same target? We essentially want to capture the number of messages sent converted to bytes. Something along the lines of ~ each PE's number of partners * iterations * len per message. Whats the reason for calculating it differently than the bw calculation you used above it?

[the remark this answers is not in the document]
wrrobin: Thanks Kayla. I had some doubt here as well. But, I kept it like this because of the following reason. If we use only `midpt` as before then choosing a different `l/r` would result to a wrong calculation. For example, if `midpt = 8` and `l = 5`, then using `midpt` would multiply with 8 and thus the BW calculated would be more than the correct one. 
Now, regarding the case when two initiators are sending to the same target, from initiator's perspective, each has one target and they send the `len` bytes from start to end time. So, the initiator bandwidth is multiplied by 1. For target, it has to receive the `len` bytes from two initiators within its `(end-start)` time, So, for target, the BW is calculated by multiplying by 2. I had the concern whether this is the correct way of thinking. Please let me know if it's not.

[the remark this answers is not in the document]
wrrobin: In the above case, should the initiator bandwidth also be multiplied by 2 to get the aggregated bandwidth of 2 initiators?

wrrobin:
@kseager @jdinan Ready for re-review.

wrrobin:
@kseager @jdinan If you guys do not have any comment, I am planning to merge this EOD today.

PR #732, opened 2018-07-13
Configury updates

jdinan:
* Removes `--disable-mr-scalable` and `--enable-mr-rma-event` and replaces with `--enable-ofi-mr={basic, scalable, rma-event}`
* Fixes `make distcheck` target with simple PMI
* Improves build configuration output from configure
* Renames `--with-total-data-ordering` to `--enable-total-data-ordering`

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -192,13 +192,13 @@ AS_IF([test "$enable_long_fortran_header" = "yes"], [FORTRAN_LONG_HEADER=" "], [
 AC_SUBST([FORTRAN_LONG_HEADER])
 AM_CONDITIONAL([HAVE_LONG_FORTRAN_HEADER], [test "$enable_long_fortran_header" = "yes"])
 
-AC_ARG_WITH([ofi-mr],
-    [AC_HELP_STRING([--with-ofi-mr],
+AC_ARG_ENABLE([ofi-mr],
+    [AC_HELP_STRING([--enable-ofi-mr=MODE],
                     [OFI memory registration mode: basic, scalable, or rma-event (default: scalable)])])
 
-AS_IF([test -z "$with_ofi_mr"], [with_ofi_mr="scalable"])
+AS_IF([test -z "$enable_ofi_mr"], [enable_ofi_mr="scalable"])
 
-AS_CASE([$with_ofi_mr],
+AS_CASE([$enable_ofi_mr],
         [basic],
             [],
         [scalable],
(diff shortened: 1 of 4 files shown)


jdinan:
Sample output from `../configure --with-ofi=... --enable-pmi-simple --enable-ofi-mr=basic`:
```
Compilers:
  C:              gcc
  C++:            g++ -std=gnu++11
  Fortran:        gfortran

Transports:
  Portals 4:      no
  OFI:            yes
  XPMEM:          no
  CMA:            no
  memcpy:         no

Global Options:
  Remote VA:      no
  Thread support: yes
  Error checking: no

OFI Transport:
  Memory reg.:    basic
  Addr. vector:   table
```

jdinan:
This is a fun new error.  What's weird is that it doesn't occur consistently across build.  Thanks a lot, Microsoft..  🤣 
```
[0000] ERROR: ../../src/init.c:243: shmem_internal_init
[0000]        Remote virtual addressing is enabled; however, address space layout randomization
[0000]        is present.  Disable ASLR or rebuild without '--enable-remote-virtual-addressing'.
[0000]        This error message can be disabled by setting SHMEM_DISABLE_ASLR_CHECK or building
[0000]        with --disable-aslr-check.
```

configure.ac
```diff
@@ -192,17 +192,22 @@ AS_IF([test "$enable_long_fortran_header" = "yes"], [FORTRAN_LONG_HEADER=" "], [
...
+                    [OFI memory registration mode: basic, scalable, or rma-event (default: scalable)])])
+
+AS_IF([test -z "$enable_ofi_mr"], [enable_ofi_mr="scalable"])
+
+AS_CASE([$enable_ofi_mr],
+        [basic],
+            [],
+        [scalable],
+            [AC_DEFINE([ENABLE_MR_SCALABLE], [1], [If defined, the OFI transport will use FI_MR_SCALABLE])],
+        [rma*event],
```
davidozog: Is this a wildcard?

davidozog: Should we document this new flag behavior in the README?

jdinan: You bet it is!  I was thinking it might be good to also accept 'rma_event'.  But, I don't mind restricting this, if you think it would be better.

It's helpful to document the basic/required options in the README file, but I prefer not to document all of the configure options.  Partially because it's a pain to keep such things in sync.  The README says "See 'configure --help' for documentation on available options."  To that end, there's a bunch of stuff we could delete from the README...

davidozog: Ah yes, the underscore.  That makes sense.

I don't think it matters much at all in this case, but out of curiosity, is it possible to do an m4 regex here for "-" or "_"?

Agreed on the README.

jdinan: No, it's m4sh, so this translates into a shell case statement.  It might be portable to use `?` instead of `*` for a single character wildcard.  I'll look into that.

Issue #430, opened 2017-09-08
GUPS -n parameter


davidozog:
Looks like the -n parameter (number of updates per PE) is silently ignored.

jdinan:
Should we just remove the parameter?  It doesn't appear to be part of the official benchmark specification (http://icl.cs.utk.edu/projectsfiles/hpcc/RandomAccess/).

davidozog:
Yeah, probably.  I don't see any reference to multiple updates in that link either.

jdinan:
I think our version originated here https://github.com/openshmem-org/gups-shmem, which in turn came from here http://icl.cs.utk.edu/hpcc/software/index.html, in case you're looking for breadcrumbs.

davidozog:
I had forgotten about this, but it's come up again.  On second thought, I actually recommend we implement the parameter instead of removing it.  The GUPS measurement can be extremely noisy with the default value (which is arbitrarily 4 times the number of table elements).  By adding the `-n` parameter, which is pretty trivial, it's convenient to increase the parameter where I see far less variation between executions, at the cost of just waiting a few more seconds.

PR #735, opened 2018-07-24
Add unused GUPS parameter: number of updates/PE

davidozog:
Addressing #430.

### DIFF
--- a/test/apps/gups.c
+++ b/test/apps/gups.c
@@ -178,7 +178,7 @@
 #define ZERO64B 0LL
 
 uint64_t TotalMemOpt = 8192;
-int NumUpdatesOpt = 0; /* FIXME: This option is ignored */
+int NumUpdatesOpt = 0;
 double SHMEMGUPs;
 double SHMEMRandomAccess_ErrorsFraction;
 double SHMEMRandomAccess_time;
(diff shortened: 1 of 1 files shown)


test/apps/gups.c
```diff
@@ -421,9 +421,13 @@ SHMEMRandomAccess(void)
...
 
   /* Default number of global updates to table: 4x number of table entries */
-  NumUpdates_Default = 4 * TableSize;
-  ProcNumUpdates = 4 * LocalTableSize;
-  NumUpdates = NumUpdates_Default;
+  if (NumUpdatesOpt == 0) {
+     ProcNumUpdates = 4 * LocalTableSize;
+     NumUpdates = 4 * TableSize;
+  } else {
+     ProcNumUpdates = NumUpdatesOpt;
```
jdinan: @davidozog Can you check against the MPI version of the benchmark to make sure this shouldn't be `ProcNumUpdates * NumUpdatesOpt`.

davidozog: [HPCC v1.5.0](http://icl.cs.utk.edu/hpcc/software/index.html) does not appear to support this parameter at all, it always performs `4 * TableSize` total updates.  

I'm fine just removing this adjustable parameter from the "Usage" output if you'd rather reflect HPCC.  But if we want it, either `ProcNumUpdates = NumUpdatesOpt` or `ProcNumUpdates = NumUpdatesOpt * LocalTableSize` would work, the former is just more fine-grained than the latter.  I'd guess the latter might be more puzzling to users though: "`-n`:  the scale factor of the local table size, which sets the total number of updates..." (or something similar) as opposed to: "`-n`:  the total number of updates."

jdinan: I filed #738 as a reminder to fix any remaining differences between our GUPS benchmark and HPCC RA.

jdinan: 🤷‍♂️ This seems reasonable to me.  At least the argument is not silently ignored any more.

jdinan: :shipit:

PR #741, opened 2018-08-01
Fix SHMEM_OFI_PROVIDER env var

jdinan:
Cherry-pick into v1.3.x

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,6 +59,7 @@ addons:
     apt:
         packages:
             - gfortran
+            - cargo
 
 before_install:
     ## Set up the environment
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -59,6 +59,7 @@ addons:
     apt:
         packages:
             - gfortran
+            - cargo
```
davidozog: I'm not sure I see the need for cargo.  My current understanding is that this is needed for the PRK Rust tests, but I have no idea why we'd be building those in SOS testing.

wrrobin: With the recent PRK repo, there had been some changes and if we do not install cargo, it fails because of a `cargo clean` command.

davidozog: Wow, that's weird.  So we never build or install anything with cargo, it's just the clean that fails?

Ok, I see the [issue report here](https://github.com/ParRes/Kernels/issues/361) now.  Thanks for letting me know!

PR #743, opened 2018-08-09
Check for sched_getaffinity, and include in DEBUG

davidozog:
I found this output useful for understanding what exactly is happening when passing process affinity options to the launcher such as `--bind-to core:2` or `--bind-to socket:1`.  Because it depends on `sched_getaffinity`, it'll only work in Linux.

### DIFF
--- a/src/init.c
+++ b/src/init.c
@@ -313,12 +313,11 @@ shmem_internal_init(int tl_requested, int *tl_provided)
         if (ret == 0) {
             char *cores_str = malloc(sizeof(char) * CPU_SETSIZE * 5 + 2);
             strcpy(cores_str," ");
+            size_t off = 1; /* start after " " */
             for (int i = 0; i < CPU_SETSIZE; i++) {
                 if (CPU_ISSET(i, &my_set)) {
                     core_count++;
-                    char cpunum[5];
-                    sprintf(cpunum, "%d ", i);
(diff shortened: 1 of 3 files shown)


src/init.c
```diff
@@ -292,6 +304,29 @@ shmem_internal_init(int tl_requested, int *tl_provided)
...
+            char *cores_str = malloc(sizeof(char) * CPU_SETSIZE * 5 + 2);
+            strcpy(cores_str," ");
+            for (int i = 0; i < CPU_SETSIZE; i++) {
+                if (CPU_ISSET(i, &my_set)) {
+                    core_count++;
+                    char cpunum[5];
+                    sprintf(cpunum, "%d ", i);
+                    strcat(cores_str, cpunum);
+                }
+            }
```
jdinan: I think a structure like we use in `DEBUG_MSG` where we `snprintf` directly into the string buffer might be better.  The `strcat` call is O(N) since it needs to seek to the end of the string each time (not a huge deal, but the intermediate string isn't saving any time).  Something like this avoids seeking each time and extra copies:

```
size_t off = 1; /* start after " " */
for (int i = 0; i < CPU_SETSIZE; i++) {
    if ... {
        ...
        off = snprintf(cores_str+off, CPU_SETSIZE*5+2-off, "%d ", i);
        ...
    }
...
}

PR #751, opened 2018-09-06
Optional flush of output and error streams on barrier

wrrobin:


wrrobin:
Thanks @jdinan. I am going with `SHMEM_BARRIERS_FLUSH`. Should we also add an entry in README?

jdinan:
Good catch, please add to the README.  Thanks!

wrrobin:
@jdinan This is ready for re-review.

PR #752, opened 2018-09-10
prrte in pmix testing

wrrobin:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -167,9 +167,14 @@ before_install:
     #- ./autogen.pl
     #- ./configure --prefix=$PWD/install --with-libevent=$TRAVIS_SRC/libevent-2.0.22-stable/install --with-pmix=$TRAVIS_SRC/pmix/install && make clean all install
     ## Download and install OpenMPI RPM 
-    - cd $TRAVIS_INSTALL
-    - wget http://gdurl.com/P5og -O openmpi-v3.0.x.deb
-    - sudo dpkg -i openmpi-v3.0.x.deb
+    #- cd $TRAVIS_INSTALL
+    #- wget http://gdurl.com/P5og -O openmpi-v3.0.x.deb
+    #- sudo dpkg -i openmpi-v3.0.x.deb
+    ## Download and install PRRTE
+    - cd $TRAVIS_SRC
+    - git clone --depth 10 --single-branch https://github.com/pmix/prrte.git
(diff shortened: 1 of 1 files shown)


wrrobin:
@jdinan, it looks like the Travis build has slowed down even with prrte. Any thoughts/suggestions?

davidozog:
Quick note:  I believe PMIx might need to be linked to the same `hwloc` as PRRTE.  So, we might have to update the `pmix` build to a newer version because I'm seeing that `v2.0` does not recognize the `--with-hwloc` flag. However, I'm seeing errors when I try to make this update on my laptop...  Do you see this too @wrrobin?

wrrobin:
For v2.0, I did not build PMIx with `--with-hwloc`, but the build was fine on Neptune. There were issues using the prrte launcher to launch jobs though, and not sure whether that is related to hwloc. 
I just tried with all the master branches for pmix, hwloc, and prrte and this seems to be working fine for me. I am thinking of using this for the testing setup.

config/opal_check_pmi.m4
```diff
@@ -303,6 +303,19 @@ AC_DEFUN([OPAL_CHECK_PMIX],[
...
+                                                      #include <pmix_version.h>
+                                                      #if (PMIX_VERSION_MAJOR != 4L)
+                                                      #error "not version 4"
+                                                      #endif
+                                                      ], [])],
+                                    [AC_MSG_RESULT([found])
+                                     opal_external_pmix_version=4x
+                                     opal_external_pmix_version_found=1],
+                                    [AC_MSG_RESULT([not found])])])
+
```
davidozog: Cool, this comes from [upstream here right](https://github.com/open-mpi/ompi/blob/master/config/opal_check_pmi.m4)?

I see there a lot more upstream changes too.  Do we want to bring those in as well?

wrrobin: Yes, we can bring those in. But first I am trying to validate that with PMIx4, we can resolve the issues that we are seeing with PRRTE. I just checked this branch on Jenkins, it still fails but fails with a missing library (libcrypto) error.

jdinan:
Status: PRRTE server still crashing at random points during testing.

Next steps: Reproduce crash, track down source, and bother Ralph (or fix it) (or both).

wrrobin:
Related issue: pmix/prrte#131 submitted

PR #755, opened 2018-09-13
Removal of FI_REMOTE_READ from target counter

wrrobin:

### DIFF
--- a/src/transport_ofi.c
+++ b/src/transport_ofi.c
@@ -663,7 +663,7 @@ int allocate_recv_cntr_mr(void)
 
 #if defined(ENABLE_MR_SCALABLE) && defined(ENABLE_REMOTE_VIRTUAL_ADDRESSING)
     ret = fi_mr_reg(shmem_transport_ofi_domainfd, 0, UINT64_MAX,
-                    FI_REMOTE_READ | FI_REMOTE_WRITE, 0, 0ULL, flags,
+                    FI_REMOTE_WRITE | FI_REMOTE_READ, 0, 0ULL, flags,
                     &shmem_transport_ofi_target_mrfd, NULL);
     OFI_CHECK_RETURN_STR(ret, "target memory (all) registration failed");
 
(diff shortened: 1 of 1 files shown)


src/transport_ofi.c
```diff
@@ -663,19 +663,19 @@ int allocate_recv_cntr_mr(void)
 
 #if defined(ENABLE_MR_SCALABLE) && defined(ENABLE_REMOTE_VIRTUAL_ADDRESSING)
     ret = fi_mr_reg(shmem_transport_ofi_domainfd, 0, UINT64_MAX,
-                    FI_REMOTE_READ | FI_REMOTE_WRITE, 0, 0ULL, flags,
+                    FI_REMOTE_WRITE | FI_REMOTE_READ, 0, 0ULL, flags,
```
sayantansur: Spurious change? I'm not the maintainer, so don't know if you guys care about changes like this.

wrrobin: Thanks @sayantansur. This happened because of removing `FI_REMOTE_READ` at first and then adding it back. Will take care of this.

src/transport_ofi.c
```diff
@@ -663,19 +663,19 @@ int allocate_recv_cntr_mr(void)
...
 #if ENABLE_TARGET_CNTR
     ret = fi_mr_bind(shmem_transport_ofi_target_mrfd,
                      &shmem_transport_ofi_target_cntrfd->fid,
-                     FI_REMOTE_WRITE | FI_REMOTE_READ);
+                     FI_REMOTE_WRITE);
     OFI_CHECK_RETURN_STR(ret, "target CNTR binding to MR failed");
 
     ret = fi_ep_bind(shmem_transport_ofi_target_ep,
-                     &shmem_transport_ofi_target_cntrfd->fid, FI_REMOTE_WRITE | FI_REMOTE_READ);
+                     &shmem_transport_ofi_target_cntrfd->fid, FI_REMOTE_WRITE);
```
sayantansur: Hmm. I don't know what the OFI man pages say. Can you bind the same cntr to the EP and MR, both to count incoming remote reads? And why do you need both these? need to think if we can drop the second EP/cntr bind?

wrrobin: I could not find anything on the manpages regarding not binding the same cntr to both EP and MR. Also, not sure whether we can safely remove the second one. @jdinan any thoughts?

jdinan: 🤷‍♂️

jdinan:
@wrrobin Do you want to merge this and file an issue to follow-up on binding to both the MR and EP?

wrrobin:
Sure, moving ahead. Thanks.

Issue #677, opened 2018-04-18
OFI: Rand is not thread safe


jdinan:
In `transport_ofi.h` use `rand_r` instead of `rand` to avoid problems with threading (e.g. if an application thread is also using `rand`).

davidozog:
Making this change seems surprisingly non-trivial.  We need a thread-local `unsigned int` seed for `rand_r`, and we can't safely use `shmem_internal_tid` type directly because it's either `pid_t` or `uint64_t`, right?  Not sure I feel ok about a thread-local `time()` seed either.   Even getting a portable thread-local variable with c99 seems non-trivial (see [here](https://stackoverflow.com/questions/18298280/how-to-declare-a-variable-as-thread-local-portably)).  We could use `pthread_get/set_specific` I suppose...  Any thoughts?  Maybe I'm missing something.

jdinan:
Is this for the STX allocator?  If so, there's a mutex for the STX allocator / context creation -- can we create a shared seed variable that's used by all threads in the PE and protected by that mutex?  Maybe initialize to the PE id?

PR #761, opened 2018-09-13
Fixes to the random STX allocator 

davidozog:
This makes the following changes to the random STX allocator:
* Call reentrant `rand_r` for thread safety (see #677).
* remove the "Fisher-Yates shuffle" scheme for fairness (it wasn't effective).
* Increase the "max reference" threshold on the default context by 1 for better balancing.
* Fill STX "slots" up to the threshold before picking another random STX.


jdinan:
It looks like this random allocator may say "none available" (-1) for the given threshold, even when such an STX is available.  Is that ok, or should we force it to keep probing as long as such an STX is available?

davidozog:
That's true.  I figure if random selection makes "sufficiently many" attempts, and it hasn't found such an STX, then it's just not worth the effort and best to commit to the next random selection (which is what happens with that -1).  I found that this relatively unfair randomness balances well enough (especially if the app has enough contexts relative to STXs).  

I'd be open to tracking whether such an STX is available if you'd like, but I'm not sure it's worth the extra code.  We could also increase the `shmem_transport_ofi_stx_max` heuristic to `2 * shmem_transport_ofi_stx_max` or something like that to reduce the chance this occurs.

jdinan:
@davidozog Have you had a chance to review https://github.com/davidozog/sandia-shmem/pull/5?

davidozog:
@jdinan Not yet, will review today.

PR #766, opened 2018-09-14
OFI: Only bind target cntr to MRs, not EP

### LINKED ISSUE #765: Removal of FI_REMOTE_READ in target EP for v1.4.1

jdinan:
This appears to work fine with the sockets provider.  Needs to be tested with the PSM2 provider.

Closes #765

### DIFF
--- a/src/transport_ofi.c
+++ b/src/transport_ofi.c
@@ -1277,7 +1277,10 @@ static int shmem_transport_ofi_target_ep_init(void)
 
     struct fabric_info* info = &shmem_transport_ofi_info;
     info->p_info->ep_attr->tx_ctx_cnt = 0;
-    info->p_info->caps = FI_RMA | FI_ATOMICS | FI_REMOTE_READ | FI_REMOTE_WRITE | FI_RMA_EVENT;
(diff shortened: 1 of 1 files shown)


jdinan:
@davidozog Could you test this branch with the PSM2 provider?  Thanks!

davidozog:
I presume you're interested in testing this change on PSM2 without hard polling, correct?  If so, this still deadlocks with 2 or more PEs on libfabric v1.6.x (I'd guess because we're not seeing any atomic events on the target counter).  With hard polling of course, all tests pass.

jdinan:
Thanks!  It's helpful to know that this is deadlocking on its own (versus the separate patch change to remove `FI_REMOTE_READ` on the cntr).

davidozog:
@jdinan, I added FI_RMA_EVENT to the target EP at @j-xiong's suggestion.  All unit tests then pass on PSM2.  Does that change look ok?

jdinan:
@davidozog Could you please review/test with 6530ffc?

davidozog:
Yup, good call - we only need this on the target counter.  It checks out on PSM2, with and without that target counter.

I think this is ready to merge.  I see @sayantansur is the reviewer - does it look good to you?

PR #773, opened 2018-09-25
Release the ctx lock while blocked

jdinan:
We may also want to ensure that we perform completion polling when `USE_CTX_LOCK` is enabled and the threading level is `SHMEM_THREAD_MULTIPLE`.


wrrobin:
Looks good to me.

jdinan:
@davidozog Added a check to see if the user set the polling limits before overriding them.

PR #775, opened 2018-10-02
Add proposed NBI fetching atomics as shmemx routines

jdinan:
Add "shmemx" extension to support nonblocking atomic operations that are proposed for OpenSHMEM 1.5.

## To-Do Items

- [x] Update to match proposed API
- [x] C11 generic bindings
- [x] C++ generic bindings
- [x] Portals support
- [x]  Transport none stubs

## References:
 * https://github.com/openshmem-org/specification/issues/246

### DIFF
--- a/src/transport_ofi.h
+++ b/src/transport_ofi.h
@@ -806,7 +806,7 @@ void shmem_transport_cswap(shmem_transport_ctx_t* ctx, void *target, const void 
     SHMEM_TRANSPORT_OFI_CNTR_INC(&ctx->pending_get_cntr);
 
     do {
-        ret = fi_compare_atomic(ctx->cntr_ep,
+        ret = fi_compare_atomic(ctx->ep,
                                 source,
                                 1,
                                 NULL,
@@ -861,7 +861,7 @@ void shmem_transport_cswap_nbi(shmem_transport_ctx_t* ctx, void *target, const
     SHMEM_TRANSPORT_OFI_CNTR_INC(&ctx->pending_get_cntr);
 
     do {
-        ret = fi_compare_atomicmsg(ctx->cntr_ep,
+        ret = fi_compare_atomicmsg(ctx->ep,
                                    &msg,
                                    &comparev,
                                    NULL,
@@ -883,7 +883,7 @@ void shmem_transport_mswap(shmem_transport_ctx_t* ctx, void *target, const void 
     uint64_t dst = (uint64_t) pe;
     uint64_t polled = 0;
     uint64_t key;
- cntr_cntr_   uint8_t *addr;
+    uint8_t *addr;
 
     shmem_transport_ofi_get_mr(target, pe, &addr, &key);
 
@@ -1124,7 +1124,7 @@ void shmem_transport_fetch_atomic_nbi(shmem_transport_ctx_t* ctx, void *target,
     SHMEM_TRANSPORT_OFI_CNTR_INC(&ctx->pending_get_cntr);
 
     do {
-        ret = fi_fetch_atomicmsg(ctx->cntr_ep,
+        ret = fi_fetch_atomicmsg(ctx->ep,
                                  &msg,
                                  &resultv,
                                  NULL,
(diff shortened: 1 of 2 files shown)


jdinan:
Looks like there are some API differences between this prototype and the proposed API: https://github.com/openshmem-org/specification/pull/247/files#diff-577ca9a50a06c049ba0518f4d359749a
This should get updated before merging.

jdinan:
@davidozog @wrrobin This PR is ready for review

src/transport_ofi.h
```diff
@@ -488,6 +488,8 @@ int shmem_transport_fence(shmem_transport_ctx_t* ctx)
      * non-fetching atomics to be completed in order to ensure ordering. */
     shmem_transport_put_quiet(ctx);
 #endif
+    /* Complete fetching ops; needed to support nonblocking fetch-atomics */
+    shmem_transport_get_wait(ctx);
```
wrrobin: @jdinan Was this a bug earlier? This seems to be covering any fetch operations now.

src/transport_ofi.h
```diff
@@ -833,28 +837,39 @@ void shmem_transport_cswap(shmem_transport_ctx_t* ctx, void *target, const void
...
+        ret = fi_compare_atomicmsg(ctx->ep,
+                                   &msg,
+                                   &comparev,
+                                   NULL,
+                                   1,
+                                   &resultv,
+                                   NULL,
+                                   1,
+                                   FI_INJECT);
     } while (try_again(ctx, ret, &polled));
```
wrrobin: Just to understand, the `fi_*msg` routine call is the difference between the non-blocking and blocking routine?

test/unit/cxx_test_shmem_atomic_and.cpp
```diff
@@ -37,16 +37,49 @@
...
+          break;
+#else
+#define SHMEMX_NBI_OPS_CASES(OP, TYPE)
+#endif
+
 #define TEST_SHMEM_AND(OP, TYPE)                                        \
   do {                                                                  \
     static TYPE remote = ~(TYPE)0;                                      \
+    const int mype = shmem_my_pe();                                     \
+    const int npes = shmem_n_pes();                                     \
```
wrrobin: I dont understand this. How was it working before without `mype` and `npes` defined?

test/unit/cxx_test_shmem_atomic_inc.cpp
```diff
@@ -90,6 +119,7 @@ enum op { INC = 0, ATOMIC_INC, CTX_ATOMIC_INC, FINC, ATOMIC_FETCH_INC,
             rc = EXIT_FAILURE;                                          \
           }                                                             \
           break;                                                        \
+        SHMEMX_NBI_OPS_CASES(OP, TYPE)                                  \
```
wrrobin: Adding the switch cases like this is great to read.

src/transport_ofi.h
jdinan: This was a deliberate optimization, since fence did not need to complete anything on the get counter.  With the addition of fetch-atomics, this is required to order the atomic update.  This change does have the unfortunate side effect of making fence slower in instances where nonblocking get operations are now completed, but were not previously.

src/transport_ofi.h
jdinan: Yes and also the nonblocking routine uses `FI_INJECT` whereas the blocking routine does not.   The blocking routine /must/ be completed before returning to the application because we use temporary buffering on the stack.

test/unit/cxx_test_shmem_atomic_inc.cpp
jdinan: Yeah, you can't have preprocessor conditions (check for ENABLE_SHMEMX_TESTS) in a macro definition, so you end up needing another macro to capture the conditional.  Otherwise I would have preferred put the code in here directly.

jdinan:
@wrrobin I restarted the Travis row in question.  Thanks for the feedback -- any other comments?

wrrobin:
Thanks @jdinan. No, the PR looks good. I went over the core changes in transport layer twice. The other bindings related changes and the unit tests look good to me as well. Hope, I have not missed anything in these.

jdinan:
@davidozog Ready to merge this?

PR #789, opened 2018-10-23
Check malloc return value during init.

matcabral:
malloc return value not checked reported by Klocworks.

Signed-off-by: Matias A Cabral <matias.a.cabral@intel.com>


lanl-ompi:

jdinan:
Thanks for the patch!

PR #790, opened 2018-11-02
Sanity check on PMI max key/val length

matcabral:
Security KW finding not limiting strncpy to destination buffer length

Signed-off-by: Matias A Cabral <matias.a.cabral@intel.com>

### DIFF
--- a/src/runtime-pmi.c
+++ b/src/runtime-pmi.c
@@ -239,6 +239,8 @@ shmem_runtime_put(char *key, void *value, size_t valuelen)
     if (size == 1) {
         singleton_kvs_t *e = malloc(sizeof(singleton_kvs_t));
         if (e == NULL) return 3;
+        shmem_internal_assertp(max_key_len <= SINGLETON_KEY_LEN);
+        shmem_internal_assertp(max_val_len <= SINGLETON_VAL_LEN);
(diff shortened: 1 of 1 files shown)


davidozog:
Out of curiosity, does this make KW happy?  If so, I'm impressed that it's smart enough to detect that `shmem_internal_assertp` does the right thing here.

lanl-ompi:

davidozog:
FYI: All the test errors are due to #791, and are unrelated to this PR.

matcabral:
I honestly don't know, will have to wait until the auto scan runs. However, if shows up, we can safely flag it as false positive after this patch ;)

jdinan:
Travis failures are being caused by broken Mellanox test.  I think this is safe to merge without waiting on CI.  @davidozog What do you think?

davidozog:
I think it is safe as well.  If it fixes the problem on @matcabral's end, then absolutely.  I'm curious to see if Klockwork is ok with the internal assert (doubt it, but I would be impressed if so).

matcabral:
@davidozog , FYI, KW did not understand the assert and screamed again last night. However, I just told him that the assert makes things safe and sent him back to bed ;)

PR #797, opened 2018-11-13
Classify CXX generic tests as shmemx tests

### LINKED ISSUE #796: C++ Generic Tests Are "shmemx" Tests
The C++ generic tests use an SOS extension (C++ generic bindings) and should be classified as shmemx tests.

jdinan:
Closes #796

### DIFF
--- a/test/unit/Makefile.am
+++ b/test/unit/Makefile.am
@@ -111,7 +111,10 @@ endif
 
 if HAVE_CXX
 check_PROGRAMS += \
-	cxx_test_shmem_complex \
+	cxx_test_shmem_complex
+
+if SHMEMX_TESTS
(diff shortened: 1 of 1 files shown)


davidozog:
Should these tests move to the `test/shmemx` directory?

jdinan:
I was noodling on that same thing, myself.  Since the APIs called are `shmem_*` and not `shmemx_*` it seemed appropriate to leave them here.  The rationale for the naming is that they support the C11 generic APIs in C++ translation environments (using C++ rather than C11 constructs).  But, since this isn't required by the standard, it's technically an extension.  I might be persuaded to move them for this reason, or possibly to create a `test/cxx` directory if we want a third option.

PR #804, opened 2018-11-20
TESTS: added external tests build

hoopoepg:
- added configure switch --enable-external-tests to enable build of unit
  tests using external infrastructure
- added build target to unit tests to build tests only, without
  test run

Signed-off-by: Sergey Oblomov <sergeyo@mellanox.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -245,7 +245,6 @@ AC_C_INLINE
 AC_PROG_CXX dnl required even with --disable-cxx due to automake conditionals
 m4_ifdef([AM_PROG_AR],[AM_PROG_AR])
 AC_PATH_PROG([PERL],[perl])
-AM_CONDITIONAL([EXTERNAL_TESTS], [false])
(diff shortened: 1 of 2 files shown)


lanl-ompi:

jdinan:
@hoopoepg Are you aware of https://github.com/openshmem-org/tests-sos?  If all you want are the tests, then this is a better way to get them.

hoopoepg:
@jdinan yes, we would like to use tests for internal testing of oshmem, I have seen that repo, but as I can see tests in this repo is more actual (at least file's timestamp are different).

do you think we should use https://github.com/openshmem-org/tests-sos repo?

jdinan:
The tests-sos project is updated with each SOS release and we can always update the master branch earlier if there are tests on the SOS master that folks need (the process is usually just `cp -r sos/tests sos-tests`).  I prefer not to use the approach from this PR, since it means users need to clear the full SOS configure (process manager, etc) in order to build the tests.

hoopoepg:
ok, thank you

PR #810, opened 2018-12-06
Split definitions out of shmem.h

### LINKED ISSUE #808: Separate types and constants from shmem.h for use in pshmem.h

jdinan:
Split out `shmem-defs.h` so the `pshmem.h` and `shmemx.h` headers don't need to include the full `shmem.h` header file.  This has the nice side effect of simplifying the header generation.

Closes #808


jdinan:
Ready for review

mpp/pshmem.h4
```diff
@@ -22,7 +22,7 @@ include(shmem_bind_c.m4)dnl
 #ifndef PSHMEM_H
 #define PSHMEM_H
 
-#include <shmem.h>
+#include <shmem-def.h>
 #include <shmemx.h>
```
wrrobin: @jdinan Just a thought, should we keep `shmemx.h` in pshmem? Or should we create `shmemx-def.h` and use it in the same way? At this point, we have only a couple of data types to put there, but may become handy when we have newer ones for 1.5?

jdinan: Sure, this will help to simplify `pshmem.h` a bit more.

jdinan:
@wrrobin Could you review 0ce66a7?  Thanks!

wrrobin:
@jdinan I think `mpp/shmemx-def.h.in` has not been checked in.

jdinan:
Woops, looks like it was something simple after all!  😀

jdinan:
@wrrobin Ready for re-review.

Issue #812, opened 2018-12-06
PMIX configury out of date?


davidozog:
With the `opal_check_pmi.m4` file in SOS, we can't link with the master branch of pmix.  However, the upstream version of this file (I tested OMPI v3.0.x) has some recent changes and has no problem linking with pmix master.  Perhaps this could be updated in v1.4.3?

Issue #750 is related.

jdinan:
Closed by #813

PR #813, opened 2018-12-06
Fix opal_check_pmi.m4 for newer PMIx

rhc54:
Quick fix: provide paragraph for PMIx v4.x

Signed-off-by: Ralph H Castain <rhc@open-mpi.org>

### DIFF
--- a/config/opal_check_pmi.m4
+++ b/config/opal_check_pmi.m4
@@ -13,7 +13,7 @@
 # Copyright (c) 2009-2015 Cisco Systems, Inc.  All rights reserved.
 # Copyright (c) 2011-2016 Los Alamos National Security, LLC. All rights
 #                         reserved.
-# Copyright (c) 2014      Intel, Inc. All rights reserved.
+# Copyright (c) 2014-2018 Intel, Inc.  All rights reserved.
 # Copyright (c) 2014      Research Organization for Information Science
 #                         and Technology (RIST). All rights reserved.
 # $COPYRIGHT$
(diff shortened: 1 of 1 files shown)


lanl-ompi:

jdinan:
ok to test

davidozog:
I just verified this does fix #812.  Thanks @rhc54!

In the past we've preferred to take entire upstream OPAL files as opposed to individual patches (to avoid future merges/maintenance on our end).  What do you think in this case @jdinan?

rhc54:
Ah, didn't know that - you are welcome to do so, if you like. This was taken from the upstream file.

config/opal_check_pmi.m4
```diff
@@ -303,6 +303,19 @@ AC_DEFUN([OPAL_CHECK_PMIX],[
 
            # if it does exist, then we need to parse it to find
            # the actual release series
+           AS_IF([test "$opal_external_pmix_version_found" = "0"],
+                 [AC_MSG_CHECKING([version 4x])
+                  AC_PREPROC_IFELSE([AC_LANG_PROGRAM([
+                                                      #include <pmix_version.h>
+                                                      #if (PMIX_VERSION_MAJOR < 4L)
```
jdinan: Is it better to write `PMIX_VERSION_MAJOR != 4L` so we don't need to update this for PMIx 5?

jdinan:
I'm fine with this patch for now.  We can file an issue to sync with upstream later.  🤷‍♂️

config/opal_check_pmi.m4
rhc54: Actually, we chose this method so it won't have to be changed - we are checking that the version is 4 or higher. If it is, then all the code that uses PMIx 4 or lower APIs will still be good. If you someday update to PMIx 5 APIs, then you would need to check for 5 or higher.

jdinan: Ok, as long as you guys promise to never break backward compatibility.  :neckbeard: 🤞

rhc54: Yeah, that is the promise we made!

davidozog: We'll be just fine as long as nobody retires!  😉

PR #815, opened 2018-12-11
Optimized put with signal implementation as shmemx routines

wrrobin:


src/transport_ofi.h
```diff
@@ -658,6 +658,138 @@ void shmem_transport_put_nb(shmem_transport_ctx_t* ctx, void *target, const void
...
+                                           .iov_count = 1,
+                                           .addr = GET_DEST(dst),
+                                           .rma_iov = &rma_iov_signal,
+                                           .rma_iov_count = 1,
+                                           .context = (uint8_t *) &signal,
+                                           .data = 0
+                                         };
+
+    do {
+        ret = fi_writemsg(ctx->ep, &msg_signal, FI_DELIVERY_COMPLETE | FI_FENCE | FI_INJECT);
```
wrrobin: @shefty @sayantansur Would this be a correct way to use FI flags for the signal transmission in put-with-signal (Ref: openshmem-org/specification#206)? 
Also, do we need to have `FI_DELIVERY_COMPLETE` here as we have already configured the endpoint with this flag during creation through 
```
tx_attr.op_flags          = FI_DELIVERY_COMPLETE;
```
In other words, do the flags provided with `fi_writemsg` gets OR'ed with the flags already configured? Or, they are replaced?

jdinan: What about `fi_atomicmsg`?  Manpage doesn't list `FI_DELIVERY_COMPLETE` as a flag on this operation?

jdinan: This was answered on #818.  Looks good.

jdinan: The question I have here is whether OFI guarantees single copy atomicity from `fi_writemsg`.  If not, we're relying on an extension and we should use `fi_atomicmsg` unless the user enables this at compile time (e.g. `--enable-ofi-writemsg-is-atomic`).

src/transport_ofi.h
```diff
@@ -658,6 +658,138 @@ void shmem_transport_put_nb(shmem_transport_ctx_t* ctx, void *target, const void
...
+
+            do {
+                ret = fi_writemsg(ctx->ep, &msg, FI_DELIVERY_COMPLETE);
+            } while (try_again(ctx, ret, &polled));
+
+            frag_source += frag_len;
+            frag_target += frag_len;
+        }
+        SHMEM_TRANSPORT_OFI_CTX_UNLOCK(ctx);
+        if ((*completion) != -1) (*completion)++;
```
jdinan: This isn't going to work.  😃

src/shmem_comm.h
```diff
@@ -101,6 +101,40 @@ shmem_internal_put_nb(shmem_ctx_t ctx, void *target, const void *source, size_t
...
+        shmem_internal_put_scalar(ctx, sig_addr, &signal, sizeof(uint64_t), pe);
+#elif USE_XPMEM
+        shmem_transport_xpmem_put(target, source, len, pe, node_rank);
+        shmem_internal_put_scalar(ctx, sig_addr, &signal, sizeof(uint64_t), pe);
+#elif USE_CMA
+        if (len > shmem_internal_params.CMA_PUT_MAX) {
+            shmem_transport_put_signal((shmem_transport_ctx_t *)ctx, target, source, len, sig_addr, signal, pe, completion);
+        } else {
+            shmem_transport_cma_put(target, source, len, pe, node_rank);
+            shmem_internal_put_scalar(ctx, sig_addr, &signal, sizeof(uint64_t), pe);
```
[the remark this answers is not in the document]
wrrobin: @jdinan Instead of `shmem_internal_put_scalar`, will `shmem_transport_xpmem/cma_put` make more sense? The `shmem_transport_xpmem/cma_put` internally uses memcpy/process_vm_writev.. so, using these would not require any fence, right?

src/shmem_comm.h
```diff
@@ -101,6 +101,40 @@ shmem_internal_put_nb(shmem_ctx_t ctx, void *target, const void *source, size_t
     }
 }
 
+static inline
+void
+shmem_internal_put_signal(shmem_ctx_t ctx, void *target, const void *source, size_t len, 
+                          uint64_t *sig_addr, uint64_t signal, int pe, long *completion)
```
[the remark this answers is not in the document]
wrrobin: Based on our discussion, I will revert the changes for blocking API. In that way, the `completion` argument will not be needed.

src/transport_ofi.h
wrrobin: The text for `fi_writemsg` in the man pages does not mention anything such. Switching to `fi_atomicmsg` for now.

wrrobin:
@jdinan @davidozog Ready for re-review.

src/transport_portals4.h
```diff
@@ -643,6 +643,15 @@ shmem_transport_put_ct_nb(shmem_transport_ct_t *ct, void *target, const void *so
...
 
+static inline
+void shmem_transport_put_signal_nbi(shmem_transport_ctx_t* ctx, void *target, const void *source, size_t len,
+                                    uint64_t *sig_addr, uint64_t signal, int pe)
+{
+    shmem_transport_put_nbi(ctx, target, source, len, pe);
+    shmem_transport_fence(ctx);
+    shmem_transport_put_scalar(ctx, sig_addr, &signal, sizeof(uint64_t), pe);
+}
+
```
jdinan: Could you add a "FIXME" comment and/or file an issue to go back and optimize nonblocking put-with-signal in the Portals transport?

src/shmem_comm.h
```diff
@@ -101,6 +103,42 @@ shmem_internal_put_nb(shmem_ctx_t ctx, void *target, const void *source, size_t
...
+void
+shmem_internal_put_signal_nbi(shmem_ctx_t ctx, void *target, const void *source, size_t len, 
+                              uint64_t *sig_addr, uint64_t signal, int pe)
+{
+    int node_rank;
+
+    if (len == 0) {
+        shmem_internal_put_scalar(ctx, sig_addr, &signal, sizeof(uint64_t), pe);
+        return;
+    }
```
jdinan: Code looks good.  Just wanted to suggest double checking the proposal that this is the right way to handle the `len == 0` case.

src/transport_ofi.h
```diff
@@ -658,6 +658,139 @@ void shmem_transport_put_nb(shmem_transport_ctx_t* ctx, void *target, const void
...
+                                        .iov_count = 1,
+                                        .addr = GET_DEST(dst),
+                                        .rma_iov = &rma_iov,
+                                        .rma_iov_count = 1,
+                                        .context = src_buf,
+                                        .data = 0
+                                      };
+
+        do {
+            ret = fi_writemsg(ctx->ep, &msg, FI_DELIVERY_COMPLETE);
```
jdinan: Do you mean to add the `FI_INJECT` flag here (this is the buffered send case)?

src/transport_ofi.h
```diff
@@ -658,6 +658,139 @@ void shmem_transport_put_nb(shmem_transport_ctx_t* ctx, void *target, const void
...
+
+            SHMEM_TRANSPORT_OFI_CNTR_INC(&ctx->pending_put_cntr);
+
+            do {
+                ret = fi_writemsg(ctx->ep, &msg, FI_DELIVERY_COMPLETE);
+            } while (try_again(ctx, ret, &polled));
+
+            frag_source += frag_len;
+            frag_target += frag_len;
+        }
```
jdinan: Just making an observation that when fragmentation occurs, the signal will be signaling completion of multiple prior puts.  I re-read the `FI_FENCE` text and I think this is ok.  It might not hurt to toss a comment somewhere to the effect that multiple prior operations may be fenced by the signal put when fragmentation occurs.

src/transport_ofi.h
```diff
@@ -658,6 +658,139 @@ void shmem_transport_put_nb(shmem_transport_ctx_t* ctx, void *target, const void
...
+                                           .rma_iov = &rma_iov_signal,
+                                           .rma_iov_count = 1,
+                                           .datatype = FI_UINT64,
+                                           .op = FI_ATOMIC_WRITE,
+                                           .context = (uint8_t *) &signal,
+                                           .data = 0
+                                         };
+
+    do {
+        ret = fi_atomicmsg(ctx->ep, &msg_signal, FI_DELIVERY_COMPLETE | FI_FENCE | FI_INJECT);
```
jdinan: The open question is, whether we also need to use an atomic to implement `shmem_p`.  Could you add a comment here and open an issue so we remember to revisit this?

src/transport_portals4.h
wrrobin: Done

src/shmem_comm.h
wrrobin: There is no specific text added in the proposal for this case. However, in the discussion, it is mentioned that sending the signal is the expected behavior.

src/transport_ofi.h
wrrobin: Yes, `FI_INJECT` should be added here. Thanks.

src/transport_ofi.h
wrrobin: Added.

src/transport_ofi.h
wrrobin: Issue opened.

PR #818, opened 2018-12-12
Fix for flags in fi_writemsg APIs

wrrobin:

### DIFF
--- a/src/transport_ofi.h
+++ b/src/transport_ofi.h
@@ -869,7 +869,8 @@ void shmem_transport_cswap_nbi(shmem_transport_ctx_t* ctx, void *target, const
                                    &resultv,
                                    NULL,
                                    1,
-                                   FI_INJECT);
+                                   FI_INJECT);  /* FI_DELIVERY_COMPLETE is not required as 
+                                                   it is implied for fetch atomicmsgs */
     } while (try_again(ctx, ret, &polled));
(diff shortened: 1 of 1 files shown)


jdinan:
@shefty Do we need to also make the same change to `fi_atomicmsg` API calls (we pass only `FI_INJECT` flag currently)?  The completion models aren't listed under flags for the atomic-msg APIs.

shefty:
@jdinan - I think that's just an oversight in the man pages.  FI_DELIVERY_COMPLETE is implied for any 'fetch' atomic operation.

wrrobin:
@shefty Just to understand, so, we will not need to explicitly pass `FI_DELIVERY_COMPLETE` as an argument for `fi_fetch_atomicmsg`, but need to do so for `fi_atomicmsg` and `fi_compare_atomic_msg`?

shefty:
fi_fetch_atomicmsg and fi_compare_atomicmsg can both return results back to the caller.  If results are being returned (i.e. fetched) then FI_DELIVERY_COMPLETE is implied, though it's harmless to request it.  You would need it for fi_atomicmsg.

wrrobin:
@jdinan Please re-review the new changes. Thanks.

jdinan:
Thanks, @wrrobin and @shefty!

Issue #806, opened 2018-11-28
Update README to use markdown


jdinan:
Ref #805

jdinan:
Closed by #819

Issue #805, opened 2018-11-28
Link to SOS tests from SOS github


jdinan:
Add a link somewhere so folks can find https://github.com/openshmem-org/tests-sos

jdinan:
Closed by #819

PR #819, opened 2018-12-14
Change README to use markdown

davidozog:
Here is a preview of what this would look like:
https://github.com/davidozog/sandia-shmem/tree/pr/readme_markdown#-sandia-openshmem

This addresses issues #805 and #806.


davidozog:
@jdinan  Yeah, having both a README and README.md file would be nice.  In the future we'd just need to update 2 different files for README changes.

On the other hand, I'd say the Markdown file would be much more readable if we removed the SOS logo images and reverted the parameters/env vars descriptions to use the original format instead of the HTML table.  I slightly prefer your 2 README files idea, but just thought I'd throw this alternative out there.

jdinan:
I was thinking we would treat `README.md` as a "web page" that is displayed on GitHub and include only infrequently changing details.  Hopefully that would cut down on the need to make updates in both places.

davidozog:
TODO: Link to README, Wiki, Travis, Coverity.

davidozog:
@jdinan, the badges and links are ready.  Do they look ok to you?

Side note: the Travis badge will say "error" until we fix the issue where some tests consistently time out after merging.

davidozog:
@jdinan, also that Coverity badge might be linking to my account.  Does it look like the correct "project" to you?  Feel free to link to the project on your account, since you run it more often.

jdinan:
Coverity badge looks correct to me.  The version badge isn't displaying properly, but I think this is because you don't have tagged releases on your fork (not sure I do either) and should resolve once the PR is merged.

I think it would be better to link to the README file for configure flags and env vars so that we don't need to maintain two sets of documentation.  The build instructions should probably also start with running `autogen.sh` if the configure script is not present (this is the case pretty much all of the time right now).

davidozog:
What version badge are you seeing?  I believe that badge should be pointing to the official repo, since that's where the webhook is set up.  It should be 1.4.2 for the latest release right?

jdinan:
Weird, it shows 1.4.2 for me now.  Last time it showed ?.?.?.

davidozog:
Hmm... not a great sign.  Maybe it's a browser cache issue or something like that.  I'll keep an eye on it, and I'm also open to removing it.  It requires a https://badge.fury.io/ webhook, which may or may not be a little finicky.

davidozog:
@jdinan, I just committed the `autogen.sh` description, and linked to the text README file for all configure flags and env vars.  Please have a look at that when you have the chance.

PR #837, opened 2019-01-17
Minor cleanups to PMIx runtime code

jdinan:
* nspace field size is `PMIX_MAX_NSLEN+1`
 * npes type is `uint32_t`
 * be safer about signed/unsigned type conversions

### DIFF
--- a/src/runtime-pmix.c
+++ b/src/runtime-pmix.c
@@ -49,7 +49,7 @@ shmem_runtime_init(void)
         return rc;
     }
 
-    (void)strncpy(proc.nspace, myproc.nspace, PMIX_MAX_NSLEN+1);
+    PMIX_LOAD_NSPACE(proc.nspace, myproc.nspace);
     proc.rank = PMIX_RANK_WILDCARD;
 
     if (PMIX_SUCCESS == (rc = PMIx_Get(&proc, PMIX_JOB_SIZE, NULL, 0, &val))) {
@@ -188,7 +188,7 @@ shmem_runtime_get(int pe, char *key, void *value, size_t valuelen)
     memset(value, 0, valuelen);
 
     /* setup the ID of the proc whose info we are getting */
-    (void)strncpy(proc.nspace, myproc.nspace, PMIX_MAX_NSLEN+1);
+    PMIX_LOAD_NSPACE(proc.nspace, myproc.nspace);
 
     shmem_internal_assert(pe >= 0);
     proc.rank = (uint32_t) pe;
--- a/src/runtime-pmix.c
+++ b/src/runtime-pmix.c
@@ -21,6 +21,7 @@
 
 #include <stdlib.h>
 #include <stdio.h>
+#include <stdint.h>
 #include <string.h>
 #if defined(PMI_PORTALS4)
 #include <portals4/pmi.h>
@@ -33,7 +34,7 @@
 #include "uthash.h"
 
 static pmix_proc_t myproc;
-static size_t size;
+static uint32_t size;
 
 int
 shmem_runtime_init(void)
@@ -48,7 +49,7 @@ shmem_runtime_init(void)
         return rc;
     }
 
-    (void)strncpy(proc.nspace, myproc.nspace, PMIX_MAX_NSLEN);
+    (void)strncpy(proc.nspace, myproc.nspace, PMIX_MAX_NSLEN+1);
     proc.rank = PMIX_RANK_WILDCARD;
 
     if (PMIX_SUCCESS == (rc = PMIx_Get(&proc, PMIX_JOB_SIZE, NULL, 0, &val))) {
@@ -105,14 +106,14 @@ shmem_runtime_abort(int exit_code, const char msg[])
 int
 shmem_runtime_get_rank(void)
 {
-    return myproc.rank;
+    return (int) myproc.rank;
 }
 
 
 int
 shmem_runtime_get_size(void)
 {
-    return size;
+    return (int) size;
 }
 
 // static void opcbfunc(pmix_status_t status, void *cbdata)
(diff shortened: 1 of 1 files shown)


src/runtime-pmix.c
```diff
@@ -48,7 +49,7 @@ shmem_runtime_init(void)
         return rc;
     }
 
-    (void)strncpy(proc.nspace, myproc.nspace, PMIX_MAX_NSLEN);
+    (void)strncpy(proc.nspace, myproc.nspace, PMIX_MAX_NSLEN+1);
```
davidozog: Yeah, this `+1` does kinda make sense to me - it's for the NULL terminator, correct?

Might want to see if @rhc54 has any thoughts.  If I grep the PMIx examples with:
```
grep -r "strncpy.*nspace.*PMIX_MAX_NSLEN*" * 
```
I see a lot of occurrences, but none with the `+1`.  Does it matter?

rhc54: It would be safer to use the``` PMIX_LOAD_NSPACE``` macro to ensure everything is correct (i.e., fits in space and is NULL terminated). The ```+1``` is for the latter purpose - if we tell people "here is the max name length", then we have to ensure there is one more spot for the NULL terminator.

davidozog: Thanks @rhc54.  I don't see `PMIX_LOAD_NSPACE` in the 3.0 spec.  Is it new?

rhc54: Ah, indeed - it is in the 3.1 spec. It was overlooked when we wrote the 3.0 version and so was considered a correction as opposed to an addition. See https://pmix.org/wp-content/uploads/2019/01/pmix-standard-3.1.pdf for the updated spec.

hppritcha:
bot:lanl:retest

hppritcha:
see if moving to a new jenkins wkrdir is in effect
bot:lanl:retest

hppritcha:
try again
bot:lanl:retest

hppritcha:
bot:lanl:retest

src/runtime-pmix.c
jdinan: Updated to use `PMIX_LOAD_NSPACE` in https://github.com/Sandia-OpenSHMEM/SOS/pull/837/commits/3b11877e256e3fa3613abb06ba516ce3c8b3a66b.

jdinan:
@davidozog Ready for re-review.

rhc54:
Couldn't resist approving! Just FWIW: for backwards compatibility, I have been adding the following lines to a header file in PRRTE, just so I can support older PMIx versions:
```c
#ifndef PMIX_LOAD_NSPACE
/**
 * Provide a safe version of strncpy that doesn't generate
 * a ton of spurious warnings. Note that not every environment
 * provides nice string functions, and we aren't concerned about
 * max performance here
 *
 * @param dest Destination string.
 * @param src Source string.
 * @param len Size of the dest array - 1
 *
 */
static inline void pmix_strncpy(char *dest, const char *src, size_t len)
{
    size_t i, k;
    char *new_dest = dest;

    /* use an algorithm that also protects against
     * non-NULL-terminated src strings */
    for (i=0, k=0; i <= len; ++i, ++src, ++new_dest) {
        ++k;
        *new_dest = *src;
        if ('\0' == *src) {
            break;
        }
    }
    dest[k-1] = '\0';
}

#define PMIX_LOAD_NSPACE(a, b)                      \
    do {                                            \
        memset((a), 0, PMIX_MAX_NSLEN+1);           \
        pmix_strncpy((a), (b), PMIX_MAX_NSLEN);     \
    }while(0)
#endif

```

jdinan:
@rhc54 You might get better performance if you throw `restrict` on the `dest` and `src` arguments.  I think you can also simplify the loop a bit (sorry, can't help my ocd):
```
    for (i=0; i <= len; ++i, ++src, ++dest) {
        *dest = *src;
        if ('\0' == *src) {
            break;
        }
    }
    *dest = '\0';
```

rhc54:
Performance isn't really a consideration as PMIx isn't in the critical path. However, if you care to make a specific suggestion, I'm happy to take advantage of it 😄

jdinan:
Are the iteration bounds correct here?  If I'm reading this correctly, the code will make two trips through the loop body when `len == 1`.

jdinan:
The comment is also confusing, `@param len Size of the dest array - 1`.  Why "- 1"?

jdinan:
I guess this is to ensure that `dest` is null terminated?  In which case, the loop should not execute the body even once when `len = 1`.

rhc54:
In PMIx, the max length of the `name` was set to a defined constant, which means that the `array` has to be one greater. Rather than having everyone remember to pass `constant-1` to the function, I opted to let them just pass `constant`. FWIW: we wound up defining macros and new abstract data types just to help people not make the mistake of defining an array to be length `constant` instead of `constant+1`. I agree this isn't a more generalized solution, but seems easier for people to get right for PMIx.

The whole point of creating this function was to guarantee the result fit within the bounds _and_ included NULL termination - otherwise, we would have just stuck with the libc version.

jdinan:
The lengths used by PMIx seem perfectly reasonable; I've seen the same convention used in other APIs.  What I'm asking about is whether the implementation of `pmix_strncpy` is consistent with the PMIx length convention.  For example:
```
#define PMIX_LOAD_NSPACE(a, b)                      \
    do {                                            \
        memset((a), 0, PMIX_MAX_NSLEN+1);           \
        pmix_strncpy((a), (b), PMIX_MAX_NSLEN);     \
    }while(0)
```
IIUC, `pmix_strncpy` should do exactly what libc `strncpy` would do here, which is copy up to `PMIX_MAX_NSLEN` characters from `b` to `a` and not guarantee NULL termination.  However, the doxygen comment says it should copy one less and the implementation can actually copy one more (possibly overwriting the NULL set by `memset` twice, once with a character from `b` and a second time with a new NULL).

PR #839, opened 2019-01-24
travis: misc prrte related changes

ggouaillardet:
- use distro provided `libevent` and `hwloc` to reduce build time
- use `prun -terminate` in order to kill the `prte` daemon (instead of `killall prte`)
- use `prun --daemonize` in order to start the `prte` daemon (instead of `prte&; sleep2`)


lanl-ompi:

ggouaillardet:
as far as PMIx and prrte are concerned, these versions are good enough.
you might be able to get more up-to-date versions simply by updating the travis config to run on `xenial` (aka ubuntu 16.04.5 LTS) instead of older `trusty` (aka ubuntu 14.04.5 LTS).

some extra time can be saved by using tarballs instead of git clone (faster download and no need to run `autogen.sh`). PMIx 3.1.2 can be used now, but I'd rather wait for the next `prrte` release.

Issue #685, opened 2018-05-01
Configure fails to find XPMEM on Fedora 27


tonycurtis:
XPMEM installed to /opt/xpmem:
```
/opt/xpmem
|-- etc
|   |-- 56-xpmem.rules
|   `-- init.d
|       `-- xpmem
|-- include
|   `-- xpmem.h
`-- lib
    |-- libxpmem.a
    |-- ... ...
```
Configure fails to link -lxpmem (thus reporting no support for XPMEM at the end), looking only in .../lib64
```
configure:16256: checking for xpmem.h
configure:16256: result: yes
configure:16270: checking for xpmem_init in -lxpmem
configure:16303: gcc -o conftest -std=gnu11 -O3 -ggdb -Wall  -I/opt/xpmem/includ
e  -L/opt/xpmem/lib64 conftest.c -lxpmem   >&5
/usr/bin/ld: cannot find -lxpmem
collect2: error: ld returned 1 exit status
configure:16303: $? = 1
```

davidozog:
This happens when passing `--with-xpmem=/opt/xpmem`?

tonycurtis:
Yep.  XPMEM is
```
./configure --prefix=/opt/xpmem
```
and SOS is i.a.
```
--with-xpmem=/opt/xpmem
```
Looks like I had the same issue on CentOS 7 but fudged it with a symlink.

tonycurtis:
Yeah, that typo is probably it.  Configure's actually checking for "/lib64", which succeeds, and it never tries .../lib

PR #841, opened 2019-01-28
Travis: Update to xenial

jdinan:
Yay!!  --- https://blog.travis-ci.com/2018-11-08-xenial-release

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,16 +46,19 @@ env:
           SOS_BUILD_OPTS="--disable-fortran --enable-error-checking --enable-remote-virtual-addressing --disable-aslr-check --enable-pmi-simple"
         - >
           SOS_ENABLE_ERROR_TESTS=1 SHMEM_SYMMETRIC_HEAP_USE_MALLOC=1
-          SOS_TRANSPORT_OPTS="--with-portals4=$TRAVIS_INSTALL/portals4/" 
+          SOS_TRANSPORT_OPTS="--with-portals4=$TRAVIS_INSTALL/portals4/"
           SOS_BUILD_OPTS="--disable-threads --enable-error-checking --enable-pmi-simple"
         - >
           SOS_DISABLE_FORTRAN=1
-          SOS_TRANSPORT_OPTS="--with-portals4=$TRAVIS_INSTALL/portals4/" 
+          SOS_TRANSPORT_OPTS="--with-portals4=$TRAVIS_INSTALL/portals4/"
           SOS_BUILD_OPTS="--with-cma --enable-error-checking --enable-profiling --enable-remote-virtual-addressing --enable-pmi-simple"
+        - >
+          SOS_TRANSPORT_OPTS="--with-portals4=$TRAVIS_INSTALL/portals4/"
+          SOS_BUILD_OPTS="--with-xpmem=$TRAVIS_INSTALL/xpmem --enable-error-checking --enable-pmi-simple"
         - >
           SOS_ENABLE_ERROR_TESTS=1
           SHMEM_SYMMETRIC_HEAP_USE_HUGE_PAGES=1 SHMEM_BOUNCE_SIZE=0
-          SOS_TRANSPORT_OPTS="--with-portals4=$TRAVIS_INSTALL/portals4/" 
+          SOS_TRANSPORT_OPTS="--with-portals4=$TRAVIS_INSTALL/portals4/"
           SOS_BUILD_OPTS="--enable-error-checking --enable-remote-virtual-addressing --enable-pmi-simple --enable-ofi-fence"
 
         ## OFI Builds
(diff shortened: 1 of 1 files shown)


jdinan:
Well, this is working now.  We have 1:30 margin on the CMA row in the testing matrix.  But, we also had to make some concessions in terms of testing coverage to get there.  Should we merge this?

davidozog:
I think it's ok to merge, especially if these tests are covered by Jenkins, which I believe they are.  Is that the case @wrrobin?

jdinan:
@davidozog @wrrobin This should be ready for review.  I refactored large parts of the testing script to handle OFI, Portals 4, and PMIx through separate rows in the testing matrix.  There should be no meaningful reduction in coverage now and hopefully should now be much easier to add to the build matrix.  Also learned a painful lesson about libev/libevent along the way.

jdinan:
Also, these changes leave enough headroom that we can let the full CMA testing row run without any hacks to reduce testing time.

jdinan:
One last update to add XPMEM testing to the Travis script.

jdinan:
Cleared testing, xpmem and all!  :metal:

jdinan:
@davidozog @wrrobin This is ready for review (promise to stop tinkering now).  It's a long read, so I'm happy to walk you through it at tomorrow's team meeting.

Portals, CMA, and PMIx now have separate rows in the build matrix (they used to be mixed in with every build).  This sped up the individual rows substantially.  Splitting things out and getting dependencies from the package manager shortened the logs a lot, so we should not need to view the raw logs.  Portals UDP build is still too buggy to run the tests; it quits the testing cycle early.  I also added XPMEM to the build matrix (yay!).

.travis.yml
```diff
@@ -107,41 +172,52 @@ before_install:
...
-    - wget http://www.mpich.org/static/downloads/3.2/hydra-3.2.tar.gz
-    - tar xvzf hydra-3.2.tar.gz
-    - cd hydra-3.2/
-    - ./configure --prefix=$TRAVIS_INSTALL/hydra
-    - make $TRAVIS_PAR_MAKE
-    - make install
+    - >
+      if [[ $SOS_TRANSPORT_OPTS = *"with-ofi"* ]]; then
+        cd $TRAVIS_SRC
+        git clone -b v1.6.x --depth 10 https://github.com/ofiwg/libfabric.git libfabric
```
davidozog: Should we move to v1.7.x?  I seem to recall we've discussed doing this.

.travis.yml
```diff
@@ -107,41 +172,52 @@ before_install:
...
+        ## Install XPMEM
+        cd $TRAVIS_SRC
+        git clone --depth 10 https://github.com/hjelmn/xpmem.git
+        cd xpmem
+        ./autogen.sh
+        ./configure --prefix=$TRAVIS_INSTALL/xpmem
+        make $TRAVIS_PAR_MAKE
+        make install
+        ## Hack to get SOS to see the XPMEM library
+        ln -s $TRAVIS_INSTALL/xpmem/lib $TRAVIS_INSTALL/xpmem/lib64
```
davidozog: I wonder what the cause of this is... I swear I've had to do this before, but I cannot remember which library it was.  Is it a libtool bug?

.travis.yml
jdinan: Sure, mind if I do that as a follow-up PR?

.travis.yml
jdinan: SOS configury looks for the library in the $WITH_XPMEM/lib64.  🤦‍♂️

wrrobin:
@jdinan All the changes look good to me. Also, a discussion tomorrow on the changes will be helpful as well. Looking at the last round of tests, the first few of them did run 3x faster compared to others. Is there any particular reason?

.travis.yml
jdinan: Ref #685

jdinan:
@wrrobin Those are the Portals tests.  They build and check SOS, then quit before running any of the external tests (UH, Cray, Mellanox, Parres, etc.).

.travis.yml
davidozog: Of course.

.travis.yml
davidozog: Haha!  I thought I'd seen this before.  Maybe that library was actually xpmem...  😖

wrrobin:
Ok. I missed that they did only build and not ran any test.

davidozog:
Now that I think of it, I'm surprised Portals over sockets used to pass all those test suites but not SOS tests...  Maybe we want to keep one (or more) of them?

wrrobin:
The current setup is fine with me though. We can add more Portals tests once we need to check them for any reason. Adding any one also is ok.

jdinan:
The old Travis script built the external tests with Portals, but did not run them.

davidozog:
Ah yes.  I see now that none of the suites were actually run with Portals.  I was misled by the old comment, "`Run the UH test suite (Portals)`"...

jdinan:
Yeah, those comments were very aspirational.  If we ever get librxe (soft RoCE) working, we might be able to run the verbs build of Portals in Travis.  (I checked, it's not a simple package install in Xenial)

jdinan:
Sounds like @davidozog is good with merging.  @wrrobin Go ahead or wait for tomorrow?

wrrobin:
@jdinan Sorry, i thought i approved it earlier.

PR #842, opened 2019-02-01
Update NEWS file for 1.4.3rc1

jdinan:

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -8,6 +8,9 @@ v1.4.3rc1
 - Added proposed OpenSHMEM wait/test-any/some/all to shmemx interfaces.
 - Added proposed OpenSHMEM nonblocking fetching atomic operations to shmemx
   interfaces.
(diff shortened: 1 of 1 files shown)


NEWS
```diff
@@ -8,6 +8,9 @@ v1.4.3rc1
 - Added proposed OpenSHMEM wait/test-any/some/all to shmemx interfaces.
 - Added proposed OpenSHMEM nonblocking fetching atomic operations to shmemx
   interfaces.
+- Added proposed OpenSHMEM SHMEM_CTX_INVALID constant to shmemx interfaces.
```
wrrobin: Should we say `SHMEMX_CTX_INVALID`? Although we did mention as shmemx interface. Both ok, I guess.

jdinan: Yeah, I went back and forth on that, too.

jdinan: :man_shrugging:

wrrobin: I am ok then with whatever you decide :)

PR #845, opened 2019-02-13
Update simple build to SOS v1.4.3 and OFI v1.7.0

jdinan:


jdinan:
@davidozog @wrrobin Merging without review...

PR #850, opened 2019-02-21
[nomerge] Posted for testing

jdinan:


jdinan:
Friends don't let friends combine libev and libevent.

PR #854, opened 2019-02-26
Travis: Test with libfabric 1.7.x

jdinan:


jdinan:
This checks out fine.  I'm going to stop testing here and merge, so Travis can get on to the next PR.

PR #856, opened 2019-02-28
OFI: Add support for vanilla (no STX) endpoints

jdinan:


jdinan:
@davidozog Ready for review.

src/transport_ofi.c
```diff
@@ -542,8 +544,11 @@ int bind_enable_ep_resources(shmem_transport_ctx_t *ctx)
 {
     int ret = 0;
 
-    ret = fi_ep_bind(ctx->ep, &shmem_transport_ofi_stx_pool[ctx->stx_idx].stx->fid, 0);
-    OFI_CHECK_RETURN_STR(ret, "fi_ep_bind STX to endpoint failed");
+    /* If using SOS-managed STXs, bind the STX */
+    if (ctx->stx_idx >= 0) {
```
davidozog: I thought `stx_idx` was guaranteed to be `>=0` after `shmem_transport_ofi_stx_allocate` (as in the [Note here](https://github.com/Sandia-OpenSHMEM/SOS/blob/v1.4.3/src/transport_ofi.c#L480)).

So I don't think this check is necessary, right?  But I'm ok with it.

davidozog:
Whoops, just saw that last commit.  It looks good too.

src/transport_ofi.c
jdinan: That changes with this PR.  `shmem_transport_ofi_stx_allocate` will return -1 if there are no STXs.

davidozog: Gotcha.  I was just surprised this is the only check on that condition.  It was already there before this PR in `ctx_destroy`, strangely, and it looks like that indexing doesn't show up anywhere else...

So yeah, thanks for the comment clarification - this PR is good to merge.

PR #859, opened 2019-03-05
Travis: Enable PMIx and Portals 4, both with libev

jdinan:


jdinan:
Thanks, @ggouaillardet!  💯

Issue #858, opened 2019-03-05
ICC is broken in Travis/Xenial


jdinan:
The Intel C compiler is broken in Travis after the update to Xenial.

Might need to be updated to a newer version (icc indicated the OS is unsupported).

Ref #855

jdinan:
Might be something useful we can crib from here: https://github.com/jeffhammond/HPCInfo/blob/master/buildscripts/icc-release.sh

PR #860, opened 2019-03-06
Travis: Disable icc testing until fixed

jdinan:
Ref: #858


jdinan:
@davidozog Are you ok with this as a temporary change?

PR #861, opened 2019-03-08
Add target CQ and use for manual progress

jdinan:
With this patch, manual progress no longer requires the target cntr and can be independent of hard polling in wait operations.

### DIFF
--- a/src/transport_ofi.c
+++ b/src/transport_ofi.c
@@ -58,6 +58,7 @@ struct fid_fabric*              shmem_transport_ofi_fabfd;
 struct fid_domain*              shmem_transport_ofi_domainfd;
 struct fid_av*                  shmem_transport_ofi_avfd;
 struct fid_ep*                  shmem_transport_ofi_target_ep;
+struct fid_cq*                  shmem_transport_ofi_target_cq;
 #if ENABLE_TARGET_CNTR
 struct fid_cntr*                shmem_transport_ofi_target_cntrfd;
 #endif
@@ -1232,6 +1233,16 @@ static int shmem_transport_ofi_target_ep_init(void)
     ret = allocate_recv_cntr_mr();
     if (ret != 0) return ret;
 
+     struct fi_cq_attr cq_attr = {0};
+
+     ret = fi_cq_open(shmem_transport_ofi_domainfd, &cq_attr,
+                      &shmem_transport_ofi_target_cq, NULL);
+     OFI_CHECK_RETURN_MSG(ret, "cq_open failed (%s)\n", fi_strerror(errno));
+
+     ret = fi_ep_bind(shmem_transport_ofi_target_ep,
+                      &shmem_transport_ofi_target_cq->fid, FI_TRANSMIT|FI_RECV);
+     OFI_CHECK_RETURN_STR(ret, "fi_ep_bind CQ to target endpoint failed");
+
     ret = fi_enable(shmem_transport_ofi_target_ep);
     OFI_CHECK_RETURN_STR(ret, "fi_enable on target endpoint failed");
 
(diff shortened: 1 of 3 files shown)


src/transport_ofi.c
```diff
@@ -1216,7 +1224,7 @@ static int shmem_transport_ofi_target_ep_init(void)
 #if ENABLE_TARGET_CNTR
     info->p_info->caps |= FI_RMA_EVENT;
 #endif
-    info->p_info->tx_attr->op_flags = FI_DELIVERY_COMPLETE;
+    info->p_info->tx_attr->op_flags = 0;
```
wrrobin: @jdinan One question. Does this mean we have to add `FI_DELIVERY_COMPLETE` on the other places where we assumed previously that it is already or'ed in the TX attribute?

jdinan: That's a good question.  I have always disliked modifying these fields in-place.  It would be much safer to memcpy to a temporary info structure, but deep copies are a pain..

davidozog: @wrrobin Thanks for catching that - it's a very good question.

wrrobin: @jdinan Yes, I agree. I was thinking that it is no longer required to have `FI_DELIVERY_COMPLETE` after seeing this update. But, based on the manpages, that should not be true or I am missing something.  Was there any other reason to reset it to 0?

davidozog: Here's my understanding, which could be wrong:

The target EP doesn't do any TX operations, so I assume that we want this flag to be zero (setting to `FI_DELIVERY_COMPLETE` may have been a mistake I made a while ago in commit https://github.com/Sandia-OpenSHMEM/SOS/commit/2fc798d442bca4cb3b79ab6b5673789a4de85d8d).  We also have to set it to 0 explicitly here, because it may have been set to `FI_DELIVERY_COMPLETE` in `ctx_init`.

But, since this EP doesn't transmit, I'd guess it doesn't make a difference in testing...

wrrobin: According to the manpages,
```
FI_DELIVERY_COMPLETE :Indicates that a completion should not be generated until an operation has been processed by the destination endpoint(s). Acompletion guarantees that the result of the operation is available. This completion mode applies only to reliable endpoints. Foroperations that return data to the initiator, such as RMA read or atomic−fetch, the source endpoint is also considered a destination endpoint. This is the default completion mode for such operations.
```
I thought the `FI_DELIVERY_COMPLETE` also makes that `quiet` ensuring remote completion. Without keeping it in the target endpoint, can we say that quiet ensures remote completion?

davidozog: I don't believe the target EP (`shmem_transport_ofi_target_ep` and it's associated CQ) is involved in `quiet` right?  We only need `FI_DELIVERY_COMPLETE` on the transmitting EP within each `ctx`.  That manpage says this flag generates a completion event after an operation is processed by the destination endpoint, but the target EP never has a destination endpoint...

wrrobin: Thanks @davidozog. I understand that the `target_ep` is not involved in `quiet`, but my concern was that we are making this change on `shmem_transport_ofi_info`, the `fabric_info` object. Is this ok (regarding delivery completion in quiet) to make this change here and keep line number 1290 as it is?

davidozog: Ah ok, sorry.  Yeah, we're on the same page - I'm not 100% about that either.  My hope is that the `info` object is only read by libfabric when it's passed to a routine, like `fi_endpoint` in this case; and I would also hope the EP within libfabric has internal state with params such as this...?  But I agree there must be a better way to set `info` params with a deep copy as Jim mentioned.

jdinan: It looks like we set `tx_attr->op_flags` back to `FI_DELIVERY_COMPLETE` in `shmem_transport_ofi_ctx_init`, so I this change should not cause an issue.  Any fields that are changed are currently changed in both places.

wrrobin: Thanks @jdinan. So, while at `trasport_init` we keep the flags to 0 and add `FI_DELIVERY_COMPLETE` when we are creating the default context in `transport_startup`. I wonder whether the latter change has any impact on the former one inside libfabric. Otherwise, I see that it should be ok.

jdinan: Sean confirmed that `FI_DELIVERY_COMPLETE` does not have a meaning for receives, so it's safe to remove here.

PR #862, opened 2019-03-20
Add support for decimal values in SHMEM_SYMMETIC_SIZE

jdinan:
Implements proposed extension: https://github.com/openshmem-org/specification/pull/200

Does not return an error value from `shmem_init_thread` as suggested in the proposal.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -322,6 +322,8 @@ AC_SEARCH_LIBS([clock_gettime], [rt],
 CHECK_SCHED_GETAFFINITY(
     [AC_DEFINE([HAVE_SCHED_GETAFFINITY], [1], [define if sched_getaffinity is available])], [])
 
+AC_SEARCH_LIBS([ceil], [m], [], AC_MSG_ERROR([unable to find ceil() function]))
+
 dnl check for libraries
 
 OMPI_CHECK_PORTALS4([portals4],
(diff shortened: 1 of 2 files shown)


jdinan:
@davidozog @wrrobin What do you think?  Is the proposal mature enough to merge this?

davidozog:
We decided that `shmem_init_thread` will return non-zero on invalid input, right?  At a glance it seems like we don't do that, but I'm having a tough time tracking it down...  I see that `atol_scaled` is called by `shmem_internal_getenv_size`, but where is that called?

Anyway, I think it's ok to merge this as long as there's an open an issue to address the `init_thread` return value (if that's even an issue).

wrrobin:
I am having the same trouble as @davidozog. Not able to understand where `shmem_internal_getenv_size` is called. Since it seems to be never called, the changes are definitely ok to merge. But, I wonder, how could it matter? May be, something I missed?

jdinan:
Travis: Need to add `-lm` to `LDFLAGS` or find a different way to perform the ceiling operation.

jdinan:
The calls to this function are in code generated by the C preprocessor.  Look for `shmem_internal_getenv_##KIND` in `shmem_internal_parse_env`at shmem_env.c:156.  Currently, we abort immediately rather than returning from `shmem_init_thread`.  It's a fair amount of work to change this, but fortunately aborting is allowed by the proposal.

Issue #866, opened 2019-04-26
seg fault in shmem_global_exit()


davidozog:
I'm getting a seg fault on Mac OSX with clang when calling `shmem_global_exit()`.
My call stack:

```
shmem_util_backtrace at backtrace.c:184:21
shmem_runtime_abort(exit_code=1, msg="PE 0 called shmem_global_exit with status 1") at runtime-pmi.c:139:5
shmem_internal_global_exit(status=1) at init.c:502:5
shmem_global_exit(status=1) at init_c.c:98:5
main(argc=1, argv=0x00007ffeefbff310) at shmemx_team_translate.c:25:9
```
I think my debugger is saying that 
`
char *method = shmem_internal_params.BACKTRACE;
`
results in `method` as a `NULL` pointer (though I would expect it should be an empty string), so `strcmp` seg faults.

davidozog:
Also seeing this in Ubuntu 16.04 with gcc.

wrrobin:
@davidozog Thanks for catching this and sorry that I missed your issue submission earlier.

davidozog:
@wrrobin No problem!  It was my fault, as explained in PR #867

PR #867, opened 2019-04-29
Fix global_exit seg fault when no backtracing method is available.

davidozog:
This is a quick fix for #866.

### DIFF
--- a/src/backtrace.c
+++ b/src/backtrace.c
@@ -181,6 +181,8 @@ void backtrace_gdb(void) {
 void shmem_util_backtrace(void) {
     char *method = shmem_internal_params.BACKTRACE;
 
+    if (!method) return;
+
     if (0 == strcmp(method, "execinfo")) {
 #if defined(USE_BT_EXECINFO)
         backtrace_execinfo();


wrrobin:
@davidozog Thanks for the PR. Although this does fix the problem, I am thinking to assign a default value other than "" to `BACKTRACE`. If we want to keep it disabled by default, then may be we can use a default value of "disabled" or "N/A". We have maintained a default value for all other string type variables. This way, it will be consistent as well. 

@jdinan @davidozog Your thoughts?

jdinan:
Hang on, something is wrong.  The default for this env var is `""`, so `shmem_internal_getenv_string` should assign a reference to the empty string (i.e. not `NULL`).  Is the env var being touched prior to the call to `shmem_internal_parse_env`?

jdinan:
@wrrobin `SHMEM_BACKTRACE` appears to be the only string env var where the default value is the empty string.  A better default is not a bad idea, but there is a non-NULL default value.

davidozog:
Yeah,  I have no idea why `method` is `NULL` - it should be an empty string.

It looks like this might be unique to my WIP branch.  On my end, the env var is read within `shmem_global_exit`, and long after `shmem_init`.  But, I just checked the master branch and see that `method` is `\0`, so perhaps we can close this PR...

davidozog:
I found my problem.  I had removed `#include "config.h"` from `shmem_env.h` for no good reason (I saw it was in `shmem_env.c`, so I flippantly removed it from the header...)  Sorry for the false alarm.  It's kinda hard to diagnose these issues caused by not including `config.h`...

wrrobin:
Thanks @jdinan @davidozog. Good to know that it got fixed.

PR #870, opened 2019-06-06
Enable shmem_init_thread to return error on bad SHMEM_SYMMETRIC_SIZE

jdinan:


jdinan:
The PMI/OFI row in Travis is timing out.  Seems to be a problem with PRRTE hanging again ...

jdinan:
@davidozog Could you review this PR?  I'll keep looking into the issue with PRRTE in the meantime.  Thanks!

src/init.c
```diff
@@ -145,14 +145,16 @@ shmem_internal_shutdown_atexit(void)
...
 shmem_internal_start_pes(int npes)
 {
-    int tl_provided;
+    int ret, tl_provided;
 
     shmem_internal_initialized_with_start_pes = 1;
-    shmem_internal_init(SHMEM_THREAD_SINGLE, &tl_provided);
+    ret = shmem_internal_init(SHMEM_THREAD_SINGLE, &tl_provided);
+
+    if (ret) abort();
```
jdinan: FWIW, I didn't see a better way to abort here.  The runtime isn't initialized, so we can't use that to abort.  Init would have called `abort()` in this case, so that's what I did here.  One thought was adding a new `shmem_internal_abort()` that can choose how to abort based on the state of the library.

jdinan:
I'm not seeing an easy way to address the PMIx/PRRTE hang in testing.  It appears to be an upstream issue.  I suggest we just ignore it for now and queue it up for deeper investigation when time permits.

jdinan:
Added #871 as a reminder to improve how SOS aborts.

PR #873, opened 2019-06-27
Don't deref null fi_cntr pointer in debug msg

jdinan:
When `ctx_destroy` is called to clean up a failed `ctx_init` operation, the debug message could dereference null pointers to the FI cntrs on the context.  Add check before dereferencing these pointers.


jdinan:
@davidozog Should be a quick review.

PR #875, opened 2019-06-29
Travis: check for missing pshmem symbols

jdinan:
Bit of a hack, but this should catch it if we are missing any pshmem functions in the future.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -303,7 +303,7 @@ script:
     ###
     - >
       if [[ $SOS_TRANSPORT_OPTS = *"enable-profiling"* ]]; then
-      missing_pshmem=$( nm src/.libs/libsma.so | grep shmem_ | grep -v -e shmem_internal -e shmem_transport -e shmem_free_list -e shmem_runtime -e shmem_util -e shmem_spinlock -e shmem_op -e '_$' | sed -e 's/^.* . //' | perl -e 'while (<>) { if (/^p/) { $h{substr $_,1} -= 1; } else { $h{$_} += 1 } } for $k (keys %h) { if ($h{$k}) { print $k; ++$f; } } die "Missing profiling symbols" unless $f == 0;' )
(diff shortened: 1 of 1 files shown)


jdinan:
@hppritcha Looks like LANL-CI is having trouble cloning the SOS repository.

jdinan:
@davidozog @wrrobin What do you think of this PR?

wrrobin:
Looks good to me @jdinan

davidozog:
It may be my fault, but when I try this on my teams branch, I get a bunch of `shmem_op_*` symbols from `shmem_internal_op.h` that have no pshmem definitions.  So should we add `shmem_op` to this grep filter?

jdinan:
@davidozog Those op functions should be static inline, so I wouldn't expect to see them in the built library.  We might also want to change the name to `shmem_internal_op*` to follow our naming convention.

jdinan:
Scratch that, I see the op functions showing up, too ...

jdinan:
It's a bit longer, but here's one solution that uses Perl instead of `sort` and `uniq`:
`nm src/.libs/libsma.so | grep shmem_ | grep -v -e shmem_internal -e shmem_transport -e shmem_free_list -e shmem_runtime -e shmem_util -e shmem_spinlock -e shmem_op -e '_$' | sed -e 's/^.* . //' | perl -e 'while (<>) { if (/^p/) { $h{substr $_,1} -= 1; } else { $h{$_} += 1 } } for $k (keys %h) { if ($h{$k}) { print $k; } }'`

jdinan:
Here's an explanation for the Perl script:
```
// Read a line of input into $_
while (<>) {
  // if the string begins with p, subtract one from the corresponding
  // entry in the hash (with the leading p removed by substr)
  if (/^p/) { $h{substr $_,1} -= 1; }
  // otherwise increment the entry in the hash
  else { $h{$_} += 1 }
}

// Iterate over entries in the hash and print any that are nonzero
for $k (keys %h) { if ($h{$k}) { print $k; ++$f; } }
die "Missing profiling symbols" unless $f == 0;
```

PR #878, opened 2019-07-18
Travis: Use PMIx v3.1.3 and PRRTE v1.0.0

jdinan:


wrrobin:
Looks like the Travis tests could not locate `-levent`.

jdinan:
It appears to be an issue with PMIx v3.1.3.  I switched back to the PMIx master and all is well again.

Ref: @rhc54 FYI a PMIx release that picks up libev support would also be helpful for us.

PR #828, opened 2019-01-14
Update node/shr rank API

jdinan:
Update node/shr rank API to support both on-node comms and node-level resource management.  This PR incorporates and supersedes #792.  Some notable changes relative to #792:

 * Merge node-util API with runtime API
 * Introduce runtime-util API for common utility functions used by different runtimes
 * Move "shr" rank query routines into "shmem_internal" API
 * Remove SHMEM_OFI_NODE_STX_MAX env var to simplify first version of automatic STX management

PR #884, opened 2019-09-18
SOS v1.4.4 updates

jdinan:
- [x] Coverity Scan (0 new defects, 1 eliminated)
- [x] Test with PSM2
- [x] Test with RXM/Verbs
- [x] Test with Portals 4
- [x] Post-release: Update sos-tests (https://github.com/openshmem-org/tests-sos/pull/13/)

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -6,9 +6,13 @@ v1.4.4
 ------
 - Experimental support for RXM/Verbs provider stack with libfabric 1.8.
   Requires --enable-hard-polling and --enable-ofi-mr=basic build flags.  The
-  FI_MR_CACHE_MAX_COUNT=0 environment variable should be set when threading is
-  used.  Depending on the libfabric build, SHMEM_OFI_PROVIDER="verbs;ofi_rxm"
-  may also need to be set to select the provider.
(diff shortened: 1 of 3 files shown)


NEWS
```diff
@@ -2,6 +2,34 @@
 
 Sandia OpenSHMEM NEWS -- history of user-visible changes.
 
+v1.4.4
+------
+- Experimental support for RXM/Verbs provider stack with libfabric 1.8.
+  Requires --enable-hard-polling and --enable-ofi-mr=basic build flags.  The
+  FI_MR_CACHE_MAX_COUNT=0 environment variable should be set when threading is
```
wrrobin: Should we re-write "`FI_MR_CACHE_MAX_COUNT` environment variable should be set to 0"?

davidozog:
One item that may (or may not) be missing is the addition of the node/shared-rank API (#828), with the `OFI_STX_AUTO` feature.  Looks like we released v1.4.3 then merged #828 later that same day.

There are also extreme improvements to Travis CI, but maybe that's not appropriate for the release notes...

jdinan:
Oh wow, don't know how I missed that.

PR #888, opened 2019-09-25
Add a unit test for in-place reductions

jdinan:


jdinan:
Ready for review, @davidozog @gtaylor5

jdinan:
Correct, we could use `pWrk` to eliminate the memcpy and barrier.  It would require an additional RTR (read to receive) synchronization between receiver and sender per chunk to resolve the WAR data race, since `pWrk` would be overwritten by the transfer of each chunk.  Not hard to implement, but I chose to avoid using `pWrk` since we are `pWrk` free in the teams code.

PR #890, opened 2019-10-15
Updates to utility atomics

jdinan:
Updates to the utility atomics (spinlocks and counters) to use `__atomic` rather than deprecated `__sync`, inline assembly, and compiler fences.  Rename the thread-safe counter routines to avoid name confusion with `shmem_internal_atomic_*` API used to implement the AMOs.

### DIFF
--- a/src/shmem_atomic.h
+++ b/src/shmem_atomic.h
@@ -158,31 +158,31 @@ shmem_internal_membar_store(void) {
 
 #include <stdint.h>
 
-typedef uint64_t shmem_internal_atomic_uint64_t;
+typedef uint64_t shmem_internal_cntr_t;
 
 static inline
 void
-shmem_internal_atomic_write(shmem_internal_atomic_uint64_t *ptr, uint64_t value) {
+shmem_internal_cntr_write(shmem_internal_cntr_t *ptr, uint64_t value) {
     __atomic_store_n(ptr, value, __ATOMIC_RELEASE);
     return;
 }
 
 static inline
 uint64_t
-shmem_internal_atomic_read(shmem_internal_atomic_uint64_t *val) {
+shmem_internal_cntr_read(shmem_internal_cntr_t *val) {
     return __atomic_load_n(val, __ATOMIC_ACQUIRE);
 }
 
 static inline
 void
-shmem_internal_atomic_inc(shmem_internal_atomic_uint64_t *val) {
+shmem_internal_cntr_inc(shmem_internal_cntr_t *val) {
     __atomic_fetch_add(val, 1, __ATOMIC_RELEASE);
     return;
 }
 
 static inline
 void
-shmem_internal_atomic_dec(shmem_internal_atomic_uint64_t *val) {
+shmem_internal_cntr_dec(shmem_internal_cntr_t *val) {
     __atomic_fetch_sub(val, 1, __ATOMIC_RELEASE);
     return;
 }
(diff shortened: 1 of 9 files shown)


jdinan:
@wrrobin Might be easiest to review the individual patches, since 2b29bc2 is a simple refactor with no substantive code changes.

jdinan:
This code is leading toward a shared memory transport layer, which implements all of the AMOs using XPMEM and `__atomic`.  Hence the shift to standardizing on those APIs.

src/shmem_atomic.h
```diff
@@ -57,72 +57,68 @@ static inline
...
 
 
 static inline
 void
 shmem_spinlock_lock(shmem_spinlock_t *lock)
 {
-    long val = __sync_fetch_and_add(&lock->enter, 1);
-    while (val != (lock->exit)) {
+    long val = __atomic_fetch_add(&lock->enter, 1, __ATOMIC_ACQUIRE);
+    while (val != __atomic_load_n(&lock->exit, __ATOMIC_RELAXED)) {
```
jdinan: Looking back at this ... the unlock operation performs a release on `lock->exit`.  To get ordering, I think we want the load of `lock->exit` here to be an acquire.  This eliminates the thread fence below.

src/shmem_atomic.h
```diff
@@ -77,52 +77,47 @@ static inline
 void
 shmem_spinlock_unlock(shmem_spinlock_t *lock)
 {
-    COMPILER_FENCE();
-    lock->exit++;
+    __atomic_fetch_add(&lock->exit, 1, __ATOMIC_RELEASE);
```
wrrobin: Although `fetch` is not necessary, should we use `add_fetch` instead?

jdinan: Yeah, the ``__atomic`` builtins don't provide a non-fetching atomic add operation.  I'm not sure if fetch-add or add-fetch is faster.  🤷‍♂

src/shmem_atomic.h
```diff
@@ -77,52 +77,47 @@ static inline
...
 {
-    shmem_internal_assertp(lock->enter == lock->exit);
+    shmem_internal_assertp(__atomic_load_n(&lock->enter, __ATOMIC_ACQUIRE) ==
+                           __atomic_load_n(&lock->exit, __ATOMIC_ACQUIRE));
 }
 
 
 #if (defined(__STDC_NO_ATOMICS__) || !defined(HAVE_STDATOMIC_H))
 
+/* The full memory barrier is used in cases where global ordering is requred,
```
wrrobin: `required`

src/shmem_atomic.h
```diff
@@ -77,52 +77,47 @@ static inline
...
 shmem_internal_membar_load(void) {
-#if defined(__i386__) || defined(__x86_64__)
-    if (SHMEM_INTERNAL_NEED_MEMBAR) 
-        __asm__ __volatile__ ("lfence" ::: "memory"); 
-#else
-    if (SHMEM_INTERNAL_NEED_MEMBAR) 
-        __sync_synchronize();
-#endif
+    if (SHMEM_INTERNAL_NEED_MEMBAR)
+        __atomic_thread_fence(__ATOMIC_RELEASE);
```
wrrobin: Should'nt this be `__ATOMIC_ACQUIRE`?

src/shmem_atomic.h
```diff
@@ -77,52 +77,47 @@ static inline
...
 shmem_internal_membar_store(void) {
-#if defined(__i386__) || defined(__x86_64__)
-    if (SHMEM_INTERNAL_NEED_MEMBAR) 
-        __asm__ __volatile__ ("sfence" ::: "memory");
-#else
-    if (SHMEM_INTERNAL_NEED_MEMBAR) 
-        __sync_synchronize();
-#endif
+    if (SHMEM_INTERNAL_NEED_MEMBAR)
+        __atomic_thread_fence(__ATOMIC_ACQUIRE);
```
wrrobin: Shouldn't this be `__ATOMIC_RELEASE`?

src/shmem_comm.h
```diff
@@ -107,14 +107,14 @@ shmem_internal_put_signal_nbi(shmem_ctx_t ctx, void *target, const void *source,
...
-        shmem_internal_membar_store(); /* Memory fence to ensure target PE observes stores in the correct order */
+        /* Memory fence to ensure target PE observes stores in the correct order */
+        shmem_internal_membar_release();
         shmem_transport_xpmem_put(sig_addr, &signal, sizeof(uint64_t), pe, node_rank);
 #elif USE_CMA
         if (len > shmem_internal_params.CMA_PUT_MAX) {
             shmem_transport_put_signal_nbi((shmem_transport_ctx_t *)ctx, target, source, len, sig_addr, signal, pe);
         } else {
             shmem_transport_cma_put(target, source, len, pe, node_rank);
-            shmem_internal_membar_store(); /* Memory fence to ensure target PE observes stores in the correct order */
```
wrrobin: Trying to understand, for CMA, why don't we need to ensure the order?

src/shmem_lock.h
```diff
@@ -95,14 +95,14 @@ shmem_internal_set_lock(long *lockp)
...
+        /* Lock was acquired immediately without calling SHMEM_WAIT,
+         * which provides memory ordering. Therefore, issuing an acquire
          * fence to ensure memory ordering. */
-        shmem_internal_membar_load();
+        shmem_internal_membar_acquire();
         /* Transport level memory flush is required to make memory
-         * changes (i.e. operations performed within a previous 
+         * changes (i.e. operations performed within a previous
          * critical section) visible */
-        shmem_transport_syncmem();                                                         \
```
wrrobin: 😬 .. thanks for fixing this.

src/shmem_comm.h
jdinan: I can't find any info on this.  I was assuming that since these are system calls that cross protection domains the memory fence is not needed.  But, I can't find any reference to validate this assumption.  Might be better to add it back...

Issue #891, opened 2019-10-23
Wait/test any/all/some shmemx + tests 


davidozog:
Need to 1) remove the shmemx extension from the recently ratified wait/test any/all/some + vector APIs and 2) move the shmemx wait/test any/all/some tests into the unit tests directory.

Issue #863, opened 2019-04-01
Update SHMEMX extended wait/test to match final proposal


jdinan:

davidozog:
#907 adds a `status` array to `shmem_wait_until_all`, but we also need the same update to `shmem_test_all`.

Also, most of the other wait/test routines with a `status` array still set `status[i]` to 1 when object `i` becomes satisfied, but we'll want to remove those lines.

PR #892, opened 2019-10-23
Fix and re-arrange which shmemx tests are built

davidozog:
This reminds me that the shmemx wait/test any/all/some tests should be moved out of shmemx.  I created issue #891 (and see #863 and #892). 

@jdinan, please let me know if you'd rather address those issues before merging this PR.

### DIFF
--- a/test/shmemx/Makefile.am
+++ b/test/shmemx/Makefile.am
@@ -11,15 +11,9 @@
 # information, see the LICENSE file in the top level directory of the
 # distribution.
 
-check_PROGRAMS = \
-	shmemx_wait_until_all \
-	shmemx_wait_until_any \
-	shmemx_wait_until_some \
-	shmemx_test_all \
-	shmemx_test_any \
-	shmemx_test_some \
-	c11_test_shmemx_wait_until \
-	c11_test_shmemx_test
+check_PROGRAMS =
+
+if SHMEMX_TESTS
 
 if ENABLE_PROFILING
 check_PROGRAMS += \
@@ -31,23 +25,27 @@ check_PROGRAMS += \
 	shmem_ct
 endif
 
-if SHMEMX_TESTS
-
 check_PROGRAMS += \
 	perf_counter \
 	fadd_nbi \
 	atomic_nbi \
 	put_signal \
-	put_signal_nbi
+	put_signal_nbi \
+	shmemx_wait_until_all \
+	shmemx_wait_until_any \
+	shmemx_wait_until_some \
+	shmemx_test_all \
+	shmemx_test_any \
+	shmemx_test_some \
+	c11_test_shmemx_wait_until \
+	c11_test_shmemx_test
 
 if HAVE_PTHREADS
 check_PROGRAMS += \
 	gettid_register
 endif
-endif SHMEMX_TESTS
 
 if HAVE_CXX
-if SHMEMX_TESTS
 check_PROGRAMS += \
 	cxx_test_shmem_g \
 	cxx_test_shmem_get \
@@ -65,7 +63,8 @@ check_PROGRAMS += \
 	cxx_test_shmem_wait_until \
 	cxx_test_shmem_test
 endif
-endif
+
+endif SHMEMX_TESTS
 
 TESTS = $(check_PROGRAMS)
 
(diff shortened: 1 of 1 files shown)


test/shmemx/Makefile.am
```diff
@@ -11,15 +11,9 @@
...
-	shmemx_wait_until_any \
-	shmemx_wait_until_some \
-	shmemx_test_all \
-	shmemx_test_any \
-	shmemx_test_some \
-	c11_test_shmemx_wait_until \
-	c11_test_shmemx_test
+check_PROGRAMS =
+
+if SHMEMX_TESTS
```
jdinan: We actually have three different batches of things that can be turned on independently here: profiling tests, portals4 tests, and shmem extensions tests.  This flag was already in the right place -- just around the shmemx tests.

jdinan: It's weird because some things in here are connected to the SOS build flags.

test/shmemx/Makefile.am
```diff
@@ -31,23 +25,27 @@ check_PROGRAMS += \
...
+	c11_test_shmemx_test
 
 if HAVE_PTHREADS
 check_PROGRAMS += \
 	gettid_register
 endif
-endif SHMEMX_TESTS
 
 if HAVE_CXX
-if SHMEMX_TESTS
```
jdinan: C++ generic bindings are an extension and should have `SHMEMX_TESTS` around them.

test/shmemx/Makefile.am
```diff
@@ -90,10 +89,14 @@ LDADD += $(top_builddir)/pmi-simple/libpmi_simple.la
 endif
 
 if SHMEMX_TESTS
+
+if HAVE_PTHREADS
```
jdinan: The `if` here should not be necessary.  More ifs == more opportunities for automake to automake you sad.

test/shmemx/Makefile.am
```diff
@@ -90,10 +89,14 @@ LDADD += $(top_builddir)/pmi-simple/libpmi_simple.la
...
 gettid_register_LDFLAGS = $(AM_LDFLAGS) $(PTHREAD_LIBS)
 gettid_register_CFLAGS = $(PTHREAD_CFLAGS)
 gettid_register_LDADD = $(LDADD) $(PTHREAD_CFLAGS)
+endif
 
+if HAVE_CXX
```
jdinan: Same with this `if`.  Is it needed?

test/shmemx/Makefile.am
```diff
@@ -31,23 +25,27 @@ check_PROGRAMS += \
...
-	put_signal_nbi
+	put_signal_nbi \
+	shmemx_wait_until_all \
+	shmemx_wait_until_any \
+	shmemx_wait_until_some \
+	shmemx_test_all \
+	shmemx_test_any \
+	shmemx_test_some \
+	c11_test_shmemx_wait_until \
+	c11_test_shmemx_test
```
jdinan: The move of this stuff from above down to here is the only change I would keep from this PR.

test/shmemx/Makefile.am
davidozog: It seems strange to me that turning off "SHMEMX_TESTS" should still result in building tests in the `test/shmemx` directory.  Is that really what we want?  No problem if so, I just find it a little weird.

test/shmemx/Makefile.am
davidozog: This does have `SHMEMX_TESTS` around it - if I'm not mistaken...  Perhaps the git diff is misleading?

test/shmemx/Makefile.am
davidozog: This `if` is not new to this PR, but that's reasonable.  Want to remove it?

test/shmemx/Makefile.am
davidozog: This `if` is also tangential to this PR, but I think it could probably be removed as well.  Remove it?

test/shmemx/Makefile.am
davidozog: No problem - this was the original intent of the PR, but I was confused about why some tests are built in `shmemx` even with `SHMEMX_TESTS` turned off... hence the proposed change.

test/shmemx/Makefile.am
jdinan: Effectively, `ENABLE_SHMEMX` is always enabled within SOS.  It's only for users of the external tests.

davidozog: Yeah, I noticed you have to change `configure.ac` directly to turn shmemx off.
Alright, cool - I'll push those changes shortly.

PR #896, opened 2019-11-07
Add memory sync before returning from barriers

jdinan:
These are the same memory syncs we do before returning from wait/test
operations to ensure shared memory updates and remote updates cached in
the NIC are visible in memory.

### DIFF
--- a/src/shmem_collectives.h
+++ b/src/shmem_collectives.h
@@ -76,6 +76,10 @@ shmem_internal_sync(int PE_start, int logPE_stride, int PE_size, long *pSync)
         RAISE_ERROR_MSG("Illegal barrier/sync type (%d)\n",
                         shmem_internal_barrier_type);
     }
+
+    /* Ensure remote updates are visible in memory */
+    shmem_internal_membar_acquire();
+    shmem_transport_syncmem();
 }
 
 
--- a/src/shmem_collectives.h
+++ b/src/shmem_collectives.h
@@ -78,7 +78,7 @@ shmem_internal_sync(int PE_start, int logPE_stride, int PE_size, long *pSync)
     }
 
     /* Ensure remote updates are visible in memory */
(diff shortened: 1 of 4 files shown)


src/shmem_collectives.h
```diff
@@ -76,6 +76,10 @@ shmem_internal_sync(int PE_start, int logPE_stride, int PE_size, long *pSync)
         RAISE_ERROR_MSG("Illegal barrier/sync type (%d)\n",
                         shmem_internal_barrier_type);
     }
+
+    /* Ensure remote updates are visible in memory */
+    shmem_internal_membar_acquire();
```
wrrobin: Should we instead call `membar_release`? We would like to ensure the stores become visible, right?

jdinan: That is a good question.  We put acquire fences in wait/test.  Maybe we actually need acq_rel?

jdinan: The acquire prevents operations below the fence from floating up.  But, I think we also want a release to prevent operations from before the fence falling below.

wrrobin: Yes, acq_rel should be the safest option. But, if we want to optimize to either acquire or release, we should do release, I think.

jdinan: I don't think release is sufficient, it still allows the architecture to reorder a subsequent load to occur before the fence.

jdinan: Thanks for the careful review -- will fix this.  :)

wrrobin: Yes, we might need both. Thanks.

jdinan: Fixed by 50e7890

jdinan: @jdinan This one still needs to be fixed.

jdinan:
@wrrobin Made that last acquire -> acq_rel update.

PR #886, opened 2019-09-20
Teams API topic branch

davidozog:
This is a work-in-progress branch that implements the [OpenSHMEM v1.5 teams API](https://github.com/openshmem-org/specification/pull/280).  This pull request is intended to merge the `wip/teams` branch from my fork into the `topic/teams` branch into the SOS main repository.

Here is a running checklist of open questions and TODO items:

- [x] Add more `psync` arrays for back-to-back collectives
- [x] The unit tests are more-or-less consistent with the teams proposal examples, but may need modification (most require multiple PEs).
- [ ] Write unit test(s) with more complete coverage of the teams API and various types.
- [x] Support more than 64 teams.
- [ ] Are some types missing from `shmem_internal_op.h`?
- [ ] Should I add pWrk space for each team at initialization?  I deferred this since SOS doesn't yet make use of pWrk arrays.
- [x] I'm not sure whether `shmem_internal_1st_nonzero_bit` supports big-endiannass.
- [x] The C11/C++ "extras" bindings might need some _extra_ help.
- [x] `SHMEM_TEAM_SHARED` only works with a uniform stride across the local process IDs.  Users can otherwise set `DISABLE_TEAM_SHARED`, but this could definitely be cleaner.
- [x] Whoops - are private contexts lost during team destroy? (test it)

PR #906, opened 2019-11-19
Merge topic/teams into master

jdinan:


jdinan:
@davidozog #886 when into `topic/teams`.  This PR will merge `topic/teams` into `master`.

PR #908, opened 2019-11-21
Fix incorrect type in spinlock

jdinan:

### DIFF
--- a/src/shmem_atomic.h
+++ b/src/shmem_atomic.h
@@ -66,7 +66,7 @@ static inline
 void
 shmem_spinlock_lock(shmem_spinlock_t *lock)
 {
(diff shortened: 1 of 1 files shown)


jdinan:
Simple change -- merging without a review.

davidozog:
@jdinan - I'm curious about this change - how did it come up?  And aren't the return value and the first argument of `__atomic_fetch_add` and `__atomic_load_n` supposed to match?

davidozog:
Ah, thanks!

Another reason I was confused:  I thought `lock->enter` and `lock->exit` where `long`'s, but that was based on my local (stale) source.  They must have changed to `unsigned long` relatively recently...

PR #909, opened 2019-11-22
[WIP]Unit test for max no of teams

rdesai16:


lanl-ompi:

rdesai16:
There is one change left for this test. That change includes destroying some teams when it reaches the limit for max no of teams and keep trying to create another teams. I will soon submit that change as well.

test/shmemx/team_max.c
```diff
@@ -0,0 +1,80 @@
...
+    me = shmem_my_pe();
+    npes = shmem_n_pes();
+
+    shmemx_team_t new_team[NUM_TEAMS];
+
+    i = 0;
+    while(i < 9) {
+        
+        team_size = i+1;
+        team_size = team_size % npes;
```
jdinan: Could you merge the two statements above?  Also, `team_size` can be declared inside of the while loop to restrict the scope of the variable.  This is good coding practice to avoid variables in the outer block (main) that are not used there.

test/shmemx/team_max.c
```diff
@@ -0,0 +1,80 @@
...
+        if(team_size == 0){
+            i++;
+            continue;
+        }
```
jdinan: This conditional looks like it will cause problems.  The split, sync, and reduce operations below are collective and must be called by all PEs in `SHMEMX_TEAM_WORLD` but some PEs will end the iteration here.

test/shmemx/team_max.c
```diff
@@ -0,0 +1,80 @@
...
+        shmemx_sync(SHMEMX_TEAM_WORLD);
+        shmemx_int_and_reduce(SHMEMX_TEAM_WORLD, &dest_ret, &ret, 1);
+ 
+
+        if(ret !=0){
+            printf("ERROR: Max Teams limit reached. Try to destroy some teams!!\n");
+        }else {
+            team_count++;
+        }
+        i++;
```
jdinan: If `i` is incremented on every iteration, this can be converted back to a `for` loop.

test/shmemx/team_max.c
rdesai16: This iteration will be stopped for all PE. I added this condition for a situation when (i+1)==npes. Because in that case (i+1)%npes will give 0. Hence all the PEs will come across same situation for this iteration.

test/shmemx/team_max.c
rdesai16: Done!

test/shmemx/team_max.c
rdesai16: Done!

test/shmemx/team_max.c
jdinan: Interesting, I'm not sure we specified this case fully.

jdinan:
@davidozog Could you take a look at this error? We should be returning the error to the user rather than aborting the program.  We never get to exercise the `shmemx_team_destroy` path in the unit test.
```
$ mpiexec -np 2 test/shmemx/shmemx_team_max
[0000] WARN:  ../../src/shmem_team.c:289: shmem_internal_team_split_strided
[0000]        No more teams available (max = 10), try increasing SHMEM_TEAMS_MAX
[0000] ERROR: ../../src/shmem_team.c:313: shmem_internal_team_split_strided
[0000]        Team split strided failed: child <0, 1, 1>, parent <0, 1, 2>
```

davidozog:
> @davidozog Could you take a look at this error? We should be returning the error to the user rather than aborting the program. We never get to exercise the `shmemx_team_destroy` path in the unit test.
> 
> ```
> $ mpiexec -np 2 test/shmemx/shmemx_team_max
> [0000] WARN:  ../../src/shmem_team.c:289: shmem_internal_team_split_strided
> [0000]        No more teams available (max = 10), try increasing SHMEM_TEAMS_MAX
> [0000] ERROR: ../../src/shmem_team.c:313: shmem_internal_team_split_strided
> [0000]        Team split strided failed: child <0, 1, 1>, parent <0, 1, 2>
> ```

Does #911 fix the problem?

jdinan:
@rdesai16 Thanks for the unit test! This branch has been merged into #911, which fixes the bug found by the test. Closing this PR since it is now a duplicate.

PR #911, opened 2019-12-02
Teams: Warn when psync unavailable & add comments

davidozog:
This PR addresses the issue in #909 where a program that runs out of team resources aborts unnecessarily.

I added a couple comments to clarify the delayed return when no psync is available.

### DIFF
--- a/src/shmem_team.c
+++ b/src/shmem_team.c
@@ -287,6 +287,7 @@ int shmem_internal_team_split_strided(shmem_internal_team_t *parent_team, int PE
         if (myteam->psync_idx == -1 || myteam->psync_idx >= shmem_internal_params.TEAMS_MAX) {
             RAISE_WARN_MSG("No more teams available (max = %ld), try increasing SHMEM_TEAMS_MAX\n",
                             shmem_internal_params.TEAMS_MAX);
+            /* No psync is available, but must call barrier across parent team before returning. */
             myteam->psync_idx = -1;
         } else {
             /* Set the selected psync bit to 0, reserving that slot */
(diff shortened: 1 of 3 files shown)


src/shmem_team.c
```diff
@@ -307,12 +308,15 @@ int shmem_internal_team_split_strided(shmem_internal_team_t *parent_team, int PE
...
     shmem_internal_team_release_psyncs(parent_team, SYNC);
 
-    if (my_pe >= 0 && myteam != NULL && myteam->psync_idx == -1)
-        RAISE_ERROR_MSG("Team split strided failed: child <%d, %d, %d>, parent <%d, %d, %d>\n",
-                         global_PE_start, PE_stride, PE_size,
-                         parent_team->start, parent_team->stride, parent_team->size);
-    else
+    /* If no team was available, print some team triplet info and return -1. */
+    if (my_pe >= 0 && myteam != NULL && myteam->psync_idx == -1) {
+        RAISE_WARN_MSG("Team split strided failed: child <%d, %d, %d>, parent <%d, %d, %d>\n",
```
davidozog: Should this instead be a `DEBUG_MSG`?

jdinan: Yes, that would be better. Also, why are we returning -1? Anything that is nonzero is failure, so -1 is not worse than 1. Perhaps we should (at least internal to SOS) introduce actual error codes and return those? Please use a separate PR if you want to introduce error codes (take a look at Linux `err.h` if you want inspiration).

src/shmem_team.c
```diff
@@ -287,6 +287,7 @@ int shmem_internal_team_split_strided(shmem_internal_team_t *parent_team, int PE
         if (myteam->psync_idx == -1 || myteam->psync_idx >= shmem_internal_params.TEAMS_MAX) {
             RAISE_WARN_MSG("No more teams available (max = %ld), try increasing SHMEM_TEAMS_MAX\n",
                             shmem_internal_params.TEAMS_MAX);
+            /* No psync is available, but must call barrier across parent team before returning. */
```
jdinan: Not sure I understand the comment. Is this to cleanup after the above reduction?

src/shmem_team.c
davidozog: No reason in particular for -1, it's just a placeholder for future error codes.  Perhaps we should return distinct nonzero values for the different failure cases?

Yeah, that sounds good - I will look into error codes.  Are you suggestion we _use_ `errno` (perhaps `EAGAIN` or similar in this case), or create our own "sos_errno" _similar_ to `err.h` (and/or `fi_errno`)?

src/shmem_team.c
davidozog: It may not be necessary.  It's really just a reminder that we cannot return here because all PEs need to finish the barrier across the parent team below.  My intent was also to clarify the check for `myteam->psync_idx == -1` _after_ this parent team barrier.  Maybe a cross-reference between the 2 comments would make that more clear?  Or maybe we just don't need the comments...

src/shmem_team.c
jdinan: My suggestion is to make our own header with SOS error codes. The header won't be exposed to the user yet, but it will be helpful to have meaningful values when debugging. Errno is problematic with threading.

src/shmem_team.c
jdinan: Got it. I think a comment on the barrier below might be helpful. That barrier is there to eliminate a race with back-to-back team creations, right?

jdinan:
Could you merge #909 into this branch to test the fix? Afterward, we can close #909 and merge both changes through this PR.

rdesai16:
Thank you @davidozog for your changes and fixing the issue.

PR #913, opened 2019-12-04
Travis: Add c99 compatibility row

jdinan:


jdinan:
Doesn't seem to be working, closing for now

PR #305, opened 2017-01-10
Add support for trap-on-abort

jdinan:
Define SHMEM_TRAP_ON_ABORT and the abort routine will call the
__builtin_trap() GCC intrinsic.  Useful for generating core files and
interfacing with debuggers.

Signed-off-by: James Dinan <james.dinan@intel.com>

PR #915, opened 2019-12-10
Reduce return value across PEs during team creation

davidozog:
With upcoming changes to the teams proposal [(#305)](https://github.com/openshmem-org/specification/pull/305/), we need to guarantee that the return values are equal across PEs after team creation.  We can also check whether the reduction is successful in the `shmemx_team_max` unit test.

### DIFF
--- a/src/shmem_team.c
+++ b/src/shmem_team.c
@@ -163,6 +163,7 @@ int shmem_internal_team_init(void)
                                                          shmem_internal_params.TEAMS_MAX];
 
     psync_pool_avail = shmem_internal_shmalloc(2 * N_PSYNC_BYTES);
+    if (NULL == psync_pool_avail) return -1;
     psync_pool_avail_reduced = &psync_pool_avail[N_PSYNC_BYTES];
 
     /* Initialize the psync bits to 1, making all slots available: */
@@ -177,6 +178,7 @@ int shmem_internal_team_init(void)
 
     /* Initialize an integer used to agree on an equal return value across PEs in team creation: */
     team_ret_val = shmem_internal_shmalloc(sizeof(int) * 2);
+    if (NULL == team_ret_val) return -1;
     team_ret_val_reduced = &team_ret_val[1];
 
     return 0;
@@ -325,7 +327,7 @@ int shmem_internal_team_split_strided(shmem_internal_team_t *parent_team, int PE
 
     shmem_internal_op_to_all(team_ret_val_reduced, team_ret_val, 1, sizeof(int),
                              parent_team->start, parent_team->stride, parent_team->size, NULL,
-                             psync, SHM_INTERNAL_BOR, SHM_INTERNAL_INT);
+                             psync, SHM_INTERNAL_MAX, SHM_INTERNAL_INT);
 
     shmem_internal_team_release_psyncs(parent_team, REDUCE);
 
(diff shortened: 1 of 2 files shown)


test/shmemx/shmemx_team_max.c
```diff
@@ -57,9 +57,11 @@ int main(void)
         /* If success was not global, free a team and retry */
         if (dest_ret != 0) {
             /* FIXME: The team is currently leaked in the case below */
```
jdinan: Remove FIXME comment

src/shmem_team.c
```diff
@@ -173,6 +175,10 @@ int shmem_internal_team_init(void)
     shmem_internal_bit_clear(psync_pool_avail, N_PSYNC_BYTES, SHMEM_TEAM_WORLD_INDEX);
     shmem_internal_bit_clear(psync_pool_avail, N_PSYNC_BYTES, SHMEM_TEAM_SHARED_INDEX);
 
+    /* Initialize an integer used to agree on an equal return value across PEs in team creation: */
+    team_ret_val = shmem_internal_shmalloc(sizeof(int) * 2);
```
jdinan: Need to NULL check the returned value.

src/shmem_team.c
```diff
@@ -187,6 +193,7 @@ void shmem_internal_team_fini(void)
     free(shmem_internal_team_pool);
     shmem_internal_free(shmem_internal_psync_pool);
     shmem_internal_free(psync_pool_avail);
+    shmem_internal_free(team_ret_val);
```
jdinan: Given the above NULL check comment, you may want to guard the above free with `if (team_ret_val != NULL)`...

src/shmem_team.c
```diff
@@ -310,15 +320,22 @@ int shmem_internal_team_split_strided(shmem_internal_team_t *parent_team, int PE
 
     shmem_internal_team_release_psyncs(parent_team, SYNC);
 
+    /* This OR reduction assures all PEs return the same value.  */
+    psync = shmem_internal_team_choose_psync(parent_team, REDUCE);
+
+    shmem_internal_op_to_all(team_ret_val_reduced, team_ret_val, 1, sizeof(int),
+                             parent_team->start, parent_team->stride, parent_team->size, NULL,
+                             psync, SHM_INTERNAL_BOR, SHM_INTERNAL_INT);
```
jdinan: Would `SHMEM_INTERNAL_MAX` be a better reduction op? That is guaranteed to lead to a valid error code that is meaningful, whereas OR'ing multiple error codes could lead to something totally different.

davidozog: Yeah, good point.

src/shmem_team.c
davidozog: Doesn't `shmem_internal_free` already have a NULL check?

src/shmem_team.c
```diff
@@ -177,6 +178,7 @@ int shmem_internal_team_init(void)
 
     /* Initialize an integer used to agree on an equal return value across PEs in team creation: */
     team_ret_val = shmem_internal_shmalloc(sizeof(int) * 2);
+    if (NULL == team_ret_val) return -1;
```
jdinan: If `team_init` fails, the calling context does not call `team_fini`, so we should really be trying to do the cleanup here before returning from init. I will push a patch to show you what I mean.

src/shmem_team.c
jdinan: So it does, but actually this function won't get called if init fails (see below) so we need to handle cleanup before returning from init.

src/shmem_team.c
davidozog: Oh yeah - that makes sense.  Thanks!

PR #923, opened 2019-12-18
Relax shr atomics memory model

jdinan:
Addon to #922 to relax the memory model from sequential consistency to acquire/release.

### DIFF
--- a/src/shr_transport.h4
+++ b/src/shr_transport.h4
@@ -235,7 +235,7 @@ shmem_shr_transport_swap(shmem_ctx_t ctx, void *target, void *source,
 #define SHMEM_DEF_SWAP(STYPE,TYPE,ITYPE)                                \
         case ITYPE:                                                     \
                 __atomic_exchange((TYPE *)remote_ptr, (TYPE *)source,   \
-                                  (TYPE *)dest, __ATOMIC_SEQ_CST);      \
+                                  (TYPE *)dest, __ATOMIC_ACQ_REL);      \
             break;
 
     switch (datatype) {
@@ -274,8 +274,8 @@ shmem_shr_transport_cswap(shmem_ctx_t ctx, void *target, void *source,
                                                          &expected,             \
                                                          (TYPE *)source,        \
                                                          false,                 \
-                                                         __ATOMIC_SEQ_CST,      \
-                                                         __ATOMIC_SEQ_CST);     \
+                                                         __ATOMIC_ACQ_REL,      \
+                                                         __ATOMIC_ACQUIRE);     \
                 if (swapped)                                                    \
                     *(TYPE *)dest = *(TYPE *)operand;                           \
                 else                                                            \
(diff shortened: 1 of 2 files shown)


jdinan:
Rebased onto master to incorporate edits on the CPU AMOs branch. This should be ready to review now.

jdinan:
@nspark Wouldn't mind your eyes on this if you have a moment. For context, a PR we just merged added shared memory atomics (--enable-shr-atomics) that can be used in conjunction with XPMEM support. If only XPMEM is enabled (without OFI or Portals), you get a purely shared memory build of SOS. The shared memory atomics PR used sequential consistency as an initial step. This PR relaxes the memory model to use acquire/release.

Once these changes are in place, my plan is to start accumulating "litmus tests" for the memory model under the SOS tests tree and we can run them against this shared memory build.

src/shr_transport.h4
```diff
@@ -356,22 +356,22 @@ shmem_shr_transport_atomic(shmem_ctx_t ctx, void *target, const void *source,
 
 #define SHMEM_DEF_BAND_OP(STYPE,TYPE,ITYPE)                                                     \
                 case ITYPE:                                                                     \
-                    __atomic_fetch_and((TYPE *)remote_ptr, *((TYPE *)source), __ATOMIC_SEQ_CST);\
+                    __atomic_fetch_and((TYPE *)remote_ptr, *((TYPE *)source), __ATOMIC_RELEASE);\
```
davidozog: I would have guessed the `__atomic_fetch_[op]`'s would do release-acquire, because first there's a read, then a write.  Can you please explain the reasoning for release here (and below)?

davidozog: Maybe it's because we're not actually reading the return value...?

jdinan: That was my reasoning. We only have fetching atomic intrinsics, so I used those with release consistency to implement non-fetching OpenSHMEM AMOs.

davidozog:
I think we can go ahead and merge this if you'd like - ready when you are, @jdinan.

nspark:
I'm sorry that it took me a couple of days to respond. I think what you have here for memory orderings is fine. That said, the direction I was planning on taking the OpenSHMEM memory model was to have all all atomics use `memory_order::relaxed`. Forming correct release-acquire pairs would require the judicious of atomic fences, which would be the responsibility of the application developer. This, I think, would balance performance and flexibility. Applications could get all the performance possible with relaxed atomics (e.g., GUPS), but have all the necessary mechanisms for synchronization.

jdinan:
@nspark This sounds like a good next step. My plan was to collect litmus tests in the SOS test suite so we can extend what's here and use SOS to validate our work.

FWIW, a colleague who works on compilers indicated that the relaxed memory model is not well understood or perhaps well supported. Among the challenges with relaxed ordering is the "out of thin air" problem [1]. I haven't yet had a chance to get smart about relaxed ordering; might be that the constraints imposed by the SHMEM memory model help to alleviate some of these issues.

[1] https://www.cl.cam.ac.uk/~pes20/cpp/notes42.html

PR #929, opened 2020-02-06
Signal operation type for put with signal

wrrobin:


src/shr_transport.h4
```diff
@@ -175,23 +175,36 @@ shmem_shr_transport_put(shmem_ctx_t ctx, void *target, const void *source,
...
     shmem_transport_xpmem_put(target, source, len, pe,
                               shmem_internal_get_shr_rank(pe));
     shmem_internal_membar_acq_rel(); /* Memory fence to ensure target PE observes
                                         stores in the correct order */
+    uint64_t old_signal_val = 0;
+    if (sig_op == SHMEM_SIGNAL_ADD) {
+        shmem_transport_xpmem_get(&old_signal_val, sig_addr, sizeof(uint64_t), pe,
+                                  shmem_internal_get_shr_rank(pe));
+        signal += old_signal_val;
+    }
```
jdinan: This needs to be an atomic operation, use `shmem_shr_transport_atomic`.

src/shr_transport.h4
```diff
@@ -175,23 +175,36 @@ shmem_shr_transport_put(shmem_ctx_t ctx, void *target, const void *source,
...
     shmem_transport_cma_put(target, source, len, pe,
                             shmem_internal_get_shr_rank(pe));
     shmem_internal_membar_acq_rel(); /* Memory fence to ensure target PE observes
                                         stores in the correct order */
+    uint64_t old_signal_val = 0;
+    if (sig_op == SHMEM_SIGNAL_ADD) {
+        shmem_transport_cma_get(&old_signal_val, sig_addr, sizeof(uint64_t), pe,
+                                shmem_internal_get_shr_rank(pe));
+        signal += old_signal_val;
+    }
```
jdinan: This needs to be an atomic. CMA does not support atomics; use `shmem_transport_atomic`.

src/shr_transport.h4
wrrobin: @jdinan `shmem_shr_transport_atomic` is enabled through `USE_SHR_ATOMICS`. Unless we use `--enable-shr-atomics` flag, we will not be able to use this API, right? Please let me know if I am missing something. 
The other question is, why does it need to be atomic? According to the spec, `All signaling operations on the signal data object completes as if performed atomically`. I thought the clause `as if performed atomically` gave the flexibility to use fi_inject or memcpy.

jdinan: If the target is waiting for several messages using the same signal variable, this will not work, right?

jdinan: Right, you would need to write code to support builds with and without shared memory atomics.

wrrobin: Thanks @jdinan. One question, from the current code, there seems to be no relation between the flags `USE_SHR_ATOMICS` and `USE_XPMEM`/`USE_CMA`. Was there any particular reason behind this? I am thinking shouldn't `USE_SHR_ATOMICS` be defined when either of these on-node transports is defined?

jdinan: For nearly all builds, we can't enable shared memory atomics because the networking layer is not coherent with processor atomics. By default, we only enable shared memory copy to implement put/get. This is why shared memory atomics are a separate option, and disabled by default. CMA is essentially memory copy performed by the Linux kernel and it cannot support atomics.

wrrobin: For CMA, we should directly use the transport layer atomics then, as you also suggested later. But, for XPMEM, can we say that enabling USE_XPMEM should enable USE_SHR_ATOMICS as well? We can use the shared memory atomics if `shmem_internal_get_shr_rank` returns something other than -1, right? But, currently, `shmem_shr_transport_use_atomic` also seems to be guarded by `USE_SHR_ATOMICS`.

wrrobin:
@jdinan Can you please review the changes in shr_transport? There is one test failing in Travis that I am taking a look, but looks like it is related to the unit test change.

src/shr_transport.h4
jdinan: Correct, it is always safe to use shared memory put/get because OpenSHMEM says overlapping operations (e.g. between local shared memory and remote network transport accesses) leads to undefined behavior. Atomics do allow this overlapping. Therefore, it is only safe to use shared memory atomics when you know the transport layer is coherent with processor atomics. Because of this, enabling XPMEM only enables shared memory put/get and there is an additional flag to enable shared memory atomics.

src/shr_transport.h4
```diff
@@ -27,6 +27,11 @@ include(shmem_bind_c.m4)dnl
 #include "transport_cma.h"
 #endif
 
+#if defined(USE_XPMEM) && !defined(USE_OFI) && !defined(USE_PORTALS4)
+#  define ONLY_XPMEM_TRANSPORT 1
+#else
+#  define ONLY_XPMEM_TRANSPORT 0
+#endif
```
jdinan: Oh dear, no. This will cause shared memory atomics to be enabled in some paths and not others. You should only need to check `USE_SHR_ATOMICS`. This flag is set by the configure script. If you are concerned about it being set incorrectly, please resolve those concerns at configure time, not here.

jdinan: For example, `shmem_wait` will not be using the correct atomics if `USE_SHR_ATOMICS` is not set.

src/shr_transport.h4
```diff
@@ -591,4 +568,51 @@ SHMEM_DEFINE_FOR_AMO(SHMEM_DEF_SUM_OP)
...
+#if USE_MEMCPY
+    memcpy(target, source, len);
+    if (sig_op == SHMEM_SIGNAL_ADD) *sig_addr += signal;
+    else *sig_addr = signal;
+#elif USE_XPMEM
+    shmem_transport_xpmem_put(target, source, len, pe,
+                              shmem_internal_get_shr_rank(pe));
+    shmem_internal_membar_acq_rel(); /* Memory fence to ensure target PE observes
+                                        stores in the correct order */
+    if (ONLY_XPMEM_TRANSPORT) {
```
jdinan: This should be replaced by `#if USE_SHR_ATOMICS` but other than change this looks looks ok.

src/shr_transport.h4
wrrobin: Ok. I was thinking initially to use `USE_SHR_ATOMICS` by defining it when only XPMEM is enabled in the configure script. But for some reason, I misunderstood that we wanted a separate flag.

jdinan:
The `put_signal_nbi` test is failing in the Travis row with huge pages and OFI fence enabled. Huge page support is auto-disabling, looks like possibly the issue is related to the OFI fence flag?
```
[0000] WARN:  ../../src/symmetric_heap_c.c:187: mmap_alloc
[0000]        file open failed, cannot use huge pages
         0: target[1] = 0 not matching 1
         0: target[2] = 0 not matching 2
         0: target[3] = 0 not matching 3
         0: target[4] = 0 not matching 4
         0: target[5] = 0 not matching 5
         0: target[6] = 0 not matching 6
         0: target[7] = 0 not matching 7
         0: target[8] = 0 not matching 8
         0: target[9] = 0 not matching 9
FAIL put_signal_nbi (exit status: 9)
```

wrrobin:
Yes, I am trying to reproduce the issue locally. The test is modified with this PR to include a check for ADD signal atomic. Earlier, with no signal op argument, this test was not failing.

configure.ac
```diff
@@ -414,6 +414,10 @@ AS_IF([test "$transport_xpmem" = "yes" -o "$transport_cma" = "yes"],
        AC_DEFINE([ENABLE_HARD_POLLING], [1], [Enable hard polling])
       ])
 
+if test "$enable_shr_atomics" = "no" -a "$transport_xpmem" = "yes" -a "$transport_ofi" = "no" -a "$transport_portals4" = "no" ; then
+    AC_DEFINE([USE_SHR_ATOMICS], [1], [If defined, the shared memory layer will perform processor atomics.])
+fi
```
jdinan: We have always treated the above as an invalid build; the configure script should emit a warning any time there isn't a transport selected. That having been said, this seems like a good change. You might want to update the description of the `--enable-shr-atomics` to say "default: auto" rather than "default: disabled".

A minor fix here, the check for `"$enable_shr_atomics" = "no"` is not quite right (see Example 1.1 [here](https://autotools.io/autoconf/arguments.html)). It's counterintuitive, but I think this check should be `"$enable_shr_atomics" != "no"` i.e. "not disabled".

src/shmem_internal.h
```diff
@@ -303,6 +303,20 @@ extern unsigned int shmem_internal_rand_seed;
...
+#define SHMEM_ERR_CHECK_SIG_OP(op)                                                      \
+    do {                                                                                \
+        switch(op) {                                                                    \
+            case SHMEM_SIGNAL_SET:                                                      \
+            case SHMEM_SIGNAL_ADD:                                                      \
+                break;                                                                  \
+            default:                                                                    \
+                fprintf(stderr, "ERROR: %s(): Argument \"%s\", "                        \
+                                "invalid atomic operation for signal (%d)\n",           \
+                        __func__, #op, (int) (op));                                     \
```
jdinan: Oh, how I wish these checks used `RAISE_ERROR_MSG`...

wrrobin: Do you want me to change this here? Or should we do it in a new PR changing all the instances?

jdinan: No need for a change on this PR. There's nothing broken here, I am just being OCD. 👍

davidozog:
@jdinan, do you have any opinion as to whether this should be included in the v1.4.5 release?  All testing would have to be redone, so @wrrobin and I have tentatively decided to defer this merge.

jdinan:
Our CI testing is pretty thorough, apart from Portals, and this is an extension, so I wouldn't be opposed if you feel that it's ready. Since this PR includes an API change to bring the signaling API into compliance with the latest OpenSHMEM 1.5 draft, it might be nice to include.

wrrobin:
@jdinan I have tried with different combinations of `--enable/disable-shr-atomics` and `with-xpmem/ofi` and here are the results with the current configure script:

--enable-shr-atomics -> USE_SHR_ATOMICS defined
--disable-shr-atomics -> undefined
--enable-shr-atomics --with-xpmem -> USE_SHR_ATOMICS defined
**_--disable-shr-atomics --with-xpmem -> USE_SHR_ATOMICS defined_**
**_--with-xpmem -> undefined_**
--with-xpmem --with-ofi -> undefined
**_--with-xpmem --with-ofi --enable-shr-atomics -> USE_SHR_ATOMICS defined_**
--with-xpmem --with-ofi --disable-shr-atomics -> undefined

IIUC, the **_bold italic_** ones above are producing incorrect configurations. Please let me know if you think so as well.

jdinan:
With your current code:
```
if test "$enable_shr_atomics" != "no" -a "$transport_xpmem" = "yes" -a "$transport_ofi" = "no" -a "$transport_portals4" = "no" ; then
    AC_DEFINE([USE_SHR_ATOMICS], [1], [If defined, the shared memory layer will perform processor atomics.])
fi
```
For 4, I see:
```
$ ./configure --disable-shr-atomics --with-xpmem=$HOME/opt/xpmem --enable-pmi-simple
...
$ grep USE_SHR src/config.h
/* #undef USE_SHR_ATOMICS */
```
For 5:
```
$ ./configure --with-xpmem=$HOME/opt/xpmem --enable-pmi-simple
...
$ grep USE_SHR src/config.h
#define USE_SHR_ATOMICS 1
```
That's the right behavior for 7. User got what they asked for. Can you please re-check? The above look like the correct behavior.

Another possibility. There is a portability issue with string comparison in the test utility, which is why you often see `if test "x$enable_shr_atomics" != "xno" ...`. I don't think Linux has this issue, which is why we have been lazy about using this style. You could try that if you are seeing weird behavior.

wrrobin:
I am not sure why I have seen those behaviors before. I am seeing the correct configuration now similar to yours.

wrrobin:
@davidozog Can you please review the PR? If we do not see any other issues, we can merge this for the release.

test/shmemx/put_signal.c
```diff
@@ -86,6 +86,41 @@ int main(int argc, char *argv[])
...
+        target[i] = 0;
+
+    sig_addr = 0;
+    int local_counter = 0;
+
+    shmem_barrier_all();
+    if (me == 0) {
+        shmemx_long_put_signal(target, source, MSG_SZ, &sig_addr, 1, SHMEM_SIGNAL_ADD, dest);
+        local_counter++;
+#if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 201112L
```
davidozog: All these checks for C11 seem a little strange, especially since we don't do it on `shmemx_long_put_signal`.  Can't we do something like this at the top of the tests?

```c
#if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 201112L
#define SHMEM_WAIT_UNTIL  shmem_wait_until
#define SHMEM_PUT_SIGNAL  shmemx_put_signal
#else
#define SHMEM_WAIT_UNTIL shmem_uint64_wait_until
#define SHMEM_PUT_SIGNAL shmemx_long_put_signal
#endif
```

Side note: when I try this with Clang, I get a nasty error from the generic selection.  However, when I remove [this `default: shmem_ctx_c11_generic_selection_failed` line here](https://github.com/Sandia-OpenSHMEM/SOS/blob/master/mpp/shmemx.h4#L483), it works fine...  Note that [`shmem_put` and (some) friends](https://github.com/Sandia-OpenSHMEM/SOS/blob/master/mpp/shmem.h4#L357) do not have this `default:` case either...  I feel like @jdinan and I have seen this before, but I don't remember the reason for this.

I suppose another more pertinent comment I'm making is that the C11 `shmem_put_signal` interfaces don't seem to be covered.

src/transport_portals4.h
```diff
@@ -1112,6 +1100,19 @@ int shmem_transport_atomic_supported(ptl_op_t op, ptl_datatype_t datatype)
...
+                                    uint64_t *sig_addr, uint64_t signal, int sig_op, int pe)
+{
+    /* FIXME: Need to optimize non-blocking put with signal for Portals. Current implementation below keeps
+ *      * the "fence" in between data and signal put */
+    shmem_transport_put_nbi(ctx, target, source, len, pe);
+    shmem_transport_fence(ctx);
+    if (sig_op == SHMEM_SIGNAL_ADD)
+        shmem_transport_atomic(ctx, sig_addr, &signal, sizeof(uint64_t), pe, SHM_INTERNAL_SUM, SHM_INTERNAL_UINT64);
+    else
+        shmem_transport_put_scalar(ctx, sig_addr, &signal, sizeof(uint64_t), pe);
```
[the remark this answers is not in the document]
wrrobin: Yes, I wanted to know abt this.. will change then.

test/shmemx/put_signal.c
wrrobin: They are currently defined as `shmemx_put_signal`. Are you referring to something else?

wrrobin: Ok. I spoke too soon, I see the difference, But, are you saying you are seeing these errors only for put-signal, not for NBI atomics? They should have the same error too, right?

davidozog: You're right @wrrobin, one might expect the NBI atomics to show a similar error, but I just tried it out and don't see any issue, strangely.

Anyway, I think we can ignore my suggestion above and I'll file an issue for this.

davidozog:
@wrrobin and @jdinan - I've just run through the release checklist with this branch included and had no problems.  I haven't yet tested the GNI provider with this patch though, because Cori is under maintenance - but I don't expect to see any issues.

Are we ok with merging this and including in a v1.4.5 release today?

wrrobin:
@davidozog Sounds good to me. I tested with GNI (Cori) last time and all the unit and shmemx tests passed.

PR #930, opened 2020-02-11
Update NEWS and MAJOR/MINOR_VERSION for v1.4.5

davidozog:
Here is a draft of the `NEWS` file for a v1.4.5 release.  Am I missing anything?

With regard to the open PRs, I think #929 might be ok to include in v1.4.5, but everything is tested except for that change... (except Portals, still need to check that).  #918 may change as v1.5 of the OpenSHMEM specification is released, so I tend to think we should defer that one.  Thoughts @jdinan and @wrrobin?

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -9,6 +9,8 @@ v1.4.5
 - Added support for shared memory atomic operations with the appropriate
   acquire/release memory ordering where required.  This feature requires
   the --enable-shr-transport build flag.
+- Improved experimental support for the RXM/Verbs provider stack with
+  libfabric 1.8 and newer.
 - Updated SOS to use the new OFI memory registration mode flags instead of
   the deprecated FI_MR_SCALABLE/FI_MR_BASIC modes.
 - Updated the wait/test any/all/some "status" array semantics to reflect
--- a/NEWS
+++ b/NEWS
@@ -18,7 +18,7 @@ v1.4.5
 - Updated the utility atomics (spinlocks and counters) to use __atomic built-in
   functions rather than the deprecated __sync builtins.
 - Removed redundant shmem_barrier_all from the GUPS example application.
-- Moved examples taken from the OpenSHMEM specification to the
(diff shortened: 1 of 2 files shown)


NEWS
```diff
@@ -2,6 +2,25 @@
...
+- Added the OpenSHMEM wait/test vector API to the shmemx interfaces.
+- Added support for shared memory atomic operations with the appropriate
+  acquire/release memory ordering where required.  This feature requires
+  the --enable-shr-transport build flag.
+- Updated SOS to use the new OFI memory registration mode flags instead of
+  the deprecated FI_MR_SCALABLE/FI_MR_BASIC modes.
+- Updated the wait/test any/all/some "status" array semantics to reflect
+  recent changes to the OpenSHMEM specification.
+- Added a memory sync before returning from barriers to ensure shared memory
+  updates and remote updates cached in the NIC are visible in memory.
```
wrrobin: Not sure if the sentence structure is correct for this one.  Should we change it to - 
` Added memory sync before returning from barriers to ensure shared memory
  updates and remote updates cached in the NIC that are visible in memory. `

jdinan: The "that" in the proposed change is throwing me off. What part needs correction in the original sentence?

NEWS
```diff
@@ -2,6 +2,25 @@
...
+- Updated the utility atomics (spinlocks and counters) to use __atomic built-in
+  functions rather than the deprecated __sync builtins.
+- Removed redundant shmem_barrier_all from the GUPS example application.
+- Moved examples taken from the OpenSHMEM specification to the
```
jdinan:
```suggestion
- Moved unit tests derived from OpenSHMEM specification examples to the
```

NEWS
wrrobin: Oops. Sorry for that "that". That "that" was added by me trying to correct it. I am ok with the sentence as is. Sorry for the confusion.

PR #935, opened 2020-02-25
Update SOS and libfabric versions in build-script

davidozog:
Doing the final steps in the SOS release checklist (for `tests-sos` CI).


davidozog:
Thanks!  I tested locally and it checks out.

PR #938, opened 2020-03-06
UCX Transport

jdinan:
This PR adds an SOS transport that uses the UCX/UCP interfaces. It fully supports all of the OpenSHMEM interfaces, but has not been tested/optimized for performance. A known limitation is that there are currently no context-related optimizations.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -349,31 +349,55 @@ OPAL_CHECK_OFI([ofi],
     [transport_ofi="no"])
 
 # FIXME: Temporary hack
-AC_DEFINE([USE_UCX], [1], [Enable UCX transport])
-transport_ucx="yes"
+AC_ARG_WITH([ucx],
+    [AC_HELP_STRING([--with-ucx],
+                    [Enable the UCX transport. (default: disabled)])])
 
-# If both OFI and portals4 requested, user needs to choose one:
-if test -n "$with_ofi" -a "$with_ofi" != "no" -a -n "$with_portals4" -a "$with_portals4" != "no" ; then
-    AC_MSG_ERROR([Cannot choose both OFI and portals4 transports, see --help for details])
-# Check which was requested, OFI or portals4:
+num_transports=""
+if test -n "$with_portals4" -a "$with_portals4" != "no" ; then
+    num_transports="x$num_transports"
+fi
+if test -n "$with_ofi" -a "$with_ofi" != "no" ; then
+    num_transports="x$num_transports"
+fi
+if test -n "$with_ucx" -a "$with_ucx" != "no" ; then
+    num_transports="x$num_transports"
+    transport_ucx="yes"
+fi
+
+if test -n "$num_transports" -a "$num_transports" != "x" ; then
+    AC_MSG_ERROR([Cannot select more than one transport, see --help for details])
+fi
+
+if test -n "$with_portals4" -a "$with_portals4" != "no" ; then
+    transport="portals4"
+    AC_DEFINE([USE_PORTALS4], [1], [Define if Portals 4 transport active])
 elif test -n "$with_ofi" -a "$with_ofi" != "no" ; then
     transport="ofi"
-    transport_portals4="no"
     AC_DEFINE([USE_OFI], [1], [Define if OFI transport active])
-elif test -n "$with_portals4" -a "$with_portals4" != "no" ; then
-    transport="portals4"
-    transport_ofi="no"
-    AC_DEFINE([USE_PORTALS4], [1], [Define if Portals 4 transport active])
-# If neither, set transport to "none" and print a warning:
+elif test -n "$with_ucx" -a "$with_ucx" != "no" ; then
+    transport="ucx"
+    transport_ucx="yes"
+    # FIXME: Temporary hack until UCX is wired into the build system properly
+    if test "$with_ucx" != "yes" ; then
+        CFLAGS="$CFLAGS -I$with_ucx/include"
+        LDFLAGS="$LDFLAGS -L$with_ucx/lib -lucs -lucp"
+    else
+        LDFLAGS="$LDFLAGS -lucs -lucp"
+    fi
+    AC_DEFINE([USE_UCX], [1], [Define if UCX transport active])
+    AC_DEFINE([ENABLE_HARD_POLLING], [1], [Enable hard polling])
 else
     transport="none"
-    transport_ofi="no"
     transport_portals4="no"
+    transport_ofi="no"
+    transport_ucx="no"
     AC_MSG_WARN([Neither OFI nor portals4 transport requested])
 fi
-AM_CONDITIONAL([USE_UCX], [test "$transport_ucx" = "yes"])
-AM_CONDITIONAL([USE_OFI], [test "$transport_ofi" = "yes"])
+
 AM_CONDITIONAL([USE_PORTALS4], [test "$transport_portals4" = "yes"])
+AM_CONDITIONAL([USE_OFI], [test "$transport_ofi" = "yes"])
+AM_CONDITIONAL([USE_UCX], [test "$transport_ucx" = "yes"])
 
 SANDIA_CHECK_XPMEM(
     [transport_xpmem="yes"],
@@ -844,7 +868,7 @@ AC_OUTPUT
 AS_IF([test "$enable_pmi_mpi" != "yes" -a "$enable_pmi_simple" != "yes" -a "$opal_external_pmix_version_found" != 1 -a "$ompi_check_pmi_happy" != "yes" -a -z "$pmi_type"],
          [AC_MSG_ERROR([No PMI client interface was configured, consider --enable-pmi-simple or --with-pmi])])
 
-AS_IF([test "$transport_portals4" != "yes" -a "$transport_ofi" != "yes"],
+AS_IF([test -z "$num_transports"],
       [AC_MSG_WARN([No transport found, resulting library will be unable to exchange messages])])
 
 AS_IF([test "$shmem_cv_c11_works" != "yes"],
@@ -880,6 +904,7 @@ echo ""
 echo "Transports:"
 echo "  Portals 4:      $transport_portals4"
 echo "  OFI:            $transport_ofi"
+echo "  UCX:            $transport_ucx"
 echo ""
 echo "On Node Communication:"
 echo "  XPMEM:          $transport_xpmem"
--- a/configure.ac
+++ b/configure.ac
@@ -348,10 +348,9 @@ OPAL_CHECK_OFI([ofi],
     [transport_ofi="yes"],
     [transport_ofi="no"])
 
-# FIXME: Temporary hack
-AC_ARG_WITH([ucx],
-    [AC_HELP_STRING([--with-ucx],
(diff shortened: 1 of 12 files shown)


davidozog:
Very interesting, thanks @jdinan.  I'll take a gander.  @wrrobin may also want to have a look.

jdinan:
@davidozog FYI, I also added Travis testing for the UCX build.

wrrobin:
Thanks @davidozog. Will definitely take a look.

src/shmem_comm.h
```diff
@@ -49,7 +49,10 @@ void
 shmem_internal_put_nb(shmem_ctx_t ctx, void *target, const void *source, size_t len, int pe,
                       long *completion)
 {
-    if (len == 0) return;
+    if (len == 0) {
+        *completion = 1;
```
wrrobin: Wow.. how did it work before? Guess we never checked this before?

src/shmem_env_defs.h
```diff
@@ -109,6 +109,11 @@ SHMEM_INTERNAL_ENV_DEF(OFI_STX_DISABLE_PRIVATE, bool, false, SHMEM_INTERNAL_ENV_
                        "Disallow private contexts from having exclusive STX access")
 #endif
 
+#ifdef USE_UCX
+SHMEM_INTERNAL_ENV_DEF(PROGRESS_INTERVAL, long, 1000, SHMEM_INTERNAL_ENV_CAT_TRANSPORT,
+                       "Polling interval for progress thread in microseconds (0 to disable)")
+#endif
```
wrrobin: @jdinan Does UCX have any similar parameter for users? I am thinking, for PSM2, we did not introduce any such parameter in the SHMEM layer. And this will also cause additional threading in the SHMEM, right? Also, will this parameter need to be tuned? Should we expect different behavior for different MLX NICs?

src/shmem_comm.h
jdinan: OFI doesn't use the completion flags, and I think we just never noticed with the Portals build. At some point, the transport API should be updated to have blocking operations, nonblocking operations with no handle (NBI), and if we really care about the cases where `completion` is used nonblocking operations with a transport-defined request object (NBE). With UCX, it would be much cleaner to pass around the UCX request object rather versus a user-supplied completion flag.

src/shmem_env_defs.h
jdinan: No, UCX does not spawn a thread to generate automatic progress. As with all progress threads, there is no perfect value to put here and performance will vary based on system, application, phase of the moon, etc. For benchmarking runs (and most applications) the progress thread can be disabled, but it is needed to be fully compliant with the OpenSHMEM spec.

jdinan:
Gentle nudge on this PR. We've all got a lot on our plates right now. If you can make a little time for this, it would be great to close it out to avoid divergence from master. I would be happy to set up a meeting and walk you through the UCX code if it's helpful.

davidozog:
@jdinan - Sure, I'll set up a meeting (offline) to go through this.  That should help save time figuring out some of the UCX stuff.

wrrobin:
+1

src/transport_ucx.c
```diff
@@ -0,0 +1,373 @@
...
+    /* Clean up peers table */
+    for (i = 0; i < shmem_internal_num_pes; i++) {
+        ucp_rkey_destroy(shmem_transport_peers[i].data_rkey);
+        ucp_rkey_destroy(shmem_transport_peers[i].heap_rkey);
+        ucs_status_ptr_t pstatus = ucp_ep_close_nb(shmem_transport_peers[i].ep,
+                                                   UCP_EP_CLOSE_MODE_FLUSH);
+        shmem_transport_ucx_complete_op(pstatus);
+        free(shmem_transport_peers[i].addr);
+    }
+
```
jdinan: Free shmem_transport_peers

src/transport_ucx.c
```diff
@@ -0,0 +1,373 @@
...
+}
+
+int shmem_transport_fini(void)
+{
+    ucs_status_t status;
+    int i;
+    void *progress_out;
+
+    __atomic_store_n(&shmem_transport_ucx_progress_thread_enabled, 0, __ATOMIC_RELEASE);
+    pthread_join(shmem_transport_ucx_progress_thread, &progress_out);
```
jdinan: Should we do this when the progress thread is disabled?

davidozog: I don't think it would hurt to check whether the progress thread is enabled before joining.  A quick test I did suggests `pthread_join` returns nonzero if the thread never started.

src/transport_ucx.c
```diff
@@ -0,0 +1,373 @@
+/* -*- C -*-
+ *
+ * Copyright (c) 2020 NVidia Corporation.
```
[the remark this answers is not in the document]
jdinan: Neither statement is actually necessary, but I never wanted an extra conversation with lawyers to verify this. See, for example https://en.wikipedia.org/wiki/All_rights_reserved.

That said, these statements are benign and redundant, so there's no real harm in adding them. :neckbeard:

src/transport_ucx.h
```diff
@@ -0,0 +1,845 @@
...
+    uint8_t *remote_addr;
+    ucp_rkey_h rkey;
+    ucs_status_ptr_t pstatus;
+    uint64_t value;
+
+    shmem_transport_ucx_get_mr(target, pe, &remote_addr, &rkey);
+
+    switch (len) {
+        case 1:
+            value = (uint64_t)*(uint8_t*)source;
```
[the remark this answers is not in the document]
jdinan: Yeah, isn't this great? It grabs a byte from source and then casts that byte to `uint64_t` to be passed by value to the corresponding UCX routine. Sure, the space may make it a little easier to see that this is not multiplication.

jdinan:
Pushed updates to resolve feedback from code review. I think I see a minor issue in how the threading level is being dealt with. Stay tuned for a fix to that.

jdinan:
Threading level checking fixed. This is ready on my end.

configure.ac
```diff
@@ -348,27 +348,46 @@ OPAL_CHECK_OFI([ofi],
...
+    AC_MSG_WARN([No transport requested])
 fi
-AM_CONDITIONAL([USE_OFI], [test "$transport_ofi" = "yes"])
+
 AM_CONDITIONAL([USE_PORTALS4], [test "$transport_portals4" = "yes"])
+AM_CONDITIONAL([USE_OFI], [test "$transport_ofi" = "yes"])
+AM_CONDITIONAL([USE_UCX], [test "$transport_ucx" = "yes"])
 
 SANDIA_CHECK_XPMEM(
     [transport_xpmem="yes"],
```
wrrobin, line 393: @jdinan Do we need to add `transport_ucx` check in line number 436 condition
```
if test "$enable_shr_atomics" != "no" -a "$transport_xpmem" = "yes" -a "$transport_ofi" = "no" -a "$transport_portals4" = "no" ;
```

jdinan: Good catch!

src/transport_ucx.h
```diff
@@ -0,0 +1,846 @@
...
+int
+shmem_transport_ctx_create(struct shmem_internal_team_t *team, long options, shmem_transport_ctx_t **ctx)
+{
+    /* FIXME: Contexts are not optimized yet. We could create a separate set of
+     * EPs for each context (or team), and/or we could use completion callbacks
+     * to implement completion counters at the SOS level to enable separate
+     * completion tracking per context. */
+
+    if (team == NULL)
+        return 1;
```
wrrobin, line 188: Sorry if this is repetitive, but this is slightly different than OFI and Portals. In those cases, we raise error and abort. In this case, we are returning without error. In #941, I am proposing a check for the return value from the transport layer and assigning `SHMEMX_CTX_INVALID` in case of non-zero return value. I think we may need to unify these two behaviors, any of them should be fine. @davidozog @jdinan Your thoughts?

src/transport_ucx.h
```diff
@@ -0,0 +1,846 @@
...
+{
+    /* Use software reductions, instead of atomic vector operation */
+    return 0;
+}
+
+static inline
+void shmem_transport_syncmem(void)
+{
+    return;
+}
```
wrrobin, line 718: @jdinan Trying to understand, is it similar to libfabric where there is nothing similar to PtlAtomicSync or is it done in the hardware?

src/transport_ucx.h
```diff
@@ -0,0 +1,846 @@
...
+static inline
+void
+shmem_transport_put_signal_nbi(shmem_transport_ctx_t* ctx, void *target, const void *source, size_t len,
+                               uint64_t *sig_addr, uint64_t signal, int sig_op, int pe)
+{
+    shmem_transport_put_nbi(ctx, target, source, len, pe);
+    shmem_transport_fence(ctx);
+    switch (sig_op) {
+        case SHMEMX_SIGNAL_ADD:
+            shmem_transport_atomic(ctx, sig_addr, &signal, 8, pe, SHM_INTERNAL_SUM, SHM_INTERNAL_UINT64);
```
wrrobin, line 729: Is it safe to use 8 here?

wrrobin: Just caught the changes in the atomic routines.

configure.ac
jdinan: Pushed a fix for this and also added shr atomics to the configury output:
```
$ ../configure --with-xpmem=$HOME/opt/xpmem --enable-pmi-simple --prefix=/home/jdinan/opt/sos
...
Compilers:
  C:              gcc
  C++:            g++
  Fortran:        gfortran

Transports:
  Portals 4:      no
  OFI:            no
  UCX:            no

On Node Communication:
  XPMEM:          yes
  CMA:            no
  memcpy (self):  yes
  Shr. atomics:   yes

Global Options:
  Remote VA:      no
  Thread support: yes
  Error checking: no

Backtrace:        execinfo gdb
```

src/transport_ucx.h
jdinan: I hope so, it's a uint64_t. :) But I can change to `sizeof(unit64_t)` so the value is not arbitrary.

src/transport_ucx.h
jdinan: I'll update to match OFI and Portals. IIUC, team handle checks should have been resolved at the upper layer and it's an error if NULL reaches the transport layer.

src/transport_ucx.h
jdinan: I'm not aware of one. I'll ask the experts and let you know what I hear.

jdinan:
@wrrobin Thanks for the thorough review. I have a question out regarding `shmem_transport_syncmem` and will let you know what I hear.

wrrobin:
Thanks @jdinan. This PR looks good.

src/transport_ucx.h
jdinan: Nope, nothing special that needs to be done here.

jdinan:
@wrrobin @davidozog I responded to the question above about `shmem_transport_syncmem`; there is no change needed. @davidozog If this looks good to you, please go ahead and merge. Thanks!

PR #939, opened 2020-03-12
Set ctx to SHMEM_CTX_INVALID if team is invalid

davidozog:
I believe this is a bug.  The spec says,

> If team compares equal to SHMEM_TEAM_INVALID, then a nonzero value is returned and ctx is set to SHMEM_CTX_INVALID.


rdesai16:
@davidozog Nice catch!

PR #940, opened 2020-03-18
Allocation routines return if any of the arguments are passed zero value

wrrobin:
To comply with the spec, if size or count is 0, then the `shmem_calloc` routine should not be semantically equivalent to `shmem_barrier_all` 

### DIFF
--- a/src/symmetric_heap_c.c
+++ b/src/symmetric_heap_c.c
@@ -332,7 +332,7 @@ shmem_realloc(void *ptr, size_t size)
 
     SHMEM_ERR_CHECK_INITIALIZED();
 
-    if (size == 0) return ptr;
+    if (size == 0 && ptr == NULL) return ptr;
     if (ptr != NULL) {
       SHMEM_ERR_CHECK_SYMMETRIC_HEAP(ptr);
     }
--- a/src/symmetric_heap_c.c
+++ b/src/symmetric_heap_c.c
@@ -278,9 +278,11 @@ shmem_internal_shmalloc(size_t size)
 void SHMEM_FUNCTION_ATTRIBUTES *
 shmem_malloc(size_t size)
 {
-    void *ret;
+    void *ret = NULL;
 
     SHMEM_ERR_CHECK_INITIALIZED();
+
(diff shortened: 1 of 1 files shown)


davidozog:
@wrrobin - While looking at `shmem_malloc_with_hints`, I noticed that `shmem_malloc` relies on `dlmalloc` for a return value as well, and may not return NULL when `size` is 0.  Do we need to address that as well?

wrrobin:
@davidozog Thanks for bringing this up. I noticed it as well, but I could not find any corresponding change in the text in spec, similar to `calloc`. Let me know if you see it. We can also propose this change in the spec and handle it in SOS as well.

davidozog:
I see it in the 1.5rc1 version, "When size is zero, the shmem_malloc routine performs no action and returns a null pointer."  Same for `shmem_align`, and `shmem_realloc`.  Also later on, "When no action is performed, these routines return without performing a barrier."

wrrobin:
Thanks @davidozog. Will incorporate these then.

src/symmetric_heap_c.c
```diff
@@ -329,6 +331,8 @@ shmem_realloc(void *ptr, size_t size)
     void *ret;
 
     SHMEM_ERR_CHECK_INITIALIZED();
+
+    if (size == 0) return ptr;
```
davidozog: We may have to do a little more here - "If size is 0 and ptr is not a null pointer, the block to which it points is freed."

wrrobin: Thanks @davidozog. That case is already handled further down in the code. However, I changed the if statement here to check if the ptr is also NULL to make sure that the other options get checked. Let me know if you see any issues.

davidozog: Whoops!  Missed that - thanks!

wrrobin:
@rdesai16 Gentle reminder for the review. Please let me know if you have any concerns about this PR.

rdesai16:
Sorry for the delay. I had in mind that I had submitted the review.

davidozog:
Ready to merge @wrrobin?

PR #941, opened 2020-03-20
SHMEM_CTX_INVALID for RMA, AMO, and context creation operations

wrrobin:

### DIFF
--- a/src/atomic_c.c4
+++ b/src/atomic_c.c4
@@ -230,6 +230,7 @@ SHMEM_DEFINE_FOR_EXTENDED_AMO(`SHMEM_PROF_DEF_CTX_ATOMIC_SET')
         TYPE newval;                                            \
         SHMEM_ERR_CHECK_INITIALIZED();                          \
         SHMEM_ERR_CHECK_PE(pe);                                 \
+        SHMEM_ERR_CHECK_CTX(ctx);                               \
         SHMEM_ERR_CHECK_SYMMETRIC(target, sizeof(TYPE));        \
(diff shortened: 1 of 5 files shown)


wrrobin:
@davidozog Good catch, Dave! Although, I see the clause in the spec is little different for fence and quiet. It says, no operation is performed if the `ctx` is `SHMEM_CTX_INVALID`. May be, we need to add a check to return immediately from the internal layer?

davidozog:
@wrrobin Ah yes, very good call.  Looks like we might already have that, right?

In that case, we may not need the error check here...

wrrobin:
Ah... yeah, just saw these changes already done in the internal layer, thanks!

wrrobin:
@davidozog @rdesai16 Please check the latest commit in regards to #942. I am including these changes in this PR as well since they are very related.

davidozog:
@wrrobin - I thought @jdinan had preferred to keep constants like `SHMEM_CTX_INVALID` and `SHMEM_TEAM_INVALID` outside the transport layers, but I can't remember the exact reason (perhaps it's organizational to keep the user-facing constants as high up the stack as possible).  Perhaps he can remind us?

Assuming that's ok, I think the changes from the last commit are good.

jdinan:
At the transport layer, we tend to assume that it's a valid object creation request. It's generally good defensive programming to resolve these invalid/corner cases at the top level of the stack when possible. For example, the UCX code I posted also would need this fix if it's implemented at the transport level.

wrrobin:
Thanks @davidozog. Yes, going ahead.

PR #944, opened 2020-03-31
Fetch signal operation

wrrobin:


src/data_c.c4
```diff
@@ -628,6 +630,20 @@ SHMEM_DEFINE_FOR_RMA(`SHMEM_DEF_PUT_SIGNAL_NBI')
 SHMEM_DEFINE_FOR_SIZES(`SHMEM_DEF_PUT_N_SIGNAL_NBI')
 SHMEM_DEF_PUT_N_SIGNAL_NBI(`mem', `1')
 
+uint64_t SHMEM_FUNCTION_ATTRIBUTES
+shmemx_signal_fetch(const uint64_t* sig_addr)
```
jdinan, line 634: Woah, did we miss a context argument on this function?

davidozog: Yeah, I don't see why it wouldn't have one... 😕

wrrobin: @jdinan Thanks. Yeah, I am confused about whether it needs a context argument as it is only reading a local memory. Then again, it needs to keep atomicity guarantees. In that perspective, it needs a context argument. If it is only a simple load operation, then I am not sure why we need to have an extra API. I wonder why we did not see these through during the proposal discussion.

jdinan: It makes sense to me to implement as an atomic fetch, since it can be overlapped with signal atomic updates.

wrrobin:
@jdinan @davidozog Should we merge this PR as it is for now and return to add the `ctx` once it is there in spec? Thanks for your inputs.

jdinan:
The operation that goes into OpenSHMEM 1.5 does not have a context, so no need to block on an API change.

test/shmemx/signal_fetch.c
```diff
@@ -0,0 +1,90 @@
...
+        shmem_global_exit(1);
+    }
+
+    shmem_barrier_all();
+    for (i = 0; i < npes; i++) {
+        shmemx_long_put_signal(target, source, MSG_SZ, &sig_addr, me, SHMEMX_SIGNAL_ADD, i);
+    }
+
+    uint64_t sig_value = shmemx_signal_fetch(&sig_addr);
+    while (sig_value != ((npes * (npes - 1)) / 2)) {
```
davidozog, line 69: Pedantic, perhaps - but we can avoid a potential compiler warning here with a cast:
```suggestion
    while (sig_value != (uint64_t)((npes * (npes - 1)) / 2)) {
```

test/shmemx/signal_fetch.c
```diff
@@ -0,0 +1,90 @@
+/*
+ *  Copyright (c) 2018 Intel Corporation. All rights reserved.
```
davidozog, line 2: 2020?

test/shmemx/signal_fetch.c
wrrobin: Good suggestion @davidozog. Thanks.

wrrobin:
@rdesai16 Let me know if you see any concern on this PR. Thanks.

rdesai16:
Looks good to me!

mpp/shmemx_c_func.h4
```diff
@@ -191,6 +191,8 @@ define(`SHMEM_C_CTX_PUT_N_SIGNAL_NBI',
 SHMEM_DECLARE_FOR_SIZES(`SHMEM_C_CTX_PUT_N_SIGNAL_NBI')
 SHMEM_C_CTX_PUT_N_SIGNAL_NBI(mem,1);
 
+/* Signal fetch */
+SHMEM_FUNCTION_ATTRIBUTES uint64_t SHPRE()shmemx_signal_fetch(const uint64_t *sig_addr);
```
wrrobin, line 196: @davidozog I see it in the original PR here. Not sure how it is not merged?

davidozog: @wrrobin Oh, I see it too - this might be my mistake working with a stale file or something...  I'll check.

PR #949, opened 2020-04-29
Support for shmem_malloc_with_hints

rdesai16:


lanl-ompi:

test/shmemx/shmemx_malloc_with_hints.c
```diff
@@ -0,0 +1,111 @@
...
+
+    shmem_init();
+
+    npes = shmem_n_pes();
+    mype = shmem_my_pe();
+
+    for (i = 0; i < SHMEM_ALLTOALL_SYNC_SIZE; i++)
+        pSync[i] = SHMEM_SYNC_VALUE;
+
+    passed = sumtoall_with_malloc_hint(malloc_hint, mype, npes);
```
jdinan, line 95: Looks like `malloc_hint` is not initialized here?

src/symmetric_heap_c.c
```diff
@@ -396,3 +399,27 @@ void SHMEM_FUNCTION_ATTRIBUTES * shmemalign(size_t alignment, size_t size)
...
+shmemx_malloc_with_hints(size_t size, long hints)
+{
+    void *ret = NULL;
+
+    SHMEM_ERR_CHECK_INITIALIZED();
+
+    if (size == 0) return ret;
+
+    // Check for valid hints
+    if(hints != 0 && hints != SHMEMX_MALLOC_ATOMICS_REMOTE && hints!= SHMEMX_MALLOC_SIGNAL_REMOTE){
```
jdinan, line 414: I think `SHMEMX_MALLOC_ATOMICS_REMOTE | SHMEMX_MALLOC_SIGNAL_REMOTE` is also a valid hint. I think it would work to add `#define SHMEMX_MALLOC_MAX_HINT    (1l<<2)` and check `hints >= SHMEMX_MALLOC_MAX_HINT`.

test/shmemx/shmemx_malloc_with_hints.c
rdesai16: aah yes good catch! I will remove that

src/symmetric_heap_c.c
rdesai16: Yeah valid point! will fix that. thanks!

test/shmemx/shmemx_malloc_with_hints.c
```diff
@@ -0,0 +1,111 @@
...
+            if (dst[i] != expected) {
+                printf("[%d] dst[%d] = %d, expected %d\n", mype, i, dst[i], expected);
+                failed = 1;
+            }
+        }
+    }
+
+    if (failed)
+        return 0;
+    return 1;
```
wrrobin, line 77: We can use `return !failed` to get the same effect, I think.

test/shmemx/shmemx_malloc_with_hints.c
```diff
@@ -0,0 +1,111 @@
...
+#define MAX(a,b) ((a) > (b)) ? a : b
+#define WRK_SIZE MAX(N/2+1, SHMEM_REDUCE_MIN_WRKDATA_SIZE)
+
+#define NO_TESTS 4
+
+long pSync[SHMEM_ALLTOALL_SYNC_SIZE];
+int pWrk[WRK_SIZE];
+
+
+int *dst, *src;
```
wrrobin, line 43: Do we need dst and src to be global? May be, we can keep this within the scope of `sumtoall_with_malloc_hint`?

test/shmemx/shmemx_malloc_with_hints.c
```diff
@@ -0,0 +1,111 @@
...
+    if (mype == 0)
+        printf("Sum reduction operation started\n");
+
+    dst = (int *)shmemx_malloc_with_hints(N * sizeof(*dst), hint);
+    src = (int *)shmemx_malloc_with_hints(N * sizeof(*src), hint);
+
+    for (i = 0; i < npes; i++) {
+        src[i] = mype;
+        dst[i] = -1;
+    }
```
wrrobin, line 59: `dst` and `src` are allocated `N` size, but data is assigned only for `npes` indices. We should assign all `N` values, I think.

test/shmemx/shmemx_malloc_with_hints.c
davidozog: Looks like you meant to instantiate different hints combos before each call, right?

davidozog: I might suggest 5 hints:
1. `0`
2. `SHMEMX_MALLOC_ATOMICS_REMOTE`
3. `SHMEMX_MALLOC_SIGNAL_REMOTE`
4. `SHMEMX_MALLOC_ATOMICS_REMOTE | SHMEMX_MALLOC_SIGNAL_REMOTE`
5. Something invalid.

test/shmemx/shmemx_malloc_with_hints.c
rdesai16: aah I had it for N values but got replaced in later changes I think. Thanks for bringing this to my attention!

src/symmetric_heap_c.c
```diff
@@ -396,3 +399,27 @@ void SHMEM_FUNCTION_ATTRIBUTES * shmemalign(size_t alignment, size_t size)
...
+{
+    void *ret = NULL;
+
+    SHMEM_ERR_CHECK_INITIALIZED();
+
+    if (size == 0) return ret;
+
+    // Check for valid hints. If hint is invalid fall back to shmem_malloc and raise a debug warning
+    if(hints >= SHMEMX_MALLOC_MAX_HINTS && shmem_internal_my_pe == 0) {
+        DEBUG_STR("Invalid hint for shmem_malloc_with_hints. Valid hints are 0, SHMEMX_MALLOC_ATOMICS_REMOTE and SHMEMX_MALLOC_SIGNAL_REMOTE");
```
rdesai16, line 415: Instead of RAISE_WARN_STR I've used DEBUG_STR to clear up the output. Let me know what you guys @jdinan @davidozog @wrrobin think of this

jdinan: The macro will print the function name, file, and line number. Rather than enumerating the valid hints, I would tend to prefer something like:
```suggestion
        DEBUG_MSG("Invalid hint: 0x%lx", (unsigned long) hints);
```
But also, this is an application error so DEBUG doesn't seem right here because SOS will silently ignore the bad hints unless the user sets the SHMEM_DEBUG env var. I think I would prefer RAISE_WARN_MSG. I also would not mask the warning based on PE ID, in case the user also failed to provide the same hints at every PE. At some point, when the hints are actually used, this will become a more serious error.

rdesai16: Just to confirm my understanding, in case of bad hints, according to the definition of malloc_with_hints in spec SOS is supposed to ignore it, right? with RAISE_WARN_MSG the output gets filled with long warnings and it will be difficult to debug for users if multiple hints are  used unless we print the invalid hint with the warning. So i'll change it back to RAISE_WARN_STR and will also add the invalid hint in the output. Does that sound good?

rdesai16: I agree about the PE id condition. This issue had come up in my discussion with @davidozog . But I still added it thinking that as of now it won't be an issue and we can change it later when we add more functionality/hints to malloc_with_hints. If we're adding RAISE_WARN_STR then removing this condition will make the output too long for users to debug.

jdinan: This is what the spec has to say with regard to the hints argument: "The valid usage hints are described in Table 4. Multiple hints may be requested by combining them with a bitwise OR operation. A zero option can be given if no options are requested."

I suppose it is a matter of interpretation what to do if the hints are not all valid?

rdesai16: yeah that's true.It's kind of unclear in case of invalid hints . So I'll change this to RAISE_WARN_MSG and will remove the PE id condition so we don't have to worry about that in future.

src/symmetric_heap_c.c
```diff
@@ -396,3 +399,27 @@ void SHMEM_FUNCTION_ATTRIBUTES * shmemalign(size_t alignment, size_t size)
...
+{
+    void *ret = NULL;
+
+    SHMEM_ERR_CHECK_INITIALIZED();
+
+    if (size == 0) return ret;
+
+    // Check for valid hints
+    if(hints >= SHMEMX_MALLOC_MAX_HINTS) {
+        RAISE_WARN_STR("%ld is and invalid hint for shmem_malloc_with_hints. Valid hints can be 0, SHMEMX_MALLOC_ATOMICS_REMOTE, SHMEMX_MALLOC_SIGNAL_REMOTE or a combination of these three");
```
davidozog, line 415: Looks like a "hanging" `%ld`, probably want to use `RAISE_WARN_MSG`?

src/symmetric_heap_c.c
```diff
@@ -396,3 +399,27 @@ void SHMEM_FUNCTION_ATTRIBUTES * shmemalign(size_t alignment, size_t size)
...
+shmemx_malloc_with_hints(size_t size, long hints)
+{
+    void *ret = NULL;
+
+    SHMEM_ERR_CHECK_INITIALIZED();
+
+    if (size == 0) return ret;
+
+    // Check for valid hints
+    if(hints >= SHMEMX_MALLOC_MAX_HINTS) {
```
davidozog, line 414: Also need to check whether hints are less than 0?

src/symmetric_heap_c.c
```diff
@@ -411,8 +411,8 @@ shmemx_malloc_with_hints(size_t size, long hints)
     if (size == 0) return ret;
 
     // Check for valid hints
-    if(hints >= SHMEMX_MALLOC_MAX_HINTS) {
-        RAISE_WARN_STR("%ld is and invalid hint for shmem_malloc_with_hints. Valid hints can be 0, SHMEMX_MALLOC_ATOMICS_REMOTE, SHMEMX_MALLOC_SIGNAL_REMOTE or a combination of these three");
+    if(hints >= SHMEMX_MALLOC_MAX_HINTS || hints < 0) {
+        RAISE_WARN_MSG("%ld is an invalid hint for shmem_malloc_with_hints. Valid hints can be 0, SHMEMX_MALLOC_ATOMICS_REMOTE, SHMEMX_MALLOC_SIGNAL_REMOTE or a combination of these three\n", hints);
```
davidozog, line 415: The `_MSG` macros include the `\n`, so no need for it here.

test/shmemx/shmemx_malloc_with_hints.c
```diff
@@ -0,0 +1,111 @@
...
+    shmem_int_sum_to_all(dst, src, N, 0, 0, npes, pWrk, pSync);
+
+    if(mype == 0){
+        for (i = 0; i < npes; i++) {
+            int expected;
+            expected = npes * (npes-1)/2;
+            if (dst[i] != expected) {
+                printf("[%d] dst[%d] = %d, expected %d\n", mype, i, dst[i], expected);
+                failed = 1;
+            }
```
wrrobin, line 72: Sorry for late comments. But I have 2 here.
1. minor: if<space>(mype == 0). Similar comment to Dave's earlier one. I think there is another instance below.. 
2. Shouldn't we check for all the N values in the for loop?

src/symmetric_heap_c.c
rdesai16: I had thought so but when I ran the test it didn't include \n in the output. I'll check again and remove it. Thanks!

test/shmemx/shmemx_malloc_with_hints.c
rdesai16: good points! 2nd point was overlook on my part. It should be checked for N values. I'll make the changes. Thanks!

src/symmetric_heap_c.c
rdesai16: Actually RAISE_WARN_MSG does not include ' \n' in it which is kind of strange because it should. So I'll keep this '\n' here. There might be some important reason for RAISE_WARN_MSG to not have \n in it so I would not change that.

jdinan: The `_STR` macros include the newline. `_MSG` macros assume you will format the message.

src/symmetric_heap_c.c
```diff
@@ -396,3 +399,27 @@ void SHMEM_FUNCTION_ATTRIBUTES * shmemalign(size_t alignment, size_t size)
...
+shmemx_malloc_with_hints(size_t size, long hints)
+{
+    void *ret = NULL;
+
+    SHMEM_ERR_CHECK_INITIALIZED();
+
+    if (size == 0) return ret;
+
+    // Check for valid hints
+    if(hints >= SHMEMX_MALLOC_MAX_HINTS || hints < 0) {
```
jdinan, line 414: `hints == SHMEMX_MALLOC_MAX_HINTS` is ok, right? Does the unit test check the `SHMEMX_MALLOC_ATOMICS_REMOTE | SHMEMX_MALLOC_SIGNAL_REMOTE` case?

src/symmetric_heap_c.c
davidozog: My bad - thanks.

src/symmetric_heap_c.c
```diff
@@ -396,3 +399,27 @@ void SHMEM_FUNCTION_ATTRIBUTES * shmemalign(size_t alignment, size_t size)
...
+        RAISE_WARN_MSG("%ld is an invalid hint for shmem_malloc_with_hints. Valid hints can be 0, SHMEMX_MALLOC_ATOMICS_REMOTE, SHMEMX_MALLOC_SIGNAL_REMOTE or a combination of these three\n", hints);
```
jdinan, line 415:
```suggestion
        RAISE_WARN_MSG("Ignoring invalid hint for shmem_malloc_with_hints (%ld)\n", hints);
```
If it was me, I would use the above. The existing message is too long to fit on a line (you can break it, there is a macro for the leading whitespace to align multi-line warnings) but also it will eventually go out of sync with what hints are allowed.

test/shmemx/shmemx_malloc_with_hints.c
```diff
@@ -0,0 +1,111 @@
...
+#include <shmem.h>
+#include <shmemx.h>
+
+#define N 128
+#define MAX(a,b) ((a) > (b)) ? a : b
+#define WRK_SIZE MAX(N/2+1, SHMEM_REDUCE_MIN_WRKDATA_SIZE)
+
+#define NO_TESTS 5
+
+#define SHMEMX_MALLOC_INVALID_HINT 1l<<3
```
jdinan, line 39: There's a good chance `1l<<3` will be valid in some implementation. How about something like `~(SHMEMX_MALLOC_ATOMICS_REMOTE)`?

test/shmemx/shmemx_malloc_with_hints.c
```diff
@@ -0,0 +1,111 @@
...
+#include <stdio.h>
+#include <stdint.h>
+#include <shmem.h>
+#include <shmemx.h>
+
+#define N 128
+#define MAX(a,b) ((a) > (b)) ? a : b
+#define WRK_SIZE MAX(N/2+1, SHMEM_REDUCE_MIN_WRKDATA_SIZE)
+
+#define NO_TESTS 5
```
jdinan, line 37: How about `NUM_TESTS`?

jdinan: NO_TESTS confused me for a moment. :)

rdesai16: good point! will change it:)

src/symmetric_heap_c.c
rdesai16: aah got it! I did thought that this will seem like long for a warning.

test/shmemx/shmemx_malloc_with_hints.c
rdesai16: oh okay I didn't know about that. Yeah makes sense to use something ~(SHMEMX_MALLOC_ATOMICS_REMOTE).

src/symmetric_heap_c.c
rdesai16: yeah the unit tests does check for the combination so SHMEMX_MALLOC_MAX_HINTS is also a valid hint.

jdinan:
```suggestion
    if(hints > SHMEMX_MALLOC_MAX_HINTS || hints < 0) {
```
It looks like `SHMEMX_MALLOC_MAX_HINTS == SHMEMX_MALLOC_ATOMICS_REMOTE | SHMEMX_MALLOC_SIGNAL_REMOTE`, and this is a valid hint. Am I reading this right?

rdesai16: Yeah that's right! I'll fix this.

rdesai16:
@jdinan @wrrobin @davidozog I've added all the changes suggested in the review comments. Let me know if something's missing.

davidozog:
Thanks @rdesai16 

FYI - the failed rows in Travis are from a PMIx / libevent problem that @rhc54 is looking into, and I think we can probably ignore for now.

src/symmetric_heap_c.c
davidozog: Is this issue resolved?  I'm not seeing the check against the "maximum" possible hint value.

test/shmemx/shmemx_malloc_with_hints.c
```diff
@@ -0,0 +1,111 @@
...
+        pSync[i] = SHMEM_SYNC_VALUE;
+
+    passed = sumtoall_with_malloc_hint(0, mype, npes);
+    passed += sumtoall_with_malloc_hint(SHMEMX_MALLOC_ATOMICS_REMOTE, mype, npes);
+    passed += sumtoall_with_malloc_hint(SHMEMX_MALLOC_SIGNAL_REMOTE, mype, npes);
+    passed += sumtoall_with_malloc_hint(SHMEMX_MALLOC_ATOMICS_REMOTE | SHMEMX_MALLOC_SIGNAL_REMOTE, mype, npes);
+    passed += sumtoall_with_malloc_hint(SHMEMX_MALLOC_INVALID_HINT, mype, npes);
+
+
+    fail = NO_TESTS - passed;
```
davidozog, line 100: This should be `NUM_TESTS` now, right? (and below)

src/symmetric_heap_c.c
```diff
@@ -64,6 +64,9 @@
 #pragma weak shfree = pshfree
 #define shfree pshfree
 
+#pragma weak shmemx_malloc_with_hints = pshmem_malloc_with_hints
+#define shmemx_malloc_with_hints pshmem_malloc_with_hints
+
```
wrrobin, line 69: @davidozog @rdesai16 It needs to be `pshmemx_...`, right?

davidozog: Yes, it should (at least, that would be consistent with what we've done in the past).

Great catch @wrrobin!

rdesai16: yeah good point! will change that.

rdesai16:
@jdinan I've made all the required changes so let me know if you see something's missing. Thanks!

jdinan:
No need to wait on my approval, looks like you have all done excellent work as usual! :shipit:

Issue #952, opened 2020-06-11
Missing prototype for shmemx_signal_fetch


davidozog:
I see this warning with `-Wall -Wextra`:

```
  CC       data_c.lo
data_c.c:2162:1: warning: no previous prototype for function 'shmemx_signal_fetch' [-Wmissing-prototypes]
shmemx_signal_fetch(const uint64_t* sig_addr)
^
1 warning generated.
```

davidozog:
Whoops - false alarm!

PR #953, opened 2020-06-11
Add missing prototype for shmemx_signal_fetch

davidozog:
This should fix issue #952


davidozog:
Looks like it's already here:
https://github.com/Sandia-OpenSHMEM/SOS/pull/953/files#diff-2f5480a67ddac885891c7cc3bb552e33L195

Sorry for the false alarm @wrrobin.  I'm not sure what state my local repo was in when I ran into this, but it's my fault.

Issue #954, opened 2020-06-11
Autoconf picking up UCX and OFI


jdinan:
Reported to me by a user --- It looks like there is an issue with autoconf picking up and trying to build both UCX and OFI:
```
CC       cache_management_c.lo
In file included from transport.h:55:0,
                 from transport_ucx.h:17,
                 from transport_ucx.c:15:
transport_ofi.h:197:38: error: redeclaration of enumerator ‘FI_BAND’
 #define SHM_INTERNAL_BAND            FI_BAND
                                      ^
transport_ucx.h:24:5: note: in expansion of macro ‘SHM_INTERNAL_BAND’
     SHM_INTERNAL_BAND,
     ^~~~~~~~~~~~~~~~~
In file included from transport_ofi.h:20:0,
                 from transport.h:55,
                 from transport_ucx.h:17,
                 from transport_ucx.c:15:
```
The user was trying to build with OFI support, but autoconf seems to auto-detected UCX support and enabled both. The fix was to add `--disable-ucx` to the configure command.

A change like this might fix it:
```
diff --git a/configure.ac b/configure.ac
index 682745b7..6de1baae 100755
--- a/configure.ac
+++ b/configure.ac
@@ -369,12 +369,18 @@ fi
 if test -n "$with_portals4" -a "$with_portals4" != "no" ; then
     transport="portals4"
+    transport_ucx="no"
+    transport_ofi="no"
     AC_DEFINE([USE_PORTALS4], [1], [Define if Portals 4 transport active])
 elif test -n "$with_ofi" -a "$with_ofi" != "no" ; then
     transport="ofi"
+    transport_ucx="no"
+    transport_portals="no"
     AC_DEFINE([USE_OFI], [1], [Define if OFI transport active])
 elif test -n "$with_ucx" -a "$with_ucx" != "no" ; then
     transport="ucx"
+    transport_ofi="no"
+    transport_portals="no"
     AC_DEFINE([USE_UCX], [1], [Define if UCX transport active])
     AC_DEFINE([ENABLE_HARD_POLLING], [1], [Enable hard polling])
 else
```

davidozog:
Closed by #959

PR #959, opened 2020-07-28
Resolve configury issue with multiple transports detected

davidozog:
I encountered this problem when testing Portals4 on a system with libfabric automatically detected.  It's the same issue as #954, which should be resolved by this PR.

Thanks for the patch @jdinan!

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -355,31 +355,39 @@ OMPI_CHECK_UCX([ucx],
 num_transports=""
 if test -n "$with_portals4" -a "$with_portals4" != "no" ; then
     num_transports="x$num_transports"
+fi
+if test -n "$with_ofi" -a "$with_ofi" != "no" ; then
+    num_transports="x$num_transports"
+fi
+if test -n "$with_ucx" -a "$with_ucx" != "no" ; then
+    num_transports="x$num_transports"
+fi
+
+if test -n "$num_transports" -a "$num_transports" != "x" ; then
+    AC_MSG_ERROR([Cannot select more than one transport, see --help for details])
+fi
+
+if test -n "$with_portals4" -a "$with_portals4" != "no" ; then
     transport="portals4"
     transport_ofi="no"
     transport_ucx="no"
     AC_DEFINE([USE_PORTALS4], [1], [Define if Portals 4 transport active])
-fi
-if test -n "$with_ofi" -a "$with_ofi" != "no" ; then
-    num_transports="x$num_transports"
+elif test -n "$with_ofi" -a "$with_ofi" != "no" ; then
     transport="ofi"
     transport_portals="no"
     transport_ucx="no"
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -369,12 +369,18 @@ fi
...
 elif test -n "$with_ofi" -a "$with_ofi" != "no" ; then
     transport="ofi"
+    transport_portals="no"
+    transport_ucx="no"
     AC_DEFINE([USE_OFI], [1], [Define if OFI transport active])
 elif test -n "$with_ucx" -a "$with_ucx" != "no" ; then
     transport="ucx"
+    transport_ofi="no"
+    transport_portals="no"
     AC_DEFINE([USE_UCX], [1], [Define if UCX transport active])
```
wrrobin, line 384: Looks good to me. Just a thought, it seems we can merge this if-else block to the multiple if statements above. We are any way error'ing out when multiple transports are selected using `num_transport`.

davidozog: @wrrobin I did the merge you suggested, and I think it's clearer this way.

@jdinan @rdesai16 - might want to do a quick re-review if possible to make sure we didn't make a mistake.

configure.ac
```diff
@@ -355,33 +355,31 @@ OMPI_CHECK_UCX([ucx],
...
-elif test -n "$with_ucx" -a "$with_ucx" != "no" ; then
-    transport="ucx"
-    AC_DEFINE([USE_UCX], [1], [Define if UCX transport active])
-    AC_DEFINE([ENABLE_HARD_POLLING], [1], [Enable hard polling])
 else
     transport="none"
-    transport_portals4="no"
-    transport_ofi="no"
-    transport_ucx="no"
     AC_MSG_WARN([No transport requested])
```
jdinan, line 383: It looks like the `else` case will now be triggered even when a transport is selected successfully.

davidozog: Yikes... thanks.

wrrobin: Yeah, it should be an `else`. We can check if num_transports is empty and then we assign transport as none.

davidozog: @jdinan @wrrobin - Sorry about that... does the new version look correct?

wrrobin: @davidozog Thanks. Looks good to me.

jdinan: I think this is ok. I have a slight preference for the old structure, which did the error check for multiple transports before setting up the transport, but I think this new code should also work.

davidozog: Yeah I had a similar thought after this last change - it's "6 of one half-dozen of the other" in my opinion.

I think I'll just revert to the first commit on this PR, unless @wrrobin has any qualms.

wrrobin: I am ok either way. Current plan sounds good.

davidozog:
I reverted to the original commit.  Can I get a +1 on a squash and merge?

PR #964, opened 2020-08-03
Disabling PMIx tests on Travis temporarily

wrrobin:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,11 +36,11 @@ env:
           SOS_DISABLE_EXTERNAL_TESTS=1
           SOS_TRANSPORT_OPTS="--with-portals4=$TRAVIS_INSTALL/portals4/"
           SOS_BUILD_OPTS="--enable-pmi-simple"
-        - >
-          SOS_DISABLE_EXTERNAL_TESTS=1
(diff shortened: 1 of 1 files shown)


jdinan:
What is the issue with PMIx? Could this be fixed by updating Travis to use the last known good version of PMIx/PRRTE rather than disabling testing?

davidozog:
@jdinan - There are a lot of details, but in short, there is a build issue with prrte `v1.0.0` on Ubuntu 18.04, (so for that a a few other pmix-compatibility reasons) we need to update prrte to master (there is no recent prrte tag and likely will not be for a little while).  The other issue is with the `libev-dev` apt package(s) - they won't work with prrte master at the moment, so we'd need to move back to libevent.

While writing an issue describing this yesterday I stumbled upon a configuration that works in Ubuntu 18.04 - prrte/pmix masters, both linked with libevent 2.1.10.  I think I'll push that through Travis as one last ditch effort, but if this doesn't work out we might need to cut our losses and fix it later.  We've spent too much time on this already - it's a bit of a whack-a-mole problem.

davidozog:
I drafted an alternative to disabling all PRRTE/PMIx rows in #966. 

There are trade-offs, and I'm open to either approach.  Any preferences from @wrrobin @jdinan @rdesai16 @rhc54 ?

wrrobin:
Workaround by #966

PR #965, opened 2020-08-03
Fix completion logic bug

### LINKED ISSUE #958: Deadlock in `zero_comms` with portals transport
This came up while testing the Portals transport for a v1.5.0 release: 

Unit test `zero_comms` (doing a shmem_putmem with len=0 and src=dest=NULL) appears to be stuck in `PtlEQWait` after a small change to support individually completed puts in UCX in this commit: https://github.com/Sandia-OpenSHMEM/SOS/commit/64885a10a97ba04549626a5a124757c1d9a65ae7

It appears to only be an issue when bounce buffering is enabled.  In Portals, the default context seems to be hard-coded to do bounce buffering (even when passing --disable-bounce-buffering), but I'm not sure why...  If I remove that hard-coded option, then all seems ok, so perhaps that could be a quick fix to this issue?

Otherwise, reverting the len==0 change in https://github.com/Sandia-OpenSHMEM/SOS/commit/64885a10a97ba04549626a5a124757c1d9a65ae7 will likely break zero-length puts in UCX.  Is that true @jdinan?

On that note, the spec suggests len=0 and ptr=NULL should have defined behavior, but not necessarily that any communication/completion event must occur.  Unless I'm missing something?  I don't see much clarification in the RMA section, but perhaps it's elsewhere.

Anyway, any other thoughts on this?  Would it be ok to disable bounce buffering on the default context in Portals?

jdinan:
Closes #958

### DIFF
--- a/src/shmem_comm.h
+++ b/src/shmem_comm.h
@@ -49,10 +49,8 @@ void
 shmem_internal_put_nb(shmem_ctx_t ctx, void *target, const void *source, size_t len, int pe,
                       long *completion)
 {
-    if (len == 0) {
-        *completion = 1;
+    if (len == 0)
         return;
-    }
 
     if (shmem_shr_transport_use_write(ctx, target, source, len, pe)) {
         shmem_shr_transport_put(ctx, target, source, len, pe);
--- a/src/transport_ucx.h
+++ b/src/transport_ucx.h
@@ -128,7 +128,7 @@ ucs_status_t shmem_transport_ucx_release_op(ucs_status_ptr_t req) {
 static inline
 ucs_status_t shmem_transport_ucx_post_cb_op(ucs_status_ptr_t req, void *completion) {
     if (req == NULL) {
-        __atomic_store_n((long*)completion, 1, __ATOMIC_RELEASE);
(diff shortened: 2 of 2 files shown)


jdinan:
Nope, that is a pointer to `completion`, not its value.

jdinan:
Curious about these testing failures. Are any of those new with this PR? For example, #964 suggests there is a PMIx issue and I see multiple timeouts on CI for that PR.

davidozog:
@jdinan - I'm happy to approve/merge this now if you prefer.  But I was thinking it might be better to review/merge #966 first, especially if you want to include UCX in the SOS transport tests (please see my [comment there](https://github.com/Sandia-OpenSHMEM/SOS/pull/966#issuecomment-669304351)).  I'm testing that now on my fork, and I think it's passing everything.

jdinan:
@davidozog It would be good to enable the UCX tests.

davidozog:
@jdinan - I think this is ready to merge when you're ready.  We need it for portals to run, anyway.

jdinan:
@davidozog Good to merge.

PR #966, opened 2020-08-05
Slim down Travis CI

davidozog:
This includes various updates to make Travis go green again.

An alternative to the PRRTE/PMIx changes below is to disable it entirely temporarily, as in #964.

Here is a summary of the changes:
* Update PRRTE launch commands to latest version
* Checkout stable PRRTE/PMIx commits in Xenial Travis
* Remove Portals+PMIx Travis row (libev is required by Portals, but not working with latest PRRTE/PMIx)
* Switch to libevent 2.1.10-stable for PRRTE/PMIx
* ~~Force UH tests to use prun in PMIx rows~~
* Remove external tests from UCX rows in Travis
* Remove external test from manual-progress row in Travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -121,14 +121,14 @@ env:
           SOS_TRANSPORT_OPTS="--with-ofi=$TRAVIS_INSTALL/libfabric/"
           SOS_BUILD_OPTS="--enable-error-checking --enable-remote-virtual-addressing --enable-pmi-simple"
         - >
-          SOS_ENABLE_ERROR_TESTS=1
-          SHMEM_BARRIER_ALGORITHM=dissem SHMEM_BCAST_ALGORITHM=tree SHMEM_REDUCE_ALGORITHM=tree
+          SOS_DISABLE_EXTERNAL_TESTS=1 SOS_ENABLE_ERROR_TESTS=1
+          SHMEM_BARRIER_ALGORITHM=tree SHMEM_BCAST_ALGORITHM=tree SHMEM_REDUCE_ALGORITHM=tree
           SHMEM_OFI_STX_MAX=0
           SOS_TRANSPORT_OPTS="--with-ofi=$TRAVIS_INSTALL/libfabric/"
           SOS_BUILD_OPTS="--enable-error-checking --enable-remote-virtual-addressing --enable-pmi-simple --enable-manual-progress"
         - >
           SOS_DISABLE_EXTERNAL_TESTS=1 SOS_ENABLE_ERROR_TESTS=1
-          SHMEM_BARRIER_ALGORITHM=tree SHMEM_REDUCE_ALGORITHM=recdbl SHMEM_FCOLLECT_ALGORITHM=recdbl
+          SHMEM_BARRIER_ALGORITHM=dissem SHMEM_REDUCE_ALGORITHM=recdbl SHMEM_FCOLLECT_ALGORITHM=recdbl
           SHMEM_OFI_STX_AUTO=1
           SOS_TRANSPORT_OPTS="--with-ofi=$TRAVIS_INSTALL/libfabric/"
           SOS_BUILD_OPTS="--enable-error-checking --enable-remote-virtual-addressing --enable-pmi-simple --enable-manual-progress --enable-hard-polling"
--- a/.travis.yml
+++ b/.travis.yml
@@ -156,6 +156,7 @@ env:
 
         ## UCX Builds
         - >
+          SOS_DISABLE_EXTERNAL_TESTS=1
           SOS_TRANSPORT_OPTS="--with-ucx=$TRAVIS_INSTALL/ucx/"
           SOS_BUILD_OPTS="--enable-pmi-simple"
         - >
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -367,6 +379,11 @@ script:
     - export OSHRUN_LAUNCHER="$SOS_PM"
     - cd $TRAVIS_SRC/tests-uh
     - make $TRAVIS_PAR_MAKE C_feature_tests
+    - >
+      if [[ $SOS_BUILD_OPTS = *"with-pmix"* ]]; then
+          sed -i '/RUN_CMD=\"oshrun\"/c\our $RUN_CMD=\"\";' common/testrunner.pl
+          export RUN_OPTIONS=$SOS_PM
+      fi
     - $SOS_PM_PRE
```
wrrobin, line 387: @davidozog We have already disabled external tests for pmix, right? UH tests should not even run.

.travis.yml
```diff
@@ -287,28 +288,39 @@ before_install:
...
+    - cd libevent-2.1.10-stable
+    - ./autogen.sh
+    - ./configure --prefix=$PWD/install
+    - make clean all install
     ## Build PMIx
     - >
       if [[ $SOS_BUILD_OPTS = *"with-pmix"* ]]; then
         cd $TRAVIS_SRC
         git clone --depth 10 https://github.com/pmix/pmix pmix
+        git checkout e4a27268
```
wrrobin, line 304: Is this necessary or just for safety?

.travis.yml
davidozog: Oh yeah, good call.  This is what's needed to get `prun` working with `tests-uh`, but there just isn't enough time to get through all the external tests so I ended up taking it out.  I'll revert this so it's at least in our commit history if we need it later.

.travis.yml
davidozog: Yeah, more for stability.  We could take it out if you prefer, or leave a TODO to take the next PRRTE/PMIx tag combo that works in Xenial.  There isn't a tag on PRRTE we can use at the moment.

wrrobin: No, I am ok to keep it. May be we can leave a comment here so that we remember to check it out later if we need to.

.travis.yml
```diff
@@ -36,11 +36,6 @@ env:
           SOS_DISABLE_EXTERNAL_TESTS=1
           SOS_TRANSPORT_OPTS="--with-portals4=$TRAVIS_INSTALL/portals4/"
           SOS_BUILD_OPTS="--enable-pmi-simple"
-        - >
-          SOS_DISABLE_EXTERNAL_TESTS=1
-          SOS_PM="prun -host localhost -oversubscribe" SOS_PM_PRE="prte --daemonize" SOS_PM_POST="prun -terminate"
-          SOS_TRANSPORT_OPTS="--with-portals4=$TRAVIS_INSTALL/portals4/"
-          SOS_BUILD_OPTS="--with-pmix=$TRAVIS_INSTALL/pmix"
         - >
```
wrrobin, line 39: For Portals, we are removing the PMIx test because of existing issues, right? I am wondering whether we would still keep them commented for any future reasons. Either way is fine though.

davidozog: Ah thanks, that reminds me, I was planning to create an SOS github issue to track it.  Commenting it out would work too, but I have a feeling it might take some time for this to get resolved.  I would argue it's pretty low priority for all parties involved.

Any preference?

davidozog:
An aside comment:  I believe the UCX rows are not running the SOS tests because of this block:
```
      if [[ $SOS_TRANSPORT_OPTS = *"with-ofi"* ]]; then
        make VERBOSE=1 TEST_RUNNER="$SOS_PM -np 2" check
      else
        $SOS_PM -np 1 test/unit/hello
      fi
```
but I think it would be valuable to include UCX here, especially since we can't include the external tests.  Something like:
```
-      if [[ $SOS_TRANSPORT_OPTS = *"with-ofi"* ]]; then
+      if [[ $SOS_TRANSPORT_OPTS = *"with-ofi"* || $SOS_TRANSPORT_OPTS = *"with-ucx"* ]]; then
```
@jdinan - do you want to include that change?  It might give some peace-of-mind on PR #965, for example.

.travis.yml
jdinan: I would prefer to comment this out. This is a particularly tricky case and it's good to test.

.travis.yml
jdinan: Comment here would be helpful.

rhc54:
Is it true that libev is not working with PMIx? I know it doesn't work with PRRTE, but that is a different story. If it is working with PMIx, then another solution is to:

* checkout and build a copy of PMIx using libevent
* checkout and build PRRTE using libevent and the above copy of PMIx, configuring PRRTE `--enable-static --disable-shared`
* rebuild PMIx using libev
* build SOS against libev and the rebuilt version of PMIx

This would resolve the problem while preserving your testing range.

davidozog:
@rhc54 I'm not 100% sure.  I have a static build with a backtrace that _might_ suggest a libev issue with PMIx, but I can't say for sure.  Looks like it could be something else, like `--disable-shared` not working somewhere along the line.  I put this into a separate issue #967 we can discuss there. 

I haven't tried rebuilding PMIx using libev... that's interesting.  I'll let you know what happens.

davidozog:
Thanks for the feedback @jdinan @rhc54 @wrrobin @rdesai16.

@rhc54's suggestion to rebuild PMIx with libev actually works quite well - it seems like a reasonable workaround to me...  

Please review this ticket when you can and let me know if we are ok with merging this.

davidozog:
As per #965, I've added in the SOS tests for the UCX transport in commit ce346d6.

.travis.yml
```diff
@@ -338,7 +367,7 @@ script:
     - make $TRAVIS_PAR_MAKE check TESTS=
     - $SOS_PM_PRE
     - >
-      if [[ $SOS_TRANSPORT_OPTS = *"with-ofi"* ]]; then
+      if [[ $SOS_TRANSPORT_OPTS = *"with-ofi"* ]] || [[ $SOS_TRANSPORT_OPTS = *"with-ucx" ]]; then
```
wrrobin, line 370: We might need a `*` at the end of `"with-ucx"` as the flag will definitely have a value.

davidozog: Whoops!  Thanks - it probably wasn't running then...

davidozog:
Darn, looks like there's a hang on the `max_reduction` SOS test in the last 2 Travis rows with UCX.

Any ideas what's causing that @jdinan / @akhillanger?

wrrobin:
@davidozog Would it be ok if we postpone the UCX tests until the final release? Thinking, may be we can skip it for rc1

davidozog:
@wrrobin Either way is fine with me, I think we can let @jdinan decide this one.  I don't think we _need_ all the tests to pass, it's just convenient when they do ;)

jdinan:
I do expect the UCX transport to actually work and pass all the tests. :)

Could you split out the part that enables the UCX tests as a separate PR and assign it to me? I'll try to find some time to look into the issue.

davidozog:
Thanks @jdinan, will do.

davidozog:
@jdinan - I assume you want to include the #965 (completion bug fix) in that PR?  I think it might help narrow down the problem since that fix pushes Travis along much further.

Issue #963, opened 2020-07-29
shmem_internal_team_release_psyncs impl is broken


akhillanger:
For SYNC op, there is only one array and that can used back-to-back w/o waiting for prior SYNC ops to complete.
For REDUCE, you have more than one psync array per team and it is chosen selectively using choose_psync fn. Hence you need to selectively release them also.

Also, I see that there is no release_psync fn corresponding to choose_psync at line - https://github.com/Sandia-OpenSHMEM/SOS/blob/master/src/shmem_team.c#L277

Tagging @davidozog and @jdinan

davidozog:
@akhillanger Wow, good catch.  IIUC, it looks like releasing psyncs for REDUCE/BCAST/COLLECT/ALLTOALL doesn't actually do _anything_... This must have been an oversight on my part. 🤦  But SYNC is ok, right?

Off the cuff, it seems like it might help to carry around an index to help facilitate selectively choosing/releasing.  Something like:
```
long* choose_psync(shm_team_t *team, shm_team_op_t op, size_t *idx)
void release_psyncs(shm_team_t *team, shm_team_op_t op, const size_t idx)
```
where `choose` sets `*idx` and `release` resets `team->psync_avail[idx]` to 1.  `idx` isn't needed with SYNC, so could be NULL.  Seems slightly kludgy, so perhaps there's a better approach you can think of?

And yeah, after we fix this there should be a release corresponding to shmem_team.c#L277

Please feel free to post an alternative solution.  Sounds like you might be more familiar with the code than I am at the moment - I'm having to remind myself of many things...

akhillanger:
> @akhillanger Wow, good catch. IIUC, it looks like releasing psyncs for REDUCE/BCAST/COLLECT/ALLTOALL doesn't actually do _anything_... This must have been an oversight on my part. 🤦 But SYNC is ok, right?
> 
Yes, that is correct David. Just ensure that your sync implementation is ok using the same psync array for back-to-back sync collective.
> Off the cuff, it seems like it might help to carry around an index to help facilitate selectively choosing/releasing. Something like:
> 
> ```
> long* choose_psync(shm_team_t *team, shm_team_op_t op, size_t *idx)
> void release_psyncs(shm_team_t *team, shm_team_op_t op, const size_t idx)
> ```
> 
[...]

Issue #960, opened 2020-07-28
Some bugs in teams implementation


akhillanger:
1. https://github.com/Sandia-OpenSHMEM/SOS/blob/master/src/shmem_team.c#L193
Should be `shmem_internal_psync_pool = NULL`, similarly in L197,L201
2. https://github.com/Sandia-OpenSHMEM/SOS/blob/master/src/shmem_internal.h#L550
Should be  `return (ptr[index / CHAR_BIT] >> (index % CHAR_BIT)) & 1`;
3. https://github.com/Sandia-OpenSHMEM/SOS/blob/master/src/shmem_team.c#L491
Should be `return &shmem_internal_psync_pool[team->psync_idx * PSYNC_CHUNK_SIZE]`;

Tagging @davidozog and @jdinan

davidozog:
Awesome - thanks @akhillanger

These fixes all look correct to me at a glance.  Out of curiosity, did you run into any specific problems you can share?  I see that these bugs have some very undesirable side effects...  Perhaps we could add some better unit tests to cover team creation/resource management.

Would you mind posting a PR with the fixes?  We should include these in v1.5.0, which we're hoping to release pretty soon.

akhillanger:
David, I came across these while trying to understand the code. I don't have test cases yet. Here is a patch with these fixes.

diff --git a/src/shmem_internal.h b/src/shmem_internal.h
```
index 3583b17..4e62f20 100644
--- a/src/shmem_internal.h
+++ b/src/shmem_internal.h
@@ -547,7 +547,7 @@ void shmem_internal_bit_clear(unsigned char *ptr, size_t size, size_t index)
 static inline
 unsigned char shmem_internal_bit_fetch(unsigned char *ptr, size_t index)
 {
-    return (ptr[index / CHAR_BIT] >> index) & 1;
+    return (ptr[index / CHAR_BIT] >> (index % CHAR_BIT)) & 1;
 }

 static inline
diff --git a/src/shmem_team.c b/src/shmem_team.c
index 0fa418d..56f352d 100644
--- a/src/shmem_team.c
+++ b/src/shmem_team.c
@@ -190,15 +190,15 @@ cleanup:
     }
     if (shmem_internal_psync_pool) {
         shmem_internal_free(shmem_internal_psync_pool);
-        shmem_internal_team_pool = NULL;
+        shmem_internal_psync_pool = NULL;
     }
     if (psync_pool_avail) {
[...]

PR #968, opened 2020-08-10
Teams bug fixes

davidozog:
Please see #960 and #963 for a summary of the bug fixes and the addition of a quiet operation to `shmem_internal_choose_pync`.

### DIFF
--- a/src/shmem_internal.h
+++ b/src/shmem_internal.h
@@ -527,7 +527,7 @@ void shmem_internal_bit_set(unsigned char *ptr, size_t size, size_t index)
 {
     shmem_internal_assert(size > 0 && (index < size * CHAR_BIT));
 
-    size_t which_byte = index / size;
+    size_t which_byte = index / CHAR_BIT;
     ptr[which_byte] |= (1 << (index % CHAR_BIT));
 
     return;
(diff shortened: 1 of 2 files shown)


davidozog:
Thanks @jdinan and @akhillanger.

I think @wrrobin may have mentioned checking the `shmem_internal_bit_fetch` change (please correct me if I'm wrong).

wrrobin:
Thanks @davidozog for the explanation. It seems to be the logic is ok. I am just not sure why the bit manipulation for `fetch` is different from `set` and `clear`. May be I am missing something. For `fetch`, we dont have the `size` argument in the function definition. Otherwise, the logic could have been different, right?

davidozog:
@wrrobin Oh geez... that's a great question.  Now I'm thinking there's a problem with `which_byte` in `set` and `clear`.  Shouldn't `which_byte = index / CHAR_BIT`?  That would resolve the inconsistency you're talking about.

Currently, the `size` argument to `set` and `clear` has only been 8, so it makes no difference now, but if `size` were any different, I believe this would be a bug.  I suppose the only reason to pass `size` is to assert that the `set`/`clear`/`fetch` is within the bounds of the bit-array (given `index` and the starting `ptr`).

@akhillanger can you please confirm this?  Thanks again for catching these issues!

jdinan:
The changes to `shmem_internal_bit_fetch` look correct to me.

davidozog:
@jdinan - Do you agree that `which_byte` (in `shmem_internal_bit_set` and `shmem_internal_bit_clear`) should be set to `index / CHAR_BIT` and _not_ `index / size`?

jdinan:
If `size` is the length of the byte array, I agree this is not the right input to use when locating a particular bit in the array.

akhillanger:
> @jdinan - Do you agree that `which_byte` (in `shmem_internal_bit_set` and `shmem_internal_bit_clear`) should be set to `index / CHAR_BIT` and _not_ `index / size`?

@davidozog I agree this is the right thing to do.

davidozog:
Thanks guys.  Can I please get a re-review on the new changes?

davidozog:
@jdinan We certainly could.  I figure it doesn't hurt much, since `shmem_internal_assert` is usually `#idef`'d out anyway.  And keeping `size` encourages correct usage, but yeah we could remove it...

Issue #974, opened 2020-08-18
UCX progress deadlock


davidozog:
Seemingly random unit tests (also seemingly during back-to-back collectives/syncs) freeze during `shmem_transport_probe` within a `shmem_internal_sync`.

Here is and example of a backtrace with 2 PEs on the `alltoalls` unit test:

PE 0:
```
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
pthread_spin_lock () at ../sysdeps/x86_64/nptl/pthread_spin_lock.S:24
24      ../sysdeps/x86_64/nptl/pthread_spin_lock.S: No such file or directory.
(gdb) bt
#0  pthread_spin_lock () at ../sysdeps/x86_64/nptl/pthread_spin_lock.S:24
#1  0x00007fa0051e7cb5 in ucs_spin_lock (lock=0x7fa006348018) at /home/dave/Repos/ucx/build-v1.8.1/../src/ucs/type/spinlock.h:79
#2  0x00007fa0051ee2ac in ucp_worker_progress (worker=0x7fa006348010) at ../../../src/ucp/core/ucp_worker.c:1947
#3  0x00007fa005da1677 in shmem_transport_probe () at ../../src/transport_ucx.h:93
#4  0x00007fa005da796c in shmem_internal_sync_linear (PE_start=0, PE_stride=1, PE_size=2, pSync=0x6020a0 <pSync>) at ../../src/collectives.c:283
#5  0x00007fa005da3065 in shmem_internal_sync (PE_start=0, PE_stride=1, PE_size=2, pSync=0x6020a0 <pSync>) at ../../src/shmem_collectives.h:61
#6  0x00007fa005da3263 in shmem_internal_barrier (PE_start=0, PE_stride=1, PE_size=2, pSync=0x6020a0 <pSync>) at ../../src/shmem_collectives.h:99
#7  0x00007fa005dab4ee in shmem_internal_alltoalls (dest=0x800022e0, source=0x80002250, dst=1, sst=2, elem_size=4, nelems=7, PE_start=0, PE_stride=1, PE_size=2, pSync=0x6020a0 <pSync>) at ../../src/collectives.c:1291
[...]

PR #975, opened 2020-08-25
UCX: Flush instead of fence with shared memory

jdinan:
Addresses one of the issues on #974


jdinan:
@davidozog This is ready to review.

jdinan:
@davidozog 👋

jdinan:
No worries, sorry about the poke. :)

We need to flush the network transport if communication with a particular peer can be a mix of shared memory and network transport. If, for each peer, it's fully one or the other then a fence is sufficient.

PR #976, opened 2020-09-01
Fix C99 header issue

jdinan:
Building with `-std=c99` fails on the tests with the following error:
```
../../../test/unit/shmem_team_collect_active_set.c:108:32: error: too few arguments to function call, expected 4, have 1
            shmem_sync(new_team);
            ~~~~~~~~~~         ^
../../mpp/shmem.h:1383:1: note: 'shmem_sync' declared here
SHMEM_FUNCTION_ATTRIBUTES void shmem_sync(int PE_start, int logPE_stride, int PE_size, long *pSync);
^
../../mpp/shmem.h:36:40: note: expanded from macro 'SHMEM_FUNCTION_ATTRIBUTES'
#     define SHMEM_FUNCTION_ATTRIBUTES __attribute__((visibility("default")))
                                       ^
1 error generated.
```

There are also issues with the spec examples with `CFLAGS=-std=c99`. Not sure how to handle those.

### DIFF
--- a/mpp/shmem.h4
+++ b/mpp/shmem.h4
@@ -567,9 +567,6 @@ define(`SHMEM_CXX_TEST_SOME_VECTOR',
 }')dnl
 SHMEM_CXX_DEFINE_FOR_SYNC(`SHMEM_CXX_TEST_SOME_VECTOR')
 
-static inline int SHPRE()shmem_sync(shmem_team_t team) {
-    return shmem_team_sync(team);
-}
 
 /* C11 Generic Macros */
 #elif (defined(__STDC_VERSION__) && __STDC_VERSION__ >= 201112L && !defined(SHMEM_INTERNAL_INCLUDE))
@@ -1214,7 +1211,6 @@ define(`SHMEM_C11_GEN_TEST_SOME_VECTOR', `        $2*: shmem_$1_test_some_vector
 SHMEM_BIND_C11_SYNC(`SHMEM_C11_GEN_TEST_SOME_VECTOR', `, \') \
     )(__VA_ARGS__)
 
-/* shmem_sync */
 #define shmem_sync(...) \
     _Generic(SHMEM_C11_TYPE_EVAL_PTR_OR_SCALAR(SHMEM_C11_ARG0(__VA_ARGS__)), \
         shmem_team_t: shmem_team_sync, \
--- a/test/unit/shmem_team_collect_active_set.c
+++ b/test/unit/shmem_team_collect_active_set.c
@@ -105,7 +105,11 @@ int main(void)
             for (j = 0; j < MAX_NPES*MAX_NPES; j++)
                 dst[j] = -1;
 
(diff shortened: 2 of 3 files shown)


jdinan:
BTW, I thought we had a `-std=c99` row in Travis. Am I misremembering?

davidozog:
I only see `-std=c99` in the PRK check for some reason.

I'll look into adding it.  Checking `-std=c99` is our release checklist, and I'm sure we checked it somewhere... so I'm not sure what went wrong here.

Also, @jdinan what compiler/version are you using?  (I think we've given up on gcc 4.8.5 and earlier - perhaps we should make that an explicit requirement?)

jdinan:
Ahh, you're right. It's in the release checklist. FYI, a bunch of the spec examples will fail to build with a c99 compiler. I think I think we detect lack of c11 support in configury, so the easiest thing may be to disable spec-examples in that case.

This was a user-reported bug:
```
$ gcc --version
gcc (GCC) 9.3.0
Copyright (C) 2019 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

davidozog:
@jdinan I'm trying to understand how we missed this and it looks like it may be because libtool is throwing in `-std=gnu11` in addition to what I'm requesting - `CFLAGS=-std=c99` or `CC="gcc -std=c99" CFLAGS=` etc (I'm on gcc v9.2.1).  Do you have a method/trick to reproduce this?

jdinan:
I was able to reproduce on my Mac with clang by configuring with `CFLAGS=-std=c99`.

davidozog:
Interesting, I can also reproduce on Mac w/ clang, but just like on Linux, I see libtool is also passing both flags on the Mac: `-std=gnu11 -std=c99`.  It looks like the configure checks from #314 have made it tough to set `-std=c99` in isolation... Do we really need to add `-std=gnu11` to `CFLAGS` if C11 generics support is detected?  If so, we might need to delete it when doing the release check...

Side note: if I remove the "hard-coded" `-std=gnu11` above and try to test `-std=c99` in Linux, I get an error in pmi-simple:
```
../../pmi-simple/simple_pmi.c: In function ‘PMII_Connect_to_pm’:
../../pmi-simple/simple_pmi.c:912:45: error: ‘struct hostent’ has no member named ‘h_addr’
     memcpy( (void *)&sa.sin_addr, (void *)hp->h_addr, hp->h_length);
                                             ^~
```
and probably much more where that came from... for instance, `M_PI` is undefined in the `pi` unit test with `-std=c99`.  However, **`-std=gnu99`** works fine when building SOS (but runs into the unit test issues this PR addresses).

As far as the spec examples go, I agree, we may have to disable those tests if the configure C11 generics test fails.

davidozog (approved):

PR #982, opened 2020-11-03
Update tests for shmalloc & shmem_calloc when size is zero

davidozog:
These tests fail when `nProcs` is 1 because they try to allocate a buffer with zero size while expecting a valid pointer.  Updating the tests so they don't fail in that case.

### DIFF
--- a/test/unit/shmem_calloc.c
+++ b/test/unit/shmem_calloc.c
@@ -193,14 +193,16 @@ main(int argc, char **argv)
 
         result_sz = (nProcs-1) * (nWords * sizeof(DataType));
         result = (DataType *)shmem_calloc((nProcs-1)*nWords, sizeof(DataType));
-        if (! result)
-        {
-            perror ("Failed result memory allocation");
-            shmem_finalize();
-            exit (1);
+        if (result_sz != 0) {
+            if (! result)
+            {
+                perror ("Failed result memory allocation");
+                shmem_finalize();
+                exit (1);
+            }
+            for(dp=result; dp < &result[(result_sz/sizeof(DataType))];)
+                *dp++ += 1;
         }
-        for(dp=result; dp < &result[(result_sz/sizeof(DataType))];)
-            *dp++ += 1;
 
 
         target_sz = nWords * sizeof(DataType);
--- a/test/unit/shmem_calloc.c
+++ b/test/unit/shmem_calloc.c
@@ -191,18 +191,23 @@ main(int argc, char **argv)
 
     for(l=0; l < loops; l++) {
 
-        result_sz = (nProcs-1) * (nWords * sizeof(DataType));
-        result = (DataType *)shmem_calloc((nProcs-1)*nWords, sizeof(DataType));
-        if (result_sz != 0) {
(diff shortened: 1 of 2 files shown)


davidozog:
@wrrobin I think I understand - it might be good to check that a null pointer is returned with the requested size is 0.  One question though - are you suggesting that we exit the program here if `result_sz` is 0?  I think we want to continue with the other malloc/calloc checks below this one, because they are all valid/useful checks, right?

Another question is what happens in the for loop immediately after this block when the initialization expression (`dp=result`) is a null assignment...  I can't find any mention of this in the C standard, but I tried it in clang and gcc and seems to work fine... just skips the loop.

wrrobin:
No, We should not exit if the allocation size is 0. We should check and if the requested size was 0, then we proceed to the next tests. 
Yes, I was thinking the same. If the for loop does not skip, we have to add another terminating condition for the loop.

davidozog:
Yeah, I think we should probably avoid the for loop if `result_sz` is 0... and it would be nice to not have to make any substantial changes here.  The main goal is simply to get the test to pass when npes is 1.

How about making this change?
```
-        result_sz = (nProcs-1) * (nWords * sizeof(DataType));
+        result_sz = nProcs * (nWords * sizeof(DataType));
```
It's not as though this test is doing any communication.  I don't see any reason these tests to only allocate `nProcs-1` words...

If you want, we could then do a check for a null result when the requested size is 0 separately and it would be a lot simpler.

wrrobin:
Yes, this change is ok. 
I agree that we dont need substantial changes here.. I was just thinking how to add this check for 0 size alloc, which we have not added in any of our tests. If we allocate with nProcs, then there is no point of adding a check for size 0, since it is not expected to run with nprocs = 0. It will fail in shmem_init, I guess, right?

davidozog:
Agreed - I can make the `(nProc-1) -> nProc` change and include a separate check for 0-sized alloc in this same test and PR if that's ok with you - it should be simpler that way.

wrrobin:
Thanks @davidozog. Is there any other case where we would have a 0 sized allocation in this test? If not, then it does not make sense to add an extra check here.

davidozog:
> Thanks @davidozog. Is there any other case where we would have a 0 sized allocation in this test? If not, then it does not make sense to add an extra check here.

Maybe this next commit will clarify what I mean - pushing shortly.

wrrobin:
Yeah.. this makes sense. Thanks for clarifying 👍

test/unit/shmem_calloc.c
```diff
@@ -191,7 +191,14 @@ main(int argc, char **argv)
 
     for(l=0; l < loops; l++) {
 
-        result_sz = (nProcs-1) * (nWords * sizeof(DataType));
+        result = (DataType *)shmem_calloc(0);
```
wrrobin, line 194: `calloc` will need two arguments.

davidozog: Lol - thanks

wrrobin (approved):

PR #983, opened 2020-11-13
Added an examples directory

kholland-intel:
The examples directory has a few simple OpenSHMEM programs, a
simple Makefile, a LICENSE file, and a README with build
instructions.

Also, Getting Started Guide can be found here: https://github.com/kholland-intel/SOS/wiki/Getting-Started-Guide.


lanl-ompi:

examples/LICENSE
```diff
@@ -0,0 +1,28 @@
+Copyright (c) 2017 Intel Corporation. All rights reserved.
```
wrrobin, line 1: 2017 -> 2020

examples/LICENSE.txt
```diff
@@ -0,0 +1,28 @@
+Copyright 2011 Sandia Corporation. Under the terms of Contract
```
wrrobin, line 1: Do we need a txt file as well? @davidozog @kholland-intel

examples/README
```diff
@@ -0,0 +1,8 @@
+Build all examples using "make" or "make all".
```
wrrobin, line 1: Should we mention about adding `oshcc` and `oshrun` in the PATH? Also, adding the library path.

davidozog: Yeah, we probably should mention the bare minimum requirements to run these examples.  Something like:

> Requirements to build and run the examples:
> * Have the SOS installation "bin" directory to your PATH.
> * Have a compatible process launcher, like Hydra version 3.2, on your path.
> * The hello world example can be run with 4 processes like so:
>           oshrun -n 4 ./hello

Although it might be better to keep this `examples` directory self-contained, perhaps we could also provide a link to the new wiki page "for more detailed information":
https://github.com/Sandia-OpenSHMEM/SOS/wiki/Getting-Started-Guide
Specifically, we could directly link to the ["Building"](https://github.com/kholland-intel/SOS/wiki/Getting-Started-Guide#compiling-an-openshmem-program) and ["Running"](https://github.com/kholland-intel/SOS/wiki/Getting-Started-Guide#running-an-openshmem-program) OpenSHMEM programs sections.

I put a placeholder document on the official SOS wiki for now - feel free to update it now @kholland-intel, it looks ready to me and we can fine tune it later if needed ;)

examples/pi_reduce.c
```diff
@@ -0,0 +1,59 @@
...
+    shmem_barrier_all();
+
+    /*
+    ** Reduce each PE's inside & total value into the same symmetric address
+    */
+    shmem_sum_reduce(SHMEM_TEAM_WORLD, &inside, &inside, 1);
+    shmem_sum_reduce(SHMEM_TEAM_WORLD, &total, &total, 1);
+
+    if(me == 0) {
+
```
davidozog, line 51: Can remove this extra newline

examples/pi_reduce_array.c
```diff
@@ -0,0 +1,86 @@
...
+
+        if(x*x + y*y < 1) {
+            ++localInside;
+        }
+    }
+
+    /*
+    ** Store each PE's local inside/total values in the inside array at an index
+    ** matching the PE's unique id
+    */
```
davidozog, line 59: I think we should remove the pi "array" example to keep things simple.  This version ends up with the correct calculation, but each PE uses only 1 index of the entire array (wasting memory).  Probably best to just use the `pi_reduce.c` version, which looks perfect!

wrrobin: I agree. Had the same feeling.

wrrobin: Actually, on a second thought, I was thinking whether `pi.c` and `pi_reduce.c` could be combined to have a single program with two different options using compile flags.

davidozog: @wrrobin - Did you see the pi "exercise" in Kieran's new "Getting Started Guide"?  We might not want to give the answer away to the interested reader 😉

wrrobin: oh.. ok. I see..

kholland-intel: I just removed pi_reduce_array.c for now, if we want to combine them I can do that as well. I do believe that it may make the simple examples a little harder to follow.

examples/pi_reduce.c
kholland-intel: Done.

examples/README
kholland-intel: Done.

examples/LICENSE.txt
kholland-intel: Yeah, that was exactly it. I'll fix this mistake.

examples/LICENSE
kholland-intel: Done.

examples/README
```diff
@@ -0,0 +1,21 @@
...
+
+To build examples individually, enter "make" followed by the filename (omitting the
+file extension)
+
+To remove all the executables run "make clean"
+
+Requirments to build and run the examples:
+  * Have the SOS installation "bin" directory set in the environment PATH
+  * Have a compatible process laucher (like Hydra version 3.2) set in the
+    environemt PATH
```
davidozog, line 11: Misspelling on "environment"

examples/README
```diff
@@ -0,0 +1,21 @@
...
+  * An example of setting the environment PATH on bash:
+    $ export PATH=<path-to-SOS-install>/bin:<path-to-hydra-install>/bin:$PATH
+
+The hello world example can be run with 4 processes like so: 
+  oshrun -n 4 ./hello
+
+For more detailed information visit the Getting Started Guide:
+  https://github.com/Sandia-OpenSHMEM/SOS/wiki/Getting-Started-Guide
+
+The Licensing information can be found in the "LICENSE.txt"
```
davidozog, line 21: "LICENSE.txt" -> LICENSE file.

davidozog (changes requested):
Looks good!  Just a couple typos...

davidozog (approved):
Looks good!  Let's give @wrrobin a chance to re-review before merging.

examples/Makefile
```diff
@@ -0,0 +1,20 @@
...
+
+pi: pi.c
+	${CC} pi.c -o $@
+
+pi_reduce: pi_reduce.c
+	${CC} pi_reduce.c -o $@
+
+.PHONY: clean
+clean:
+	${RM} *.o hello pi pi_reduce
```
wrrobin, line 20: Should we use `${RM}` here? It may not have been defined by the user / system.

examples/README
```diff
@@ -0,0 +1,21 @@
+Build all examples using "make" or "make all"
+
+To build examples individually, enter "make" followed by the filename (omitting the
+file extension)
+
```
wrrobin, line 5: Minor change:
```
To build examples individually, use "make" followed by the filename (omitting the
file extension). For example, to build the hello world program, use "make hello".
```

examples/README
```diff
@@ -0,0 +1,21 @@
...
+To remove all the executables run "make clean"
+
+Requirments to build and run the examples:
+  * Have the SOS installation "bin" directory set in the environment PATH
+  * Have a compatible process laucher (like Hydra version 3.2) set in the
```
wrrobin, line 10: laucher -> launcher

examples/README
```diff
@@ -0,0 +1,21 @@
+Build all examples using "make" or "make all"
+
+To build examples individually, enter "make" followed by the filename (omitting the
+file extension)
+
+To remove all the executables run "make clean"
```
wrrobin, line 6: Use a comma after "executables"

examples/README
```diff
@@ -0,0 +1,21 @@
...
+
+Requirments to build and run the examples:
+  * Have the SOS installation "bin" directory set in the environment PATH
+  * Have a compatible process laucher (like Hydra version 3.2) set in the
+    environment PATH
+  * An example of setting the environment PATH on bash:
+    $ export PATH=<path-to-SOS-install>/bin:<path-to-hydra-install>/bin:$PATH
+
+The hello world example can be run with 4 processes like so:
```
wrrobin, line 15: like so -> as below

examples/README
```diff
@@ -0,0 +1,21 @@
...
+
+Requirments to build and run the examples:
+  * Have the SOS installation "bin" directory set in the environment PATH
+  * Have a compatible process laucher (like Hydra version 3.2) set in the
+    environment PATH
+  * An example of setting the environment PATH on bash:
+    $ export PATH=<path-to-SOS-install>/bin:<path-to-hydra-install>/bin:$PATH
+
+The hello world example can be run with 4 processes like so: 
+  oshrun -n 4 ./hello
```
wrrobin, line 16: We have asked to add `hydra` in the PATH, but then using `oshrun` here. Might it be better to use `hydra` here?

examples/hello.c
```diff
@@ -0,0 +1,28 @@
+#include <shmem.h>
+#include <stdio.h>
+#include <stdlib.h>
```
wrrobin, line 3: `stdlib` is not required

examples/hello.c
```diff
@@ -0,0 +1,28 @@
...
+    ** Fetch the number or processes
+    ** Some implementations use num_pes();
+    */
+    myshmem_n_pes = shmem_n_pes();
+    /*
+    ** Assign my process ID to me
+    */
+    me = shmem_my_pe();
+
+    printf("Hello World from %d of %d\n",me,myshmem_n_pes);
```
wrrobin, line 23: For pretty printing, use a space after each comma.

examples/pi.c
```diff
@@ -0,0 +1,61 @@
...
+
+        if(x*x + y*y < 1) {
+            ++inside;
+        }
+    }
+
+    shmem_barrier_all();
+
+    if(me == 0) {
+        for(int i = 1; i < myshmem_n_pes; ++i) {
```
wrrobin, line 45: Minor comment: Since the user might run this program with older gcc, it might be better to declare `i` at the top before.

examples/pi_reduce.c
```diff
@@ -0,0 +1,58 @@
+#include <shmem.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <assert.h>
+#include <time.h>
+#include <math.h>
```
wrrobin, line 6: Same comment as before. Please check whether all the header files are used.

examples/pi_reduce.c
```diff
@@ -0,0 +1,58 @@
...
+
+    /*
+    ** Reduce each PE's inside & total value into the same symmetric address
+    */
+    shmem_sum_reduce(SHMEM_TEAM_WORLD, &inside, &inside, 1);
+    shmem_sum_reduce(SHMEM_TEAM_WORLD, &total, &total, 1);
+
+    if(me == 0) {
+        double approx_pi = 4.0*inside/(double)total;
+        printf("Pi from %llu points on %d PEs: %lf\n",total,myshmem_n_pes,approx_pi);
```
wrrobin, line 52: Same comment as before.

examples/Makefile
wrrobin: Yes, I also see it everywhere. We can keep it as it is I think. Got a useful read: https://www.gnu.org/software/make/manual/html_node/Implicit-Variables.html

wrrobin:
LGTM. I have left some minor comments.

examples/README
kholland-intel: Since the Getting Started Guide encourages the use of `oshrun` to run the programs should I instead just make a note that `oshrun` invokes `hydra` and that is why it is needed in the `PATH`?

davidozog: Seems simpler to me to encourage using `oshrun` for these examples and referring to the getting started guide for more details about the launcher.  

The guide should be clear about what happens by invoking `oshrun` - i.e., it picks up the following launchers in this order of preference: yod, mpirun, mpiexec, and srun.  (But we only test/guarantee compatibility with Hydra.)  Also the user can set the `OSHRUN_LAUNCHER` env var for a custom launcher.

wrrobin: ok, sounds good

examples/pi_reduce.c
davidozog: Looks like this came from `hello.c` and `pi.c`.  Do we want to change those there as well @wrrobin?  (I'd say might as well while we're at it).

wrrobin: Yes, it is a good idea to make these same now.

wrrobin (approved):

Issue #977, opened 2020-09-02
Some OpenSHMEM specification examples are C11 only


davidozog:
These tests should be disabled if the compiler only supports C99.

This could be detected in `configure.ac` (the `shmem_cv_c11_works` check) and passed along to `test/spec-examples/Makefile.am` and disabled, if necessary.

Also related to issue #932.

davidozog:
Closed by #986

PR #986, opened 2020-12-11
Update spec-examples Makefile.am for v1.5

davidozog:
It looks like we may have missed this during code review:

* The spec-examples are not shmemx in v1.5, so we can remove the `SHMEMX_TESTS` guard in the Makefile 
* `shmem_team_sync` and `shmem_reduce_example` require c11, so we can guard those w/ `if HAVE_C11`

This should address #977

### DIFF
--- a/test/spec-example/Makefile.am
+++ b/test/spec-example/Makefile.am
@@ -12,15 +12,7 @@
 # distribution.
 
 check_PROGRAMS = \
-        shmem_ctx_pipelined_reduce
-
-if HAVE_OPENMP
-check_PROGRAMS += \
-    shmem_ctx
-endif
-
-if SHMEMX_TESTS
-check_PROGRAMS += \
+        shmem_ctx_pipelined_reduce \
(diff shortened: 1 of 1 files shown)


davidozog:
@jdinan, any chance you can review this?  I think you proposed a slightly different fix (in PR #976), but this approach might be a tad bit easier and will still run the c99-compatible spec examples when c11 isn't available (that, and Wasi's unavailable for a little while ;).

wrrobin:
@davidozog Thanks for catching this and it looks good to me. Seems to be a better way to handle the issue #976 for now.

wrrobin (approved):

davidozog:
Thanks @wrrobin - do you think we should merge for inclusion with v1.5.0?  I've only tested on OSX+clang and Ubuntu20.04+gcc with `-std=gnu99`.

wrrobin:
@davidozog I am ok with merging this. We can run some tests on the Jenkins and see effects.

PR #987, opened 2020-12-14
(v1.4.x) cherry-pick features/fixes ahead of v1.4.5

davidozog:
I went ahead and created a `v1.4.x` branch on SOS that is synced with the `v1.4.5` tag.

This PR cherry-picks several commits from the master branch to include in `v1.4.x` to fix some bugs and have a useful Travis script.  Here's a short list of what's included:

* 8de8a785075b3cf5498c8753d98814506ce5e396 to allow higher thread levels in the tests and adds to `v1.4.x`. 
* All commits from PR #968 (teams bug fixes)
* ea3ea06749af57fb12fb48ddac23eaf6ca01244f to fix XPMEM testing in Travis
* A manual merge of the updates from PR #966  (primarily to update PMIx support in Travis)

I'll also add the shmem_perf_suite trial multiplier in PR #988 (and sync with https://github.com/openshmem-org/tests-sos) to `v1.4.x` after it's approved/merged into master.

All should feel free to request other features/bug fixes for `v1.4.x`.

Issue #17, opened 2015-10-21
CMA and Portals


jdinan:
CMA currently makes a Portals call in shmem_internal_put_nb, causing a build failure when CMA is used with the OFI transport.  Is calling the transport the right solution or should CMA also work in a transport-less shared memory build?

jdinan:
Presently, Sandia SHMEM doesn't support a build without a network transport.  The current design requires that some functionality always be provided through the transport, such as the atomics.

PR #988, opened 2020-12-15
shmem_perf_suite: add multiplier for trials value

davidozog:
This feature is courtesy of @minsii.  PR #17 in tests-sos has more info:
https://github.com/openshmem-org/tests-sos/pull/17

After reviewing and merging this into master, I'll then include it in the v1.4.x branch as well (in PR #987).

### DIFF
--- a/test/performance/shmem_perf_suite/common.h
+++ b/test/performance/shmem_perf_suite/common.h
@@ -145,6 +145,10 @@ typedef struct perf_metrics {
     char *src, *dest;
     op_style opstyle;
 
+    float trials_multiplier; /* adjust trials value through env var
+                              * SOS_TRIALS_MULTIPLIER. Invalid for
+                              * user inputs*/
+
     /* parameters for threaded tests */
     int nthreads;
     int thread_safety;
@@ -172,10 +176,16 @@ long bar_psync[SHMEM_BARRIER_SYNC_SIZE];
 /* default settings with no input provided */
 static inline 
 void set_metric_defaults(perf_metrics_t *metric_info) {
+    char *val = NULL;
+    metric_info->trials_multiplier = 1.0; /* Default 1 */
+    val = getenv("SOS_TRIALS_MULTIPLIER");
+    if (val && strlen(val))
+        metric_info->trials_multiplier = atof(val);
+
     metric_info->start_len = START_LEN;
     metric_info->max_len = MAX_MSG_SIZE;
     metric_info->size_inc = INC;
-    metric_info->trials = TRIALS;
+    metric_info->trials = TRIALS * metric_info->trials_multiplier;
     metric_info->window_size = WINDOW_SIZE; /*back-to-back msg stream*/
     metric_info->warmup = WARMUP; /*number of initial iterations to skip*/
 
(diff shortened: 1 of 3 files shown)


test/performance/shmem_perf_suite/common.h
```diff
@@ -172,10 +176,16 @@ long bar_psync[SHMEM_BARRIER_SYNC_SIZE];
 /* default settings with no input provided */
 static inline 
 void set_metric_defaults(perf_metrics_t *metric_info) {
+    char *val = NULL;
+    metric_info->trials_multiplier = 1.0; /* Default 1 */
+    val = getenv("SOS_TRIALS_MULTIPLIER");
```
wrrobin, line 181: Should we use `SHMEM_PERF_SUITE_TRIALS_MULTIPLIER`? Actually, I will prefer adding this as an argument flag rather than an env variable? Thoughts?

davidozog: Yeah, that would definitely be more consistent with the existing tool...

Would that work on your end @minsii?  While the env var is slightly easier b/c you just set it once, we really have no precedent for such an env var in shmem_perf_suite, and adding a new flag may be generally more useful.

minsii:
@davidozog How do we use argument flag with `make check` ? As long as I can set it with `make check` command, it is fine to me.

wrrobin:
@minsii We have a `-n` and `-p` flag to change the number of trials and warm-up iterations for small message sizes. For larger message sizes (>=64K), we keep the default (10 warmup, 100 trials). Let us know if you think any other behavior would make sense to you.
If `make check` is the only intended way to run this, then I am not sure we can add these flags in the command. We can merge the current change with the env var in that case to add this extra knob.

minsii:
@wrrobin Thanks for the explanation. Our test framework simply runs `make check` to trigger all tests.

davidozog:
@minsii - Do you configure SOS with `--enable-lengthy-tests`?   Otherwise, I don't believe the shmem_perf_suite tests would run (they'd just build).

Also, are you ok with @wrrobin's proposed name change to `SHMEM_PERF_SUITE_TRIALS_MULTIPLIER`?  This test suite is technically disjoint from "SOS", so it might be for the best.

Given the "make check" usage, it seems like we should keep the env var.  @wrrobin do you think we should add the argument flag in addition to that?  It would be easy to add... but maybe not as useful - I imagine most users would simply change "trials" if they're running individual performance tests.

wrrobin:
@davidozog Yes, adding it to the argument list makes sense as most users might run individual tests. If @minsii agrees with the var name, I will make those changes and we can merge that in.

minsii:
The env var renaming sounds good to me. Thanks!

davidozog:
> @davidozog Yes, adding it to the argument list makes sense as most users might run individual tests. If @minsii agrees with the var name, I will make those changes and we can merge that in.

I think I already have this ready @wrrobin - hopefully you haven't started ;). Just testing now and I'll push for a review in a few minutes.

wrrobin:
@davidozog Ok. Thanks. Will wait for your changes.

davidozog:
@wrrobin the latest commit is ready to review - thanks!

wrrobin (approved):
LGTM. Thanks.

davidozog:
Great - I'll merge after it gets through testing, but I don't expect any surprises.

PR #991, opened 2021-01-14
Updated SOS build-system file for OFI

kholland-intel:
Updated opal_check_ofi.m4 to the latest upstream version from Open MPI.
Commented out block of code which required pmix for OFI.


lanl-ompi:

configure.ac
```diff
@@ -344,7 +344,9 @@ OMPI_CHECK_PORTALS4([portals4],
     [transport_portals4="yes"],
     [transport_portals4="no"])
 
-OPAL_CHECK_OFI([ofi],
+OPAL_CHECK_OFI
+
+AS_IF([test "$opal_ofi_happy" = "yes"],
     [transport_ofi="yes"],
     [transport_ofi="no"])
```
wrrobin, line 352: @kholland-intel Thanks for the changes.. Just checking, did commenting out the PMIx check resolve the error on Travis and locally?

kholland-intel: Yes, after commenting out the code block which required PMIx the error was resolved both locally and on Travis. The code was commented out on lines 136-140 of opal_check_ofi.m4:

wrrobin (approved):
LGTM

davidozog (approved):
Looks good.

One concern might be that some of the `OPAL_*` m4 interfaces used here are meant to be used with the latest upstream versions as well... (like `OPAL_CHECK_WITHDIR`, `OPAL_SUMMARY_ADD`, etc, for example)  but I think we can probably merge this one now and include it in a future PR which updates the other OPAL scripts.

kholland-intel:
@davidozog @wrrobin 
I double-checked each of the files from this PR with the latest upstream versions from ompi and most of them are still up-to-date with ompi's [master branch](https://github.com/open-mpi/ompi/tree/master). The only file that is not aligned with ompi's master branch is opal_check_pmi.m4. This is because ompi has removed this file and replaced it with two new files that now make PMIx a requirement (which, from what I understand, we are not ready to do). However, The version of opal_check_pmi.m4 that is included in this PR is inline with the lastest ompi release build: [v1.4.0](https://github.com/open-mpi/ompi/tree/v4.1.0)

With that being said, is there anything left to check before merging this PR?

configure.ac
```diff
@@ -457,7 +459,7 @@ AC_ARG_ENABLE([pmi-mpi], [AC_HELP_STRING([--enable-pmi-mpi],
 
 AS_IF([test "$enable_pmi_simple" = "yes" -o "$enable_pmi_mpi" = "yes"],
      [TEST_RUNNER='$(abs_top_builddir)/src/oshrun -n $(NPROCS)'],
-     [OPAL_CHECK_PMI(pmi_type)])
+     [OPAL_CHECK_PMIX(pmi_type)])
```
wrrobin, line 462: Can this check return `pmi_type` as pmi1 or pmi2? If not, the checks on line 477 and 478 may not make sense any more. But, for now, seems to be ok.

kholland-intel:
Since ompi moved in the direction of making PMIx a requirement in their opal_check_pmi.m4 file, I have reverted the changes for that file back to [v1.5.0](https://github.com/Sandia-OpenSHMEM/SOS/tree/v1.5.0) of Sandia OpenSHMEM. I have also aligned the rest of the build files in the config area with the latest ompi release build: [v4.1.0](https://github.com/open-mpi/ompi/tree/v4.1.0).

davidozog (dismissed):
I've tested and reviewed this as best I can... I think it's as good a time as any to go ahead and merge.

@kholland-intel - I'm thinking you should choose the "Rebase and merge" option instead of a standard merge.  That will group all the configury commits together at the tip of master, which could potentially make it easier to track down any issues in the future.  Does that make sense?

Any other concerns @wrrobin?

davidozog:
Also, there appears to be a problem with the UCX build which seems to be unrelated to this PR...  I'm still investigating, but please hold off on merging until that's certain.  We did update the UCX configury here...

```
../../src/transport_ucx.h: In function ‘shmem_transport_put_nb’:
4694../../src/transport_ucx.h:261:5: error: unknown type name ‘ucp_request_param_t’
4695     ucp_request_param_t param = {
```

wrrobin:
@davidozog I dont have any other concerns. If we can resolve the UCX configury update, I am also ok to merge this.

davidozog:
@kholland-intel - could you please update this PR with the latest master so we can see if there are any issues?  I'm pretty sure the problem with UCX is simply due to this PR being a little behind our latest Travis script (which builds UCX v1.9.0 instead of v1.8.1)

davidozog (approved):

PR #994, opened 2021-02-09
Fix small leak on kvs_value in runtime-pmi.c

davidozog:
Valgrind says there's a consistent 1KB leak here, but only when NPES == 1.

The problem appears to be that the address of `kvs_value` was being set, so it becomes an invalid free during finalization.  Setting the value instead seems to be the original intent.

### DIFF
--- a/src/runtime-pmi.c
+++ b/src/runtime-pmi.c
@@ -123,6 +123,14 @@ shmem_runtime_fini(void)
         free(location_array);
     }
 
+    if (kvs_key) {
+        free(kvs_key);
+    }
+
+    if (kvs_value) {
+        free(kvs_value);
+    }
+
     if (initialized_pmi) {
         PMI_Finalize();
         initialized_pmi = 0;
@@ -266,7 +274,7 @@ shmem_runtime_get(int pe, char *key, void *value, size_t valuelen)
         HASH_FIND_STR(singleton_kvs, kvs_key, e);
         if (e == NULL)
             return 3;
-        kvs_value = e->val;
+        *kvs_value = *(e->val);
     }
     else {
         if (PMI_SUCCESS != PMI_KVS_Get(kvs_name, kvs_key, kvs_value, max_val_len)) {
--- a/src/runtime-pmi.c
+++ b/src/runtime-pmi.c
@@ -119,17 +119,10 @@ shmem_runtime_init(int enable_node_ranks)
 int
 shmem_runtime_fini(void)
 {
-    if (location_array) {
-        free(location_array);
-    }
-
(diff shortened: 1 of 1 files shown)


src/runtime-pmi.c
```diff
@@ -266,7 +274,7 @@ shmem_runtime_get(int pe, char *key, void *value, size_t valuelen)
         HASH_FIND_STR(singleton_kvs, kvs_key, e);
         if (e == NULL)
             return 3;
-        kvs_value = e->val;
+        *kvs_value = *(e->val);
```
jdinan, line 277: I think what you really want here is:
```
strncpy(kvs_value, e->val, max_val_len);
```

jdinan: Otherwise, you are just copying a single character. :)

davidozog: Thanks @jdinan!

davidozog:
Any chance you can review this one @jdinan?  @wrrobin and @kholland-intel are without power in Austin, TX...

src/runtime-pmi.c
```diff
@@ -123,6 +123,14 @@ shmem_runtime_fini(void)
         free(location_array);
     }
 
+    if (kvs_key) {
```
jdinan, line 126: This one is also leaked:
```
if (kvs_name) free(kvs_name);
```

jdinan (approved):
Looks ok to me. One additional allocation you can free.

src/runtime-pmi.c
davidozog: Yes indeed, thanks.

I suppose we don't need the non-null checks on these pointers either, right?  I almost forgot `free` does no action on a NULL pointer.

jdinan: That's a good point. As far as I can tell, this behavior from `free` is portable.

jdinan (approved):

jdinan:
I don't recall having seen those particular Cray tests fail before. I wonder if something changed with XPMEM that could be causing a memory consistency issue?

In any case, the changes on this PR are certainly not the cause..

davidozog:
> I don't recall having seen those particular Cray tests fail before. I wonder if something changed with XPMEM that could be causing a memory consistency issue?
> 
> In any case, the changes on this PR are certainly not the cause..

Hmm... relatively lately (guessing since mid-2020) I've possibly noticed a few more failures in Cray tests other than `finc_cswap`.  But they aren't persistent, as usual.  We'll start tracking whether they are related to XPMEM , but my hunch is that they are pretty random across all Travis rows.

PR #1008, opened 2021-07-29
Modified team based broadcast to update root dest

davidozog:
Looks like we missed this v1.5 requirement for broadcast: "The dest object is updated on all PEs."


davidozog:
I'd say we should reduce the Travis footprint by merging #1009 before testing this again.

wrrobin (approved):

PR #1009, opened 2021-07-30
Disable all external Travis CI tests for now

davidozog:
* Remove some Portals4 testing rows
* Add a row to enable deprecated API tests

For those unaware, Travis CI has changed it's billing policies so we can no longer rely on it for a wide variety of long-running tests.  I applied for SOS to receive free Open-Source-Software (OSS) credits, but currently we only have ~25k credits (core-minutes).  That's _total_ credits... like forever.  That would give us only ~2 full SOS builds.  I asked if we can get recurring plan, but responses from Travis support are very slow.

For now I propose we slim down Travis as much as we can by disabling all external tests and removing some portals4 rows.  Hopefully this buys us a little time to figure out a more thorough CI strategy (likely using a Jenkins backend).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,8 @@ env:
         - SOS_PM=mpiexec.hydra
         - SOS_PM_PRE=true
         - SOS_PM_POST=true
+        ## Disable all external tests until SOS has recurring Travis credits for OSS
+        - SOS_DISABLE_EXTERNAL_TESTS=1
         ## An encrypted value for INTEL_SERIAL_NUMBER:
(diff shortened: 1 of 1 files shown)


wrrobin:
I wonder if we should stop fetching all these since we are no longer running them 
```
    ## Fetch UH Tests
    - cd $TRAVIS_SRC
    - git clone --depth 10 https://github.com/openshmem-org/tests-uh.git tests-uh
    # Fetch Cray Tests
    - cd $TRAVIS_SRC
    - git clone --depth 10 https://github.com/openshmem-org/tests-cray.git tests-cray
    # Fetch Mellanox Tests
    - cd $TRAVIS_SRC
    - git clone --depth 10 https://github.com/openshmem-org/tests-mellanox.git tests-mellanox
    ## Fetch ISx
    - cd $TRAVIS_SRC
    - git clone --depth 10 https://github.com/ParRes/ISx.git ISx
    ## Fetch PRK
    - cd $TRAVIS_SRC
    - git clone --depth 10 https://github.com/ParRes/Kernels.git PRK
    - echo -e "SHMEMCC=oshcc -std=c99\nSHMEMTOP=$$TRAVIS_INSTALL\n" > PRK/common/make.defs
   ```

**Comment by davidozog** [comment]
> I wonder if we should stop fetching all these since we are no longer running them

Thanks @wrrobin.  I removed those git clones when external tests are disabled, and it's ready for review.

File: .travis.yml [hunk 1]
```diff
@@ -280,22 +259,21 @@ before_install:
...
-    - git clone --depth 10 https://github.com/openshmem-org/tests-mellanox.git tests-mellanox
-    ## Fetch ISx
-    - cd $TRAVIS_SRC
-    - git clone --depth 10 https://github.com/ParRes/ISx.git ISx
-    ## Fetch PRK
-    - cd $TRAVIS_SRC
-    - git clone --depth 10 https://github.com/ParRes/Kernels.git PRK
-    - echo -e "SHMEMCC=oshcc -std=c99\nSHMEMTOP=$$TRAVIS_INSTALL\n" > PRK/common/make.defs
+    - >
+      if [ -z "$SOS_DISABLE_EXTERNAL_TESTS" ]; then
```
wrrobin, line 263: Should we use `-n`? We did not assign 0.

wrrobin: I am confused here.. the Travis rows seem to be not downloading these anymore.. but our condition says if the string length is 0. It should say if it is non-zero (-n), right? :(

davidozog: I went with this approach because that's how we handle disabling tests elsewhere (e.g. [here](https://github.com/Sandia-OpenSHMEM/SOS/blob/v1.5.0/.travis.yml#L404), [here](https://github.com/Sandia-OpenSHMEM/SOS/blob/v1.5.0/.travis.yml#L413), and [here](https://github.com/Sandia-OpenSHMEM/SOS/blob/v1.5.0/.travis.yml#L392)).

Travis is not downloading these repos anymore because I set `SOS_DISABLE_EXTERNAL_TESTS=1` globally on line 25 of this PR.  So because $SOS_DISABLE_EXTERNAL_TESTS is nonzero, we don't clone the repos.  If we remove line 25, then the script will clone all those repos.

davidozog: Maybe it helps to keep in mind these are bash _string_ comparisons.. So it's not actually checking whether the value is 0 or non-zero, it's checking there the string is empty or non-empty (I think...).

wrrobin: Ok.. thanks for pointing me to line 25. That's what I missed when I was thinking that we did not assign anything.. Good to go then

wrrobin (approved):
Please merge it in. Looks good.

PR #1018, opened 2021-10-05
Fix shmem_team_get_config function signature & add a unit test

### LINKED ISSUE #992: shmem_team_get_config - wrong signature, not fully implemented
`shmem_team_get_config` should return an `int` but is currently returning `void`.  It's also not passing the `config_mask`.  And it isn't covered in the unit tests.

davidozog:
This should fix #992, but there is some uncertainty regarding how this interface behaves when `config_mask` is 0 and/or `config` is a null pointer (https://github.com/openshmem-org/specification/issues/447).

For now, I'm assuming `config` is unmodified if `config_mask` is 0. If `config_mask` is nonzero and `config` is a null pointer, the user made a mistake and the behavior is undefined.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,16 +10,136 @@ env:
   SOS_INSTALL_DIR: ${{ github.workspace }}/install/sos
   LIBFABRIC_INSTALL_DIR: ${{ github.workspace }}/install/libfabric
   UCX_INSTALL_DIR: ${{ github.workspace }}/install/ucx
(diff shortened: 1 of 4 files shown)


davidozog:
Thanks for the suggestions @dalcinl and @wrrobin - my changes are ready for review.

dalcinl:
@davidozog After a closer look, I'm not OK with this implementation. You should not make `config` a pointer in the team structure, keep things as they where. In the split routines, you should not stash the pointer the user passes, but rather copy from it to team->config, as it was done before your changes. Otherwise, the user is responsible of keeping the `shmem_config_t` memory alive as long as the team.

PR #1027, opened 2021-10-27
Pulled in v4.1.1 ompi config changes

kholland-intel:


config/opal_check_pmi.m4
```diff
@@ -11,23 +11,19 @@
 # Copyright (c) 2004-2005 The Regents of the University of California.
 #                         All rights reserved.
 # Copyright (c) 2009-2015 Cisco Systems, Inc.  All rights reserved.
-# Copyright (c) 2011-2016 Los Alamos National Security, LLC. All rights
+# Copyright (c) 2011-2014 Los Alamos National Security, LLC. All rights
```
wrrobin, line 14: Are we going to a previous version? Sorry if I missed this discussion.

wrrobin: My bad.. just saw the rest of the content...

davidozog: It is kinda funny... maybe a small mistake?

wrrobin:
It looks good to me. 

The failed tests for PMIx may be related to the updated repo that we are using in CI. Earlier, in Travis, we were checking out specific commit for PMIx and PRRTE, that was validated manually. May be, we need to do these steps manually first to see if they work. Otherwise, we may have to stick to the appropriate versions that we were using in Travis. @davidozog Thoughts?

davidozog:
@wrrobin - I wouldn't guess the failed PMIx tests are due to the new github workflow, because the tests have passed several commits/PRs up to this point.

Also something to keep in mind - I found that the Travis script was never actually checking out particular commits of PMIx.  Here for example:
https://github.com/Sandia-OpenSHMEM/SOS/blob/master/.travis.yml#L292

we didn't change into the pmix directory after cloning, so it was always on the master branch (you can see that error in the logs).  The same was true for prrte.

davidozog (approved):
Looks good, thanks.

Just a quick note: it looks like OMPI v4.1.x has a couple new checks in the UCX config file: `ucp_ep_flush_nbx` and `UCP_OP_ATTR_FLAG_MULTI_SEND`.  I don't think we need those though, so this is good to go.

Issue #1015, opened 2021-09-29
Missing shmem_signal_wait_until


dalcinl:
Looks like `shmem_signal_wait_until` is missing, despite it is documented in the manpages.

```
$ grep -r shmem_signal_wait_until /home/devel/shmem/sandia
/home/devel/shmem/sandia/share/man/man3/shmem_signal_wait_until.3:shmem_signal_wait_until \- 
```

dalcinl:
@davidozog My guess is that this should be easy to implement for you. Could give it a shot?

davidozog:
Oh, interesting.  Thanks for catching that - sure thing.

wrrobin:
Can we close this now? @dalcinl @davidozog

davidozog:
Yes, thanks - closed by #1028

PR #1028, opened 2021-11-08
shmem_signal_wait_until

davidozog:
This implements `shmem_signal_wait_until` and adds a simple unit test.

`shmem_signal_wait_until` differs from the other `wait_until` operations in that it returns the contents of the signal data object that satisfied the wait condition.  I plumbed that through `shmem_synchronization.h` without affecting the other `wait_until` routines, but there may be a cleaner way to do it.

I also included a fix to the GitHub workflow where "sudo make install" was done in user-space and resulted in some funny permission issues on my fork's CI.  @kholland-intel - there is a chance this could be related to the problem on #1027.  It might be worth trying a cherry-pick of https://github.com/Sandia-OpenSHMEM/SOS/commit/12d3e4b2fe8e668879177827d92ebfe68d955127.

### DIFF
--- a/src/shmem_synchronization.h
+++ b/src/shmem_synchronization.h
@@ -86,7 +86,7 @@ shmem_internal_fence(shmem_ctx_t ctx)
     } while(0)
 
 #define COMP_SIGNAL(type, a, b, ret) ({                  \
-    uint64_t satisfied_val;                              \
+    uint64_t satisfied_val = 0;                          \
     COMP(type, a, b, ret);                               \
     if (ret) satisfied_val = a;                          \
     satisfied_val;                                       \
--- a/src/shmem_synchronization.h
+++ b/src/shmem_synchronization.h
@@ -85,12 +85,11 @@ shmem_internal_fence(shmem_ctx_t ctx)
         }                                                \
     } while(0)
 
-#define COMP_SIGNAL(type, a, b, ret) ({                  \
-    uint64_t satisfied_val = 0;                          \
-    COMP(type, a, b, ret);                               \
-    if (ret) satisfied_val = a;                          \
-    satisfied_val;                                       \
-})
+#define COMP_SIGNAL(type, a, b, ret, sat_value)          \
+    do {                                                 \
+        COMP(type, a, b, ret);                           \
+        if (ret) sat_value = a;                          \
+    } while(0)
 
 #ifdef USE_SHR_ATOMICS
 #define SYNC_LOAD(var) __atomic_load_n(var, __ATOMIC_ACQUIRE)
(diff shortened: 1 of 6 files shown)


davidozog:
@dalcinl  This should fix #1015 - does this pass testing on your end?  Also please feel free to review if you're available to.

dalcinl:
@davidozog Test in my end are passing.  However, I warn you that my tests for this API are not serious at all. I'll add better ones in the future (it may take two months), and I'll let you know if I have any issues.

test/unit/signal_wait_until.c
```diff
@@ -0,0 +1,105 @@
...
+    for (i = 0; i < MSG_SZ; i++)
+        target[i] = 0;
+
+    shmem_barrier_all();
+
+    for (i = 0; i < npes; i++) {
+        shmem_long_put_signal(target, source, MSG_SZ, &sig_addr, 1, SHMEM_SIGNAL_SET, i);
+    }
+
+    shmem_signal_wait_until(&sig_addr, SHMEM_CMP_EQ, 1);
```
wrrobin, line 88: In equality comparison, its not required to check the return value. In other comparisons, we might have to check the return value. Also, do we have to check the value in such cases with the return value from `shmem_signal_fetch`?

davidozog: Ooo, good point.

I think we should check the return value in this test to make sure it's what we expect (1 in this case, `npes` above).  Does that sound good?

I think I understand what you mean about "other comparison", but I'd say the _application_ determines whether it's required to check the return value.  Is that what you mean?  I think the same probably applies to `shmem_signal_fetch`.

I'm not sure we have great testing coverage over the different comparisons for all the varieties of `shmem_wait` and `shmem_test`.  I don't see any coverage for `SHMEM_CMP_LE` and `SHMEM_CMP_LT` for example, but it makes sense that they wouldn't naturally be used very often.  Maybe I could create an issue for that?

wrrobin: Yes, makes sense. 

I meant to say, for equality case, even if we dont check the return value, we know what `sig_addr` has now as we already just did the equality test. But, for any other comparison, the application might have to check for the return value.. so, we may want to add one such test. I am not really sure when would be that case providing we can only do `SIGNAL_ADD` and `SIGNAL_SET` .. and we will always know what the `sig_addr` is supposed to have at any point. Not quite sure why we added the return value here... 

It is also getting confusing that different `wait_*` routines return different value semantics. Should we unify these? Thoughts?

davidozog: Ah, that makes sense - I'll see if I can change the test a bit to cover the non-equality comparisons as well.

davidozog:
Oops, when building this with `--enable-picky` I noticed that macros returning a value (using "Statement expressions") is actually a GNU extension...

So it would be far more portable to take the statement expressions out of this PR.  I'll work that out unless you oppose @wrrobin @kholland-intel?

davidozog:
@wrrobin and @kholland-intel - the recent changes are ready for review.

kholland-intel (approved):
Hey Dave, this PR Looks good to me. Thanks!

src/shmem_synchronization.h
```diff
@@ -140,16 +158,40 @@ shmem_internal_fence(shmem_ctx_t ctx)
...
+    do {                                                                \
+        uint64_t target_cntr;                                           \
+        int cmpret;                                                     \
+                                                                        \
+        COMP_SIGNAL(cond, SYNC_LOAD(var), value, cmpret, sat_value);    \
+        while (!cmpret) {                                               \
+            target_cntr = shmem_transport_received_cntr_get();          \
+            COMPILER_FENCE();                                           \
+            COMP(cond, SYNC_LOAD(var), value, cmpret);                  \
+            if (cmpret) break;                                          \
```
wrrobin, line 171: If we break from here, does the `sat_value` get updated with the correct value? We are updating it only in `COMP_SIGNAL`.. Let me know if I am not seeing anything obvious.

test/unit/signal_wait_until.c
wrrobin: Thanks for the changes. The test looks good to me now.

src/shmem_synchronization.h
davidozog: Yikes, this is definitely a mistake - it should be `COMP_SIGNAL`.  Thanks very much @wrrobin!

I'll merge after you approve this fix and testing finishes.

wrrobin (approved):

PR #1038, opened 2021-11-30
Fix warning comparing ints in signal_wait test

davidozog:
This resolves several compiler warnings when comparing ints in the following unit test:
```
../../../test/unit/signal_wait_until.c:68:16: warning: comparison of integers of different signs: 'uint64_t' (aka 'unsigned long long') and 'int' [-Wsign-compare]
    if (retval > npes)
...
```

### DIFF
--- a/test/unit/signal_wait_until.c
+++ b/test/unit/signal_wait_until.c
@@ -66,15 +66,15 @@ int main(int argc, char *argv[])
 
     uint64_t retval = shmem_signal_wait_until(&sig_addr, SHMEM_CMP_LE, npes);
 
-    if (retval > (uint64_t)npes)
+    if (retval > (uint64_t) npes)
         errors++;
 
     retval = shmem_signal_wait_until(&sig_addr, SHMEM_CMP_EQ, npes);
-    if (retval != (uint64_t)npes)
+    if (retval != (uint64_t) npes)
         errors++;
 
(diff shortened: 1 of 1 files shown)


test/unit/signal_wait_until.c
```diff
@@ -65,15 +65,16 @@ int main(int argc, char *argv[])
     }
 
     uint64_t retval = shmem_signal_wait_until(&sig_addr, SHMEM_CMP_LE, npes);
-    if (retval > npes)
+
+    if (retval > (uint64_t)npes)
```
wrrobin, line 69: Nitpick.. Do we use a space between `(type)` and `operand` usually?

wrrobin (approved):

test/unit/signal_wait_until.c
davidozog: Huh, I see we're inconsistent with the type-cast space throughout SOS (e.g., [here](https://github.com/Sandia-OpenSHMEM/SOS/blob/v1.5.0/src/data_c.c4#L439) vs. [here](https://github.com/Sandia-OpenSHMEM/SOS/blob/v1.5.0/src/transport_ofi.h#L1201)), but my hunch is we usually include a space.

I'll add the space here and try to be consistent about this style in the future.

PR #1041, opened 2021-12-17
Pshmem symbol check script

wrrobin:


scripts/pshmem_symbol_check.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/sh
+
+# Simple script to check for missing profiling symbols in 
+# SHMEM library. The readelf tool requires the library to 
+# be a shared object.
+#
+# Example usage: pshmem_symbol_check.sh <path to SHMEM library>
+#
+# Exit if any command fails
+set -e
```
davidozog, line 10: Is `set -e` needed?  If I run this tool against a library that did not include `--enable-profiling` (accidentally), then I get a silent error after the first `grep`, and it's not clear what went wrong.

davidozog (commented):
First, sorry for the delayed review.  This is a helpful script - thanks @wrrobin.

As you've probably seen this exposes `shmem_internal_*`, `shmem_transport_*`, and various Fortran routines as not having pshmem symbols, which is expected.  I think that's ok for now, but maybe we should create an issue to track it?  It would be convenient to not have to filter through the ~180 false positives to see if a pshmem symbol is missing.

scripts/pshmem_symbol_check.sh
wrrobin: Thanks for catching that. Will add something for that. 

I added `-e` thinking this is helpful, but definitely not a requirement.

wrrobin: @davidozog Let me know if the recent changes look ok to you. Also, do you prefer `set -e` to be removed? I have not yet, but am ok to remove if there is any preference.

davidozog: @wrrobin - Yes, that change is good - thanks!

I'd say it's ok to can keep `set -e` (at least for now).

davidozog (approved):

wrrobin:
@davidozog Thanks for the comment, sounds good to me. I was actually thinking that having these checked and being reported is helpful to see these do not have pshmem symbols (and which should be the case). We can add further checks to ensure we skip these.. but not sure if we can generalize these skip checks for any library.

PR #1044, opened 2022-01-31
Added a SOS Dockerfile script and a README

kholland-intel:


davidozog (commented):
Looks very useful @kholland-intel - thanks!

I have just a few suggestions/questions.

scripts/docker/README.md
```diff
@@ -0,0 +1,76 @@
+# Sandia-OpenSHMEM Docker Image
+Sandia-OpenSHMEM (SOS) utlizes Docker images to provide users a means to create
+containers that provide a configurable Ubuntu sandbox for running Sadia
```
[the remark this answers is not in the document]
wrrobin: We can also just use `SOS` since we already introduced SOS in the same sentence.

scripts/docker/README.md
```diff
@@ -0,0 +1,76 @@
...
+images.
+
+## Creating SOS Containers from the Dockerfile 
+This Dockerfile provides a configurable Ubuntu sandbox for running Sadia
+OpenSHMEM (SOS) applications and benchmarks.
+
+### Building the Image from the Dockerfile
+To build the image from the Dockerfile, run the following run the following
+command in the directory that contains the Dockerfile:
+
```
wrrobin, line 17: Minor edit comment: same words have been used with different capitalizations. For example, `Image` and `image`, `Dockerfile` and `dockerhub`.. I am not sure which one to use. just mentioning it if we need to correct / unify anything here.

scripts/docker/Dockerfile
```diff
@@ -0,0 +1,119 @@
...
+#FROM intel/oneapi-basekit:devel-ubuntu18.04
+
+ENV SOS_INSTALL=/home/sos
+RUN mkdir /home/sos
+ENV DEBIAN_FRONTEND=noninteractive
+ENV CC=gcc
+ENV CXX=g++
+# For OneAPI Compilers use clang and clang++
+  #ENV CC=clang
+  #ENV CXX=clang++
```
davidozog, line 11: Unfortunately, I don't think `clang` will work anymore starting with oneAPI v2022.x.  Have you tried it recently?  I'm curious what version is included with this `devel-ubuntu18.04` image?

I can't find any release notes about it yet online, but `compiler/2022.0.2/linux/bin-llvm/README.txt` in the 2022.1.2 Base Toolkit it says: 

```quote
1. If invoking "clang" or "clang++" directly, please try to use icx/icpx/dpcpp
   instead. Direct use of clang as a driver is not supported. The "clang"
   drivers do not have the same behavior as either icx or upstream LLVM, and
   are not widely tested.

2. If for some reason you must use the OneAPI clang/clang++, please report the
   issue. Until the issue is resolved, you can find the clang/clang++ driver's
   location by running "dpcpp --print-prog-name=clang".
```

Anyway, this is probably ok, especially if it still works - just something to keep in mind...

kholland-intel: `clang` does still work for now, as long as the Dockerfile builder's `PATH` points to the OneAPI compiler bin area where `clang`, `icx`, etc. are (The builder's `PATH` is updated in the Dockerfile using the `ENV` command). However, one thing to note is that when testing out this Dockerfile with the `icx` compiler, the build failed when building ucx due to a NaN and infinity comparison in one of their headers.

davidozog: Ok - thanks for getting to the bottom of that @kholland-intel

davidozog (approved):

scripts/docker/README.md
```diff
@@ -0,0 +1,76 @@
+# Sandia-OpenSHMEM Docker Image
+Sandia-OpenSHMEM (SOS) utlizes Docker images to provide users a means to create
```
wrrobin, line 2: We dont have a '-' between Sandia and OpenSHMEM, right?

scripts/docker/README.md
```diff
@@ -0,0 +1,76 @@
...
+## Creating SOS Containers from the Prebuilt Docker Images
+Coming Soon: Information regarding the SOS public Docker Hub repos and prebuilt
+images.
+
+## Creating SOS Containers from the Dockerfile 
+This Dockerfile provides a configurable Ubuntu sandbox for running SOS
+applications and benchmarks.
+
+### Building the Image from the Dockerfile
+To build the image from the Dockerfile, run the following run the following
```
wrrobin, line 15: I an seeing "run the following" twice here... I am having some eye issues recently and if you dont see the same on your end, please ignore.

scripts/docker/README.md
```diff
@@ -0,0 +1,76 @@
...
+```
+$ docker run -it <name-of-image>
+```
+
+Where the `-it` option makes the container interactive and allows the user to
+actually step into the container and execute commands in the new environment.
+
+### Dockerfile Options
+Throughout the Dockerfile there are lines that can be commented/uncommented to
+configure a particular testing environment and SOS build.
```
wrrobin, line 36: Should we add an example here?

scripts/docker/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+#### Ubuntu Image
+The first option this file provides is the the type of Ubuntu image docker
```
wrrobin, line 39: I am seeing two "the" here..

scripts/docker/README.md
```diff
@@ -0,0 +1,76 @@
...
+should pull from.
+
+By default the Dockerfile pulls a lightweight Ubuntu 20.04 image.
+The other option (which is commnented out) is for OneAPI image which uses
+Ubunutu 18.04. The idea for this image was to allow users to test/run SOS
+applications with non gcc compilers.
```
wrrobin, line 45: non-gcc compilers

scripts/docker/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+#### Building Libafabic
+The next set of options are presented when it comes time to build Libfabric.
+
+By default, a basic libfabric configuration is built and installed to an
```
wrrobin, line 50: L  in libfabric can be capitalized to match other cases..

scripts/docker/README.md
```diff
@@ -0,0 +1,76 @@
...
+By default, SOS is built with a basic Libfabric provider.
+The next few options are:
+* Build SOS with oneapi compilers (i.e. clang and clang++)
+* Build SOS with the GNI provider
+* Build SOS with the Libfabric provider and PMIx
+* Build SOS with Portals 4
+* Build SOS with UCX
+
+As a reminder which ever configuration option is chosen must be uncommented,
+while the other options must be commented.
```
wrrobin, line 68: Rewrite "As a reminder,  chosen configuration option must be uncommented while the other options must be commented out."

scripts/docker/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+#### Enabling the PRRTE Server
+The last portion of code that is optional in this Dockerfile is that which
+enables the PRRTE Server.
+This is mostly used to run SOS tests when the GNI provider has been enabled in
+the Libfabric build.
+When testing on the GNI provider uncomment out the lines in this portion to
```
wrrobin, line 75: It might be better to say the specific line numbers to be uncommented. Same for the previous cases.

wrrobin (approved):

scripts/docker/README.md
kholland-intel: Instead of pointing to a specific line number, I suggested user comment out the code directly under a specific subheading (I felt like the headers were less likely to change than the line numbers). Does this work or do you still prefere that we point them to a specific line number? Examples can be found in the README changes at lines 52 and 73.

scripts/docker/Dockerfile
davidozog: I'm not quite sure... if you can think of a good upstream fix, but all means go for it (I can't think of an easy fix at the moment, but it must be possible).

Otherwise, I think a comment in the README is probably fine.  We could also just remove it or comment it out... UCX is not being used in the default build right?

kholland-intel: Correct, UCX is not being used by default. Maybe I can comment it out and put a note in the README that it does not work with the `icx` compiler.

scripts/docker/README.md
```diff
@@ -36,41 +36,40 @@ Throughout the Dockerfile there are lines that can be commented/uncommented to
...
+The first option this file provides is the type of Ubuntu image docker
 should pull from.
 
 By default the Dockerfile pulls a lightweight Ubuntu 20.04 image.
 The other option (which is commnented out) is for OneAPI image which uses
 Ubunutu 18.04. The idea for this image was to allow users to test/run SOS
-applications with non gcc compilers.
+applications with non-gcc compilers.
 
 #### Building Libafabic
```
davidozog, line 52: Libfabric

davidozog, line 52: Libfabric

scripts/docker/README.md
```diff
@@ -44,6 +44,11 @@ The other option (which is commnented out) is for OneAPI image which uses
 Ubunutu 18.04. The idea for this image was to allow users to test/run SOS
 applications with non-gcc compilers.
 
+Note: Currenlty, building the UCX libraries with the `icx` compiler leads to a
```
davidozog, line 47: Currently

scripts/docker/README.md
```diff
@@ -36,41 +36,40 @@ Throughout the Dockerfile there are lines that can be commented/uncommented to
...
+the other options must be commented out.
 
 #### Enabling the PRRTE Server
 The last portion of code that is optional in this Dockerfile is that which
 enables the PRRTE Server.
 This is mostly used to run SOS tests when the GNI provider has been enabled in
 the Libfabric build.
-When testing on the GNI provider uncomment out the lines in this portion to
-build a container which runs "make check" on the PRRTE Server.
+When testing on the GNI provider uncomment out the lines under "Start PRRTE
```
davidozog, line 78:
```suggestion
When testing on the GNI provider uncomment the lines under "Start PRRTE
```

scripts/docker/README.md
```diff
@@ -36,41 +36,40 @@ Throughout the Dockerfile there are lines that can be commented/uncommented to
...
 
-By default, a basic libfabric configuration is built and installed to an
-"install" directory.
-The next option is to build Libfabric with the non-gcc compilers, in this
-case, using clang and clang++.
-Another option is to enable the GNI provider while building Libfabric.
+By default, a basic Libfabric configuration is built and installed to an
+"install" directory (This option is currently uncommented).
+The other option is to enable the GNI provider while building Libfabric. (For
+this option uncomment the line of code "To build Libfabric to use GNI provider")
```
davidozog, line 58:
```suggestion
this option uncomment the line of code under "To build Libfabric to use GNI provider")
```

davidozog:
Issues #1047 and #1048 manifested here, but are certainly unrelated to this PR.

PR #1045, opened 2022-02-16
Topic: pmi-simple - update to upstream (MPICH) state

davidozog:
This requires some slight modifications for SOS:
  * include `"config.h"` instead of `"mpichconf.h"`
  * define `MPI_MAX_PORT_NAME `manually (w/out including `"mpi.h"`)
  * carry new pmi-simple requirements from MPICH (`mpir_mem.h`/`mpl_sockaddr.h`)
  * ignore `strncasecmp`/`strnicmp` requirement (not used in pmi-simple)

I'm not sure we want to make this change, but it does eliminate several warnings we see in our current version of pmi-simple.  If we include it, we may want to defer this for a later SOS release (e.g., v1.5.2).

### DIFF
--- a/pmi-simple/mpir_mem.h
+++ b/pmi-simple/mpir_mem.h
@@ -1,0 +1,243 @@
+/*
+ * Copyright (C) by Argonne National Laboratory
+ *     See COPYRIGHT in top-level directory
+ */
+
+#ifndef MPIR_MEM_H_INCLUDED
+#define MPIR_MEM_H_INCLUDED
+
+#include "config.h"
+
+/* Make sure that we have the definitions for the malloc routines and size_t */
+#include <stdio.h>
+#include <stdlib.h>
+/* strdup is often declared in string.h, so if we plan to redefine strdup,
+   we need to include string first.  That is done below, only in the
+   case where we redefine strdup */
+
+#if defined(__cplusplus)
+extern "C" {
+#endif
+
+#include "mpl.h"
+
+/* Define attribute as empty if it has no definition */
+#ifndef ATTRIBUTE
+#define ATTRIBUTE(a)
+#endif
+
+#if defined (MPL_USE_DBG_LOGGING)
+    extern MPL_dbg_class MPIR_DBG_STRING;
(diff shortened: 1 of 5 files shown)


davidozog:
Closing for now, tracking with issue #1046, and adding to the `v1.5.x` Github Milestone

davidozog:
This version of pmi-simple best aligns with pmodels/MPICH v3.2.1.  Beyond that there were more significant changes that may not be practical to incorporate into SOS at the moment.

However, this does eliminate all the compiler warnings from #1046 and passes CI, so I think now is possibly a good time to try it out for the v1.5.3 release.

pmi-simple/mpir_mem.h
```diff
@@ -0,0 +1,243 @@
+/*
+ * Copyright (C) by Argonne National Laboratory
+ *     See COPYRIGHT in top-level directory
+ */
```
wrrobin, line 4: Just querying, do we have to change anything here?

wrrobin (approved):
LGTM. The tests also do not show any more warnings. One question: the 3 new files - are they all verbatim to upstream or do we have some changes?

bcmIntc (approved):

bcmIntc:
Question regarding "TODOs" left in the code.  Should these be removed/resolved or remove/CR, or left as-is before upstream?

parkerha1 (approved):

davidozog:
> LGTM. The tests also do not show any more warnings. One question: the 3 new files - are they all verbatim to upstream or do we have some changes?

@wrrobin - the tests are verbatim _except_ for the 4 bullet items listed in the PR description, which are minimal but needed to be compatible with SOS.

> Question regarding "TODOs" left in the code. Should these be removed/resolved or remove/CR, or left as-is before upstream?

@bcmIntc - since we keeping these changes "upstream" with respect to MPICH, we should keep as much as possible identical so we don't confuse ourselves or create unnecessary work if/when we update.  However, if we fork off then we can remove/address these, but it's something we have avoided in the past for simplicity.

PR #1051, opened 2022-04-04
Error-checking: Add checks for dst/src overlap

davidozog:
Features:
  * Fixes bug for in-place broadcasts (on the root PE).
  * Adds error-checking macro for proper buffer overlap.
  * No support yet for the strided APIs and put w/ signal.
  * Adds a simple unit test for in-place broadcast.

### DIFF
--- a/src/shmem_internal.h
+++ b/src/shmem_internal.h
@@ -285,20 +285,28 @@ extern unsigned int shmem_internal_rand_seed;
         }                                                                               \
     } while (0)
 
-#define SHMEM_ERR_CHECK_OVERLAP(dest, source, size, complete_overlap_allowed)           \
+/* Indicates whether the memory region pointed to by ptr1 (having size1) overlaps with
+ * the buffer at ptr2 (having size2).
+ *   SHMEM usage examples:
+ *     For the collective routines (src/dest arguments):
+ *         size1 == size2 and complete_overlap_allowed is 1
+ *     For the pt2pt sync routines (ivars/status/indices arguments):
+ *         size1 != size2 and complete_overlap_allowed is 0 */
+#define SHMEM_ERR_CHECK_OVERLAP(ptr1, ptr2, size1, size2, complete_overlap_allowed)     \
     do {                                                                                \
-        const void *dst = (void*)(dest);                                                \
-        const void *src = (void*)(source);                                              \
-        const void *ptr_low  = dst > src ? src : dst;                                   \
-        const void *ptr_high = dst > src ? dst : src;                                   \
-        const void *ptr_extent = (void *)((char *)(ptr_low) + (size));                  \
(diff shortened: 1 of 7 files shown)


davidozog:
Oops - this check for overlapping buffers won't work on the point-to-point sync routines without some improvements.  I've labelled this PR as a WIP.

src/shmem_internal.h
```diff
@@ -285,6 +285,32 @@ extern unsigned int shmem_internal_rand_seed;
...
+        const void *p1 = (void*)(ptr1);                                                 \
+        const void *p2 = (void*)(ptr2);                                                 \
+        const void *ptr_low  = p1 > p2 ? p2 : p1;                                       \
+        const void *ptr_high = p1 > p2 ? p1 : p2;                                       \
+        const size_t sz_low  = p1 > p2 ? size2 : size1;                                 \
+        const void *ptr_extent = (void *)((char *)ptr_low + sz_low);                    \
+        if (complete_overlap_allowed && p1 == p2) {                                     \
+            break; /* Skip this check when buffer is allowed to completely overlap  */  \
+        }                                                                               \
+        if (ptr_extent > ptr_high) {                                                    \
```
wrrobin, line 306: Is `ptr_extent > ptr_high + sz_high` a complete overlap case?

davidozog: Good question - my understanding of the OpenSHMEM spec is that it's only a complete overlap when the two buffers start at the same address and "have the same size" (but there is only 1 size argument in APIs where complete overlap is allowed: the RMA and collective routines).

I'm not sure the spec says that explicitly though - is your interpretation different @wrrobin?  Do I understand your concern correctly?

test/unit/bcast_in_place.c
```diff
@@ -0,0 +1,64 @@
...
+    npes = shmem_n_pes();
+
+    for (int i = 0; i < NELEM; i++)
+        src[i] = me + i;
+
+    shmem_barrier_all();
+
+    shmem_long_broadcast(SHMEM_TEAM_WORLD, src, src, NELEM, 0);
+
+    /* Validate reduced data */
```
davidozog, line 52:
```suggestion
    /* Validate broadcasted data */
```

test/unit/bcast_in_place.c
```diff
@@ -0,0 +1,64 @@
...
+
+int main(void)
+{
+    int me, npes;
+    int errors = 0;
+
+    shmem_init();
+
+    me = shmem_my_pe();
+    npes = shmem_n_pes();
```
davidozog, line 43: `npes` not used.

src/shmem_internal.h
davidozog: Yes, thanks @wrrobin.  This is resolved and I corrected a small mistake in my response above: complete overlap is not allowed in RMA.

PR #1053, opened 2022-04-19
Shmem team ptr

kholland-intel:
Implemented and added a unit test for the v1.6 API, shmem_team_ptr

### DIFF
--- a/src/remote_pointer_c.c
+++ b/src/remote_pointer_c.c
@@ -30,6 +30,9 @@
 #pragma weak shmem_ptr = pshmem_ptr
 #define shmem_ptr pshmem_ptr
 
+#pragma weak shmem_team_ptr = pshmem_team_ptr
+#define shmem_team_ptr pshmem_team_ptr
+
 #endif /* ENABLE_PROFILING */
 
 
@@ -41,4 +44,20 @@ shmem_ptr(const void *target, int pe)
     SHMEM_ERR_CHECK_SYMMETRIC(target, 1);
 
     return shmem_internal_ptr(target, pe);
+}
+
+void SHMEM_FUNCTION_ATTRIBUTES *
+shmem_team_ptr(shmem_team_t team, const void *target, int pe)
+{
+    if (team == SHMEM_TEAM_INVALID) {
+        return NULL;
+    } else {
+      SHMEM_ERR_CHECK_INITIALIZED();
+      SHMEM_ERR_CHECK_PE(pe);
+      SHMEM_ERR_CHECK_SYMMETRIC(target, 1);
+
+      int global_pe = shmem_team_translate_pe(team, pe, SHMEM_TEAM_WORLD);
+
+      return shmem_internal_ptr(target, global_pe);
+    }
 }
--- a/test/unit/shmem_ptr.c
+++ b/test/unit/shmem_ptr.c
@@ -52,11 +52,21 @@ int main(void) {
     for (i = n = 0; i < npes; i++) {
         int * ptr = (int *) shmem_ptr(&shr_data, i);
 
-        if (ptr != NULL) {
+        //Kieran
+        int * team_ptr = (int *) shmem_team_ptr(SHMEM_TEAM_WORLD, &shr_data, i);
+
+        //if (ptr != NULL) {
+        // Kieran
+        if (ptr != NULL && team_ptr != NULL) {
             int shr_peer = *ptr;
+
+            //Kieran
+            int shr_team_peer = *team_ptr;
             ++n;
 
-            if (shr_peer != i) {
+            //if (shr_peer != i) {
+            //Kieran 
+            if (shr_peer != i && shr_team_peer != i) {
                 printf("%2d: Error, shr_data(%d) = %d, expected %d\n", me, i, shr_peer, i);
                 errors++;
             }
--- a/test/unit/shmem_ptr.c
+++ b/test/unit/shmem_ptr.c
@@ -52,21 +52,11 @@ int main(void) {
     for (i = n = 0; i < npes; i++) {
         int * ptr = (int *) shmem_ptr(&shr_data, i);
 
-        //Kieran
-        int * team_ptr = (int *) shmem_team_ptr(SHMEM_TEAM_WORLD, &shr_data, i);
-
-        //if (ptr != NULL) {
-        // Kieran
-        if (ptr != NULL && team_ptr != NULL) {
+        if (ptr != NULL) {
             int shr_peer = *ptr;
-
-            //Kieran
-            int shr_team_peer = *team_ptr;
             ++n;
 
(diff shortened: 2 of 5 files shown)


test/unit/shmem_team_ptr.c
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    printf("%2d: Found %d world team data segment peer(s)\n", me, n);
+    fflush(NULL);
+    shmem_barrier_all();
+
+    /* Check shmem_team_ptr on heap segment */
+    shr_heap = shmem_malloc(sizeof(int));
+    *shr_heap = me;
+
```
davidozog, line 76: I think I see a race here, do you see it @kholland-intel?  I have a feeling it's the cause of your test failure (maybe XPMEM + RVA exposes the race more readily than other configs).

davidozog, line 76: I think I see a race here, do you see it @kholland-intel?  I have a feeling it's the cause of your test failure (maybe XPMEM + RVA exposes the race more readily than other configs).

kholland-intel: I think I might... Is it that there should be a barrier after `*shr_heap = me` ? I suppose a PE could fly through the upcoming `for` loop before a particular PE in the team sets its `*shr_heap` value.

davidozog: Yup, that's it exactly.

A `shmem_barrier_all` would work, but I think you only need a `shmem_sync_all` here and that might be a little less expensive (and a good thing to cover in testing).  Would you mind putting a `shmem_sync_all` here and see if it fixes the XPMEM+RVA test issue?

kholland-intel: Awesome, my latest patch should address the data race. Thank you for catching this!

test/unit/shmem_team_ptr.c
```diff
@@ -0,0 +1,170 @@
...
+    shmem_barrier_all();
+
+    if (npes == 1) {
+        shmem_finalize();
+        return errors != 0;
+    }
+
+    // Create a team of PEs w/ odd numbered id's
+    shmem_team_t new_team;
+    shmem_team_split_strided(SHMEM_TEAM_WORLD, 1, 2, npes / 2, NULL, 0, &new_team);
```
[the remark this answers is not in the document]
kholland-intel: I updated the test to use shmem_team_sync here, thanks!

src/remote_pointer_c.c
```diff
@@ -42,3 +45,19 @@ shmem_ptr(const void *target, int pe)
...
+
+void SHMEM_FUNCTION_ATTRIBUTES *
+shmem_team_ptr(shmem_team_t team, const void *target, int pe)
+{
+    if (team == SHMEM_TEAM_INVALID) {
+        return NULL;
+    } else {
+      SHMEM_ERR_CHECK_INITIALIZED();
+      SHMEM_ERR_CHECK_PE(pe);
+      SHMEM_ERR_CHECK_SYMMETRIC(target, 1);
```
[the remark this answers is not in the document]
kholland-intel: I see, I have refactored the code to be more consistent with the other user-facing routines. Thanks!

test/unit/shmem_team_ptr.c
```diff
@@ -0,0 +1,162 @@
...
+
+    shr_world_data = me;
+
+    shmem_barrier_all();
+
+    /* Check shmem_team_ptr on data segment */
+    for (i = n = 0; i < npes; i++) {
+        int * world_ptr = (int *) shmem_team_ptr(SHMEM_TEAM_WORLD, &shr_world_data, i);
+
+        if (world_ptr != NULL) {
```
wrrobin, line 51: IIUC, this if condition is not mutually exclusive to the `else if` condition below. It is just a thought that this structure could have been written like this:
```
if (i == me && world_ptr != NULL)
{
 ...
} else {
 Error..
}
```

test/unit/shmem_team_ptr.c
```diff
@@ -0,0 +1,162 @@
...
+            else if (i == team_me) {
+                printf("%2d: Error, shmem_team_ptr(heap) returned NULL for my PE\n", team_me);
+                errors++;
+            }
+        }
+        printf("%2d: Found %d new team heap segment peer(s)\n", team_me, n);
+    }
+
+    shmem_finalize();
+
```
wrrobin, line 160: `shmem_team_destroy` ? Just for the sake of "in"sanity...

wrrobin (approved):
LGTM. Two minor comments and can be ignored.

davidozog:
@wrrobin and @kholland-intel - should this be `shmemx`?

wrrobin:
Good question @davidozog. We have not added this change as an errata, right? In either case, I think this can go as `shmemx`.

test/unit/shmem_ptr.c
```diff
@@ -62,18 +62,22 @@ int main(void) {
             }
         }
         else if (i == me) {
-                printf("%2d: Error, shmem_ptr(data) returned NULL for my PE\n", me);
-                errors++;
+            printf("%2d: Error, shmem_ptr(data) returned NULL for my PE\n", me);
+            errors++;
+        }
+        else {
+            //shmem_ptr returned NULL for the other available PEs (enable XPMEM)
```
wrrobin, line 69: If we want to leave a comment here, may be we can modify it a little:
```
NULL is returned for remotely inaccessible data objects. Enabling XPMEM / CMA will allow non-NULL value for on-node PEs
```

test/unit/shmem_team_ptr.c
```diff
@@ -125,8 +136,12 @@ int main(void) {
...
                 errors++;
             }
+            else {
+                //shmem_team_ptr returned NULL for the other available PEs (enable XPMEM)
+                ++available;
+            }
         }
-        printf("%2d: Found %d new team data segment peer(s)\n", team_me, n);
+            printf("%2d: Found %d world team data segment peer(s) (%d were inaccessible)\n", me, n, available);
         fflush(NULL);
```
wrrobin, line 145: Typo in indentation?

test/unit/shmem_team_ptr.c
```diff
@@ -152,10 +167,15 @@ int main(void) {
                 printf("%2d: Error, shmem_team_ptr(heap) returned NULL for my PE\n", team_me);
                 errors++;
             }
+            else {
+                //shmem_team_ptr returned NULL for the other available PEs (enable XPMEM)
+                ++available;
+            }
         }
-        printf("%2d: Found %d new team heap segment peer(s)\n", team_me, n);
+            printf("%2d: Found %d world team heap segment peer(s) (%d were inaccessible)\n", me, n, available);
```
wrrobin, line 175: Same as above.

wrrobin (approved):

Issue #897, opened 2019-11-11
SHMEM_ERR_CHECK_* macros should use RAISE_ERROR_*


davidozog:
In `shmem_internal.h`, the `SHMEM_ERR_CHECK*` macros should use the `RAISE_ERROR_*` macros so that the function, filename, and line number are printed.

PR #1058, opened 2022-08-04
Updated error checking macros

kholland-intel:
Invoked RAISE_ERROR_* functions from SHMEM_ERR_CHECK_* macros so that the
function, filename, and line numbers are printed.

### DIFF
--- a/src/shmem_internal.h
+++ b/src/shmem_internal.h
@@ -170,27 +170,23 @@ extern unsigned int shmem_internal_rand_seed;
 #define SHMEM_ERR_CHECK_INITIALIZED()                                   \
     do {                                                                \
         if (!shmem_internal_initialized) {                              \
-            fprintf(stderr, "ERROR: %s(): " PACKAGE_NAME " library not initialized\n", \
-                    __func__);                                          \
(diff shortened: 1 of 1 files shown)


kholland-intel:
This PR will resolve issue #897

src/shmem_internal.h
```diff
@@ -170,44 +170,38 @@ extern unsigned int shmem_internal_rand_seed;
 #define SHMEM_ERR_CHECK_INITIALIZED()                                   \
     do {                                                                \
         if (!shmem_internal_initialized) {                              \
-            fprintf(stderr, "ERROR: %s(): " PACKAGE_NAME " library not initialized\n", \
-                    __func__);                                          \
-            abort();                                                    \
+            RAISE_ERROR_STR(PACKAGE_NAME " library not initialized\n"); \
```
wrrobin, line 173: I think this is the only place where we cant reuse these macros. The SHMEM runtime has not been initialized, so we cant call `shmem_runtime_abort`. May be we can do
```
RETURN_ERROR_STR
abort()
```

wrrobin (approved):

PR #1059, opened 2022-08-04
Enable OFI provider for Slingshot

wrrobin:
Signed-off-by: Md Rahman <md.rahman@intel.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -214,10 +214,10 @@ AC_ARG_ENABLE([mr-endpoint],
 AS_IF([test "$enable_mr_endpoint" = "yes"],
       [AC_DEFINE([ENABLE_MR_ENDPOINT], [1], [If defined, the OFI transport will use FI_MR_ENDPOINT])])
 
-AC_ARG_ENABLE([fi-manual-progress],
-    [AC_HELP_STRING([--enable-fi-manual-progress],
+AC_ARG_ENABLE([ofi-manual-progress],
+    [AC_HELP_STRING([--enable-ofi-manual-progress],
                     [Use FI_MANUAL_PROGRESS for data progress control mode. (default: disabled)])])
-AS_IF([test "$enable_fi_manual_progress" = "yes"],
+AS_IF([test "$enable_ofi_manual_progress" = "yes"],
       [AC_DEFINE([ENABLE_FI_MANUAL_PROGRESS], [1], [If defined, the OFI will use FI_MANUAL_PROGRESS as data progress mode. This is currently required for cxi provider.])])
 
 AC_ARG_ENABLE([max-teams],
--- a/configure.ac
+++ b/configure.ac
@@ -208,6 +208,12 @@ AS_CASE([$enable_ofi_mr],
              AC_DEFINE([ENABLE_MR_SCALABLE], [1], [If defined, the OFI transport will use FI_MR_SCALABLE])],
         [AC_MSG_ERROR([Invalid OFI memory registration mode: $enable_ofi_mr])])
 
+AC_ARG_ENABLE([mr-endpoint],
+    [AC_HELP_STRING([--enable-mr-endpoint],
+                    [Use FI_MR_ENDPOINT to enable cxi provider. (default: disabled)])])
+AS_IF([test "$enable_mr_endpoint" = "yes"],
+      [AC_DEFINE([ENABLE_MR_ENDPOINT], [1], [If defined, the OFI will use FI_MR_ENDPOINT during fi_mr_reg. Otherwise, the basic memory registration will be used.])])
+
 AC_ARG_ENABLE([max-teams],
     [AC_HELP_STRING([--enable-max-teams=NUMBER],
                     [Default value for the maximum number of teams allowed (default: 10)])])
--- a/configure.ac
+++ b/configure.ac
@@ -220,7 +220,6 @@ AC_ARG_ENABLE([fi-manual-progress],
 AS_IF([test "$enable_fi_manual_progress" = "yes"],
       [AC_DEFINE([ENABLE_FI_MANUAL_PROGRESS], [1], [If defined, the OFI will use FI_MANUAL_PROGRESS as data progress mode. This is currently required for cxi provider.])])
 
-
 AC_ARG_ENABLE([max-teams],
     [AC_HELP_STRING([--enable-max-teams=NUMBER],
                     [Default value for the maximum number of teams allowed (default: 10)])])
--- a/configure.ac
+++ b/configure.ac
@@ -212,7 +212,14 @@ AC_ARG_ENABLE([mr-endpoint],
     [AC_HELP_STRING([--enable-mr-endpoint],
                     [Use FI_MR_ENDPOINT to enable cxi provider. (default: disabled)])])
 AS_IF([test "$enable_mr_endpoint" = "yes"],
-      [AC_DEFINE([ENABLE_MR_ENDPOINT], [1], [If defined, the OFI will use FI_MR_ENDPOINT during fi_mr_reg. Otherwise, the basic memory registration will be used.])])
+      [AC_DEFINE([ENABLE_MR_ENDPOINT], [1], [If defined, the OFI transport will use FI_MR_ENDPOINT])])
+
+AC_ARG_ENABLE([fi-manual-progress],
+    [AC_HELP_STRING([--enable-fi-manual-progress],
+                    [Use FI_MANUAL_PROGRESS for data progress control mode. (default: disabled)])])
+AS_IF([test "$enable_fi_manual_progress" = "yes"],
+      [AC_DEFINE([ENABLE_FI_MANUAL_PROGRESS], [1], [If defined, the OFI will use FI_MANUAL_PROGRESS as data progress mode. This is currently required for cxi provider.])])
+
 
 AC_ARG_ENABLE([max-teams],
     [AC_HELP_STRING([--enable-max-teams=NUMBER],
--- a/src/shmem_synchronization.h
+++ b/src/shmem_synchronization.h
@@ -174,6 +174,9 @@ shmem_internal_fence(shmem_ctx_t ctx)
         }                                                               \
     } while(0)
 
+/* Polling based wait is required for providers that need 
+ * manual progress, i.e., cxi. This is enabled through 
+ * ENABLE_FI_MANUAL_PROGRESS */
 #if defined(ENABLE_HARD_POLLING) || defined(ENABLE_FI_MANUAL_PROGRESS)
 #define SHMEM_INTERNAL_WAIT_UNTIL(var, cond, value)                     \
     SHMEM_WAIT_UNTIL_POLL(var, cond, value)
--- a/src/shmem_synchronization.h
+++ b/src/shmem_synchronization.h
@@ -174,7 +174,7 @@ shmem_internal_fence(shmem_ctx_t ctx)
         }                                                               \
     } while(0)
 
-#if defined(ENABLE_HARD_POLLING)
+#if defined(ENABLE_HARD_POLLING) || defined(ENABLE_MR_ENDPOINT)
 #define SHMEM_INTERNAL_WAIT_UNTIL(var, cond, value)                     \
     SHMEM_WAIT_UNTIL_POLL(var, cond, value)
 #define SHMEM_INTERNAL_SIGNAL_WAIT_UNTIL(var, cond, value, sat_value)   \
--- a/src/shmem_synchronization.h
+++ b/src/shmem_synchronization.h
@@ -174,7 +174,7 @@ shmem_internal_fence(shmem_ctx_t ctx)
         }                                                               \
     } while(0)
 
-#if defined(ENABLE_HARD_POLLING) || defined(ENABLE_MR_ENDPOINT)
+#if defined(ENABLE_HARD_POLLING) || defined(ENABLE_FI_MANUAL_PROGRESS)
 #define SHMEM_INTERNAL_WAIT_UNTIL(var, cond, value)                     \
     SHMEM_WAIT_UNTIL_POLL(var, cond, value)
 #define SHMEM_INTERNAL_SIGNAL_WAIT_UNTIL(var, cond, value, sat_value)   \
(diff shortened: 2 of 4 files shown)


src/shmem_synchronization.h
```diff
@@ -174,7 +174,7 @@ shmem_internal_fence(shmem_ctx_t ctx)
         }                                                               \
     } while(0)
 
-#if defined(ENABLE_HARD_POLLING)
+#if defined(ENABLE_HARD_POLLING) || defined(ENABLE_MR_ENDPOINT)
```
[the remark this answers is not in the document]
wrrobin: Yes, this is confusing. I think, `FI_PROGRESS_MANUAL` requires polling and we enable manual progress through `FI_MR_ENDPOINT`. We should use a different flag, I guess for manual progress.

src/transport_ofi.c
```diff
@@ -729,8 +767,13 @@ int publish_mr_info(void)
             heap_key = fi_mr_key(shmem_transport_ofi_target_heap_mrfd);
             data_key = fi_mr_key(shmem_transport_ofi_target_data_mrfd);
         } else {
+#ifdef ENABLE_MR_ENDPOINT
+            heap_key = 1;
+            data_key = 0;
+#else
```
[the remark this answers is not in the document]
wrrobin: Nope. This is a specific change that I verified with Alexia. Not sure why though.

src/transport_ofi.c
```diff
@@ -1155,7 +1198,11 @@ int query_for_fabric(struct fabric_info *info)
                                    for put with signal implementation */
 #endif
     hints.addr_format         = FI_FORMAT_UNSPEC;
+#ifdef ENABLE_MR_ENDPOINT
+    domain_attr.data_progress = FI_PROGRESS_MANUAL;
+#else
```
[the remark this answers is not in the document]
wrrobin: Right.. we dont have any tests with enabling MR_ENDPOINT and on sockets, I saw the progress issue that you might be pointing to. Strangely, for CXI, I do not see the progress issue with PROGRESS_MANUAL.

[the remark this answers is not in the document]
wrrobin: Any suggestion for change here, @davidozog ?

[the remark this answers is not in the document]
davidozog: No, I'm good with this change; but maybe in #1077 we can try to warn or preclude users from this configuration except when using providers where it makes sense.

src/shmem_synchronization.h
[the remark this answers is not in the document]
davidozog: Ah, I understand now.

Yes, I do think a comment would go a long way here, thanks!  A reference to  the connection between `FI_MR_ENDPOINT` and `FI_PROGRESS_MANUAL` in the ofi transport layer should be sufficient.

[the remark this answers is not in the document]
wrrobin: @davidozog - based on the email conversation with Alexia, I think, it is better not to use the `ENABLE_MR_ENDPOINT` as flag here, which will be confusing as to how MR_ENDPOINT can contribute to progress. I think one way we can handle this is not have a flag for enabling `FI_PROGRESS_MANUAL`. The problem is that we already have `--enable-manual-progress` which does not really enable `FI_PROGRESS_MANUAL` :(. I am open to thoughts / suggestions here.

src/transport_ofi.c
```diff
@@ -678,6 +678,21 @@ int allocate_recv_cntr_mr(void)
     /* Register separate data and heap segments using keys 0 and 1,
      * respectively.  In MR_BASIC_MODE, the keys are ignored and selected by
      * the provider. */
+#ifdef ENABLE_MR_ENDPOINT
```
jdinan, line 681: Hey Guys, sorry about lobbing the passing comment, just wanted to share a little historical context on this ifdef spaghetti.

There are a variety of things in the OFI transport that date back to the early days and no longer appropriate/correct given current libfabric. For example, this code should really be gated off of whether the provider returns `FI_MR_PROV_KEY` on the domain. We didn't have this back when Mr. Basic and Mr. Scalable were a thing. Enabling user-selected keys (`!FI_MR_PROV_KEY`) only for Mr. Endpoint is incorrect because a fabric can choose to return both `FI_MR_ENDPOINT` and `FI_MR_PROV_KEY`. Resolving the key value at runtime adds instructions to the critical path, so you probably want a build flag that will enable support for provider-selected memory registration keys.

wrrobin: @jdinan Thanks for your input here. The OFI manpage says that for `FI_MR_ENDPOINT`, memory registration occurs at the endpoint level, rather than domain. In such case, does it matter if provider returns `FI_MR_PROV_KEY`? 

We are also looking at #1077 where MR flags will be selected based on the provider in user, thus limit the users to provide any option for the MR flags. As you said, there is no build flag for `FI_MR_PROV_KEY` from the user. Does it make sense to select these flags based on the provider rather provide the option to the user? 

Also, looking at the provider feature matrix, it looks like `FI_MR_PROV_KEY` is required in conjunction with `FI_MR_ALLOCATED` and `FI_MR_LOCAL` - almost equivalent to what we have in SOS for `--enable-ofi-mr=basic`. I am trying to understand, if having a separate build flag for `FI_MR_PROV_KEY` will lead to the same flags for basic.

davidozog: Thanks @jdinan - we definitely missed that.

The comment at line 679 suggests that these user-selected keys are ignored by the provider when in MR_BASIC (FI_MR_PROV_KEY)  mode, but I'm not seeing anything specifying that in the libfabric manpages.  I'm thinking we should be careful here and only provide keys when the provider does not require `FI_MR_PROV_KEY` and use `fi_mr_key` when it does.

Also, is resolving the key value here really on the critical path?  I thought this occurred in `shmem_transport_init -> shmem_transport_ofi_target_ep_init`.

src/shmem_synchronization.h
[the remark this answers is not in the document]
davidozog: I can think of a few viable options:

1. Rename `--enable-manual-progress` to something like `--enable-progress-probe` and use `--enable-manual-progress` for `FI_MANUAL_PROGRESS`.  I don't like this so much because it's changing the meaning of the flag without user's knowledge.
2. Create a new flag called `--enable-fi-manual-progress`.
4. Change the configury to support something like `ENABLE_WAIT_POLLING` which is enabled for either hard-polling or `MR_ENDPOINT`.
5. Just keep this code as is for now (w/ comments) and fix things in PR #1077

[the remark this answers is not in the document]
wrrobin: I think I like option 2. I was also thinking about that. 

Option 3 is also good. The only thing I am thinking is it does not tell that the `FI_MANUAL_PROGRESS` is being enabled here. As of now, the current SOS supports `FI_MANUAL_PROGRESS` only for `MR_ENDPOINT` and that is only used for this provider. In the future, we might need to revisit if `--enable-fi-manual-progress` is needed but polling based wait is not for any particular provider.

Not trying to convince you, but trying to come up with different reasons.

src/transport_ofi.c
wrrobin: I realized later that this particular ifdef block for MR-ENDPOINT was to ensure that an `uint64_t` key is provided. I confirmed with OFI developers that we can use this key for all MR modes except `FI_MR_RAW`. So, I removed the specific check for MR-ENDPOINT. 
@jdinan - Does this look ok to you?

src/transport_ofi.c
```diff
@@ -729,8 +767,13 @@ int publish_mr_info(void)
             heap_key = fi_mr_key(shmem_transport_ofi_target_heap_mrfd);
             data_key = fi_mr_key(shmem_transport_ofi_target_data_mrfd);
         } else {
+#ifdef ENABLE_MR_ENDPOINT
+            heap_key = 1;
+            data_key = 0;
+#else
             heap_key = 1ULL;
             data_key = 0ULL;
+#endif
```
jdinan, line 762: The if and else cases look the same?

src/transport_ofi.c
jdinan: SOS uses `ENABLE_MR_SCALABLE` as the build flag for determining whether to support `FI_MR_PROV_KEY`. For an example, you can look at `shmem_transport_ofi_get_mr` where we choose at compile time whether to reference the table of MR keys. This is on the critical path, so you may want to have a build flag if you want to continue to support a static decision regarding provider-selected MR keys. Apologies, I haven't looked at #1077 to see if this is addressed there. IIUC, `FI_MR_BASIC` and `FI_MR_SCALABLE` have been long deprecated and it would be a healthy change to remove them from all of the SOS MR logic.

src/transport_ofi.c
wrrobin: Thanks for catching this. Should have been corrected with the previous commit.

src/shmem_synchronization.h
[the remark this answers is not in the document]
davidozog: Ah, I do see that issue with option 3, so yes I'm good with option 2.  That was indeed easy to convince me ;).

src/transport_ofi.c
wrrobin: Thanks @jdinan. In #1077, we are choosing MR modes based on the provider is use through `FI_PROVIDER` or `SHMEM_OFI_PROVIDER`. So, all the checks for `shmem_transport_ofi_get_mr` will be done at runtime as there will be no build flags. We may need to rethink there if that is adding too much overhead.

wrrobin:
@davidozog This is ready for a re-review. Thanks.

src/transport_ofi.h
```diff
@@ -1314,8 +1314,17 @@ void shmem_transport_atomic_fetch(shmem_transport_ctx_t* ctx, void *target,
                                   const void *source, size_t len, int pe,
                                   int datatype)
 {
+#ifdef ENABLE_MR_ENDPOINT
```
davidozog, line 1317: It seems this restriction is not directly related to `FI_MR_ENDPOINT` right?  But, I cannot think of a good alternative right now.. I'm just making the note that his might be possible to resolve/cleanup later, perhaps in #1077.

Also, I'm a little confused about this comment - let me make sure I understand.  Fetching atomics always have `FI_DELIVERY_COMPLETE` semantics via `fi_fetch_atomic` ([stated here](https://ofiwg.github.io/libfabric/v1.14.1/man/fi_atomic.3.html)), right?  But because `FI_DELIVERY_COMPLETE` atomics are not always supported, we can use `FI_INJECT` through `fi_fetch_atomicmsg` in those cases - is that it?  If so, I suppose another approach is for the blocking fetching atomic to have a branch (on the critical path 👎) that uses `fi_fetch_atomicmsg` on providers like this.

I mention that approach because I'm also concerned that only calling `fetch_atomic_nbi` here could lead to premature access of the result buffer.  Don't we need to quiet the operation before the user possibly accesses the result?  (It almost seems like a really weird use-case for `shmem_internal_pe_quiet` ;).

davidozog (commented):
I tested the changes and it's looking good ;)

I just have a naming suggestion and a concern about the NBI fetching AMO.

src/transport_ofi.h
wrrobin: Yes... I am not sure what else to use for the `ifdef` check. May be, a more restricted check will be by adding `ENABLE_FI_MANUAL_PROGRESS`. But, that will look even more weird. But, that check will definitely ensure that this is done for cxi provider.

I understand your concern. We can add that branch in the blocking fetch atomic.. but my thinking is we only want this for CXI, other providers can use `fi_fetch_atomic` with `FI_DELIVERY_COMPLETE`. Also, hoping that CXI will support this some day, and we can get rid of this small detour. 

The `quiet` here is not required, because this is originated from `shmem_internal_atomic_fetch` call at atomic_c.c4 line 396. There is a `get_wait` call there right after it. That will ensure that the result location is not accessed before the operation is complete. Let me know if I got it wrong here.

davidozog: Oh right - I keep forgetting about the `get_wait` in the top layer (and I don't see any other internal fetching atomics) - so that should be fine!

I agree on the other concerns - I'll mark this one as resolved!

configure.ac
```diff
@@ -212,7 +212,14 @@ AC_ARG_ENABLE([mr-endpoint],
     [AC_HELP_STRING([--enable-mr-endpoint],
                     [Use FI_MR_ENDPOINT to enable cxi provider. (default: disabled)])])
 AS_IF([test "$enable_mr_endpoint" = "yes"],
-      [AC_DEFINE([ENABLE_MR_ENDPOINT], [1], [If defined, the OFI will use FI_MR_ENDPOINT during fi_mr_reg. Otherwise, the basic memory registration will be used.])])
+      [AC_DEFINE([ENABLE_MR_ENDPOINT], [1], [If defined, the OFI transport will use FI_MR_ENDPOINT])])
+
+AC_ARG_ENABLE([fi-manual-progress],
```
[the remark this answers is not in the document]
wrrobin: @davidozog Made this change. Please check the latest commit.

src/transport_ofi.c
```diff
@@ -678,15 +678,17 @@ int allocate_recv_cntr_mr(void)
     /* Register separate data and heap segments using keys 0 and 1,
      * respectively.  In MR_BASIC_MODE, the keys are ignored and selected by
      * the provider. */
+    uint64_t key = 1;
```
davidozog, line 681: These key values are ignored by the provider if it requires `FI_MR_PROV_KEY` right?

I'm just checking whether or not we need a build flag for this.  My understanding is that this is a mode some providers require but SOS should support either way (which is why I hope the key value is ignored here if the mode is MR_BASIC/MR_PROV_KEY).  Is that the case?

davidozog (commented):
I think we're ready to merge this - just one more question...

src/transport_ofi.c
wrrobin: Yes, the provider is going to ignore the requested key, if SOS requests for `FI_MR_PROV_KEY` and the OFI provider also supports it. In such case, it will use its own key and we can extract that key with `fi_mr_key`.

davidozog (approved):

wrrobin:
@kholland-intel Could you try this PR on your end and see if everything looks ok? Need a review from your side as well.

kholland-intel (approved):
I have just tested the latest changes and everything passes for me. The code looks good too. 👍

PR #1062, opened 2022-09-15
NB fetch_add with 0 value for fetch nbi

wrrobin:


wrrobin:
Will track the issue using #1066. Closing without merge as this is not the complete solution.

PR #1063, opened 2022-09-27
workaround for hang at finalize

tmh97:
At Cornelis Networks we are seeing intermittent hangs at finalize when running e.g the basic ‘pingpong’ test in the unit directory. As we scale up from 2 ranks 1 ppn up to 14 ranks 7 ppn, we almost always run into a case where a process will hang.

Currently, our workaround is to call shmem_transport_quiet not just once, but several thousands of times. With this fix in place, it gives processes enough time to gracefully become inactive.

Obviously, this is not a proper solution, and certainly not scalable, and I am curious what would be the best way to resolve this issue. Also, I’m curious if anyone else has seen this issue, or if this has been discussed in the past.

### DIFF
--- a/src/transport_ofi.c
+++ b/src/transport_ofi.c
@@ -1758,7 +1758,10 @@ int shmem_transport_fini(void)
 
     /* The default context is not inserted into the list of contexts on
      * SHMEM_TEAM_WORLD, so it must be destroyed here */
-    shmem_transport_quiet(&shmem_transport_ctx_default);
+
+    for(int i=0; i< 100000; i++) {
+        shmem_transport_quiet(&shmem_transport_ctx_default);
+    }	
     shmem_transport_ctx_destroy(&shmem_transport_ctx_default);
 
     for (e = shmem_transport_ofi_stx_kvs; e != NULL; ) {


davidozog:
Thanks for reporting this @tmh97 - it's a very strange issue.  I've never come across this specifically, but have seen similar problems that were related to the transport layer configuration, progress model, or something similar.

As you probably know, there's no way we can accept this change as is - even 1 quiet is usually considered a very heavy OpenSHMEM operation ;).  If everything is working correctly, you should not have to quiet more than once here.

Is it possible to share more information about the OFI provider you are using and how exactly you are configuring SOS?  Do you see the problem go away when you do SHMEM_OFI_PROVIDER=sockets ?

And does this happen only with ping pong?  I would expect a problem like this should affect many if not all unit tests...

davidozog:
To be sure I just ran test/unit/pingpong.c on our OPA100 cluster (using the PSM2 provider) and was able to launch up to run a very large number of PPN (up to 256 PPN) on 2 nodes.  Feel free to contact me via email (david.m.ozog@intel.com) to work through it.

If you sent your program output after setting the SHMEM_DEBUG=1 environment variable, we can see what provider is picked up and how SOS was configured.

bcernohous:
re: "heavy operation".  This change impacts shmem_finalize.  Is performance a concern?

A shmem_quiet is sufficient for transmits and shouldn't be required more than once, as you say.  But a suspicion I have is the reliability protocol is involved.  A PE may have completed all local transmits but a peer may be waiting on a lost ACK and start reliability retries.  If the "done" rank doesn't drive progress, the ACK won't be retransmitted hanging the peer.

Just a theory. :) I don't have an answer if it's true.

And I've seen MPI_Finalize issues on the libfabrics github about "quiesce" issues that could be related to this.

davidozog:
@bcernohous - thanks for your thoughts!

```quote
re: "heavy operation". This change impacts shmem_finalize. Is performance a concern?
```
Right, finalize performance not a huge concern - but I think calling 10,000 quiet operations when we don't understand why the 1st one doesn't work is not the best practice ;).  I'd rather figure out what's happening with this particular libfabric provider than require all providers to do the ~10k quiets.  No (supported) provider has needed it so far...

I definitely agree there could be a lost ACK causing this like you describe.  Thinking aloud now... this last quiet on the default context during finalize does seems like a special case to me.  But we might be able to harden this, for example by having each PE drive local progress after the quiet but before the context is destroyed.  But hey, that might ends up not working or may be even more expensive than ~10k quiets.  For example, how long do we probe for progress?  We might be tempted to add a `shmem_sync` or similar before the ctx_destroy, which I think is not ideal.  Also, for all the other providers where the last quiet is working as expected, the other 9,999 quiets might actually be pretty inexpensive (but definitely not free).  So maybe it's not quite as crazy as I first thought...

Anyway, I'd still like to learn more about the configuration @tmh97.  If the details are something you'd rather discuss offline, by all means please feel free to email me.

tmh97:
> @bcernohous - thanks for your thoughts!
> 
> ```
> re: "heavy operation". This change impacts shmem_finalize. Is performance a concern?
> ```
> 
> Right, finalize performance not a huge concern - but I think calling 10,000 quiet operations when we don't understand why the 1st one doesn't work is not the best practice ;). I'd rather figure out what's happening with this particular libfabric provider than require all providers to do the ~10k quiets. No (supported) provider has needed it so far...
> 
> I definitely agree there could be a lost ACK causing this like you describe. Thinking aloud now... this last quiet on the default context during finalize does seems like a special case to me. But we might be able to harden this, for example by having each PE drive local progress after the quiet but before the context is destroyed. But hey, that might ends up not working or may be even more expensive than ~10k quiets. For example, how long do we probe for progress? We might be tempted to add a `shmem_sync` or similar before the ctx_destroy, which I think is not ideal. Also, for all the other providers where the last quiet is working as expected, the other 9,999 quiets might actually be pretty inexpensive (but definitely not free). So maybe it's not quite as crazy as I first thought...
> 
> Anyway, I'd still like to learn more about the configuration @tmh97. If the details are something you'd rather discuss offline, by all means please feel free to email me.

Thanks for all the insight David, emailing you now!

bcernohous:
> I'd rather figure out what's happening with this particular libfabric provider than require all providers to do the ~10k quiets. No (supported) provider has needed it so far...

I whole heartedly agree with figuring out the underlying issue.  I believe @tmh97 is seeing other issues/other providers which he will pursue in future issues.  This is somewhat of  work around to get us to those.   And he did open this issue with a question and would take suggestions:

>Obviously, this is not a proper solution, and certainly not scalable, and I am curious what would be the best way to resolve this issue. Also, I’m curious if anyone else has seen this issue, or if this has been discussed in the past.

I'll let you two work it out :)

tmh97:
Closing as this fix may only benefit opx!

PR #1064, opened 2022-09-27
Patch data segment init for M1 (and M2?) Macs

davidozog:
On MacOS, there's a problem with the initialization of the symmetric data segment (the symmetric heap is fine).  I can reproduce it on a 2020 M1 MacBook Pro, but not a 2016 Intel MacBook Pro.

SOS uses `get_etext` and `get_end` to find the data segment region, but [the manpage says](https://www.manpagez.com/man/3/get_etext/), "Use of these routines is very strongly discouraged.  The problem is that any program that is using UNIX link editor defined symbols (_end, _etext or _edata) is making assumptions that the program has the memory layout of a UNIX program."  I can tell that the determined memory layout is wrong on the M1 Mac, which causes any use of a global or static variable in an SOS communication routine to go out of bounds. 

It seems `mach_vm_region` is the standard API for gathering the segment properties (and it's very [poorly documented here](https://developer.apple.com/documentation/kernel/1402149-mach_vm_region)).

Oh, also `long double` is only 8 bytes on the M1 Mac, so I adjusted the iput128 unit test to use `__int128`.

Side note: MacOS is not a supported OS by any means, but it is an important environment for some developers (especially myself!).  This patch is just a best-effort hack to support laptop development of SOS on Macs.

### DIFF
--- a/src/init.c
+++ b/src/init.c
@@ -43,7 +43,9 @@
 #endif
 
 #ifdef __APPLE__
-#include <mach-o/getsect.h>
+#include <mach/mach.h>
+#include <mach/mach_vm.h>
+#include <assert.h>
 #else
 /* Declare data_start and end as weak to avoid a linker error if the symbols
  * are not present.  During initialization we check if the symbols exist. */
@@ -51,6 +53,27 @@
 #pragma weak _end
 extern int __data_start;
 extern int _end;
+#endif
+
+#ifdef __APPLE__
+vm_address_t get_base_address(mach_port_t task, size_t *region_size, void *addr)
+{
+    kern_return_t kret;
+    vm_region_basic_info_data_t info;
+    vm_size_t size;
+    mach_port_t object_name;
+    mach_msg_type_number_t count;
+    mach_vm_address_t address = (addr ? (mach_vm_address_t)addr : 0);
+    count = VM_REGION_BASIC_INFO_COUNT_64;
+
+    kret = mach_vm_region(task, &address, (mach_vm_size_t *)&size,
+                          VM_REGION_BASIC_INFO, (vm_region_info_t)&info,
+                          &count, &object_name);
+    assert(kret == KERN_SUCCESS);
+
+    *region_size = (size_t) size;
+    return address;
+}
 #endif
 
 void *shmem_internal_heap_base = NULL;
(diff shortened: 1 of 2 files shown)


src/init.c
```diff
@@ -53,6 +55,27 @@ extern int __data_start;
...
+    vm_size_t size;
+    mach_port_t object_name;
+    mach_msg_type_number_t count;
+    mach_vm_address_t address = (addr ? (mach_vm_address_t)addr : 0);
+    count = VM_REGION_BASIC_INFO_COUNT_64;
+
+    kret = mach_vm_region(task, &address, (mach_vm_size_t *)&size,
+                          VM_REGION_BASIC_INFO, (vm_region_info_t)&info,
+                          &count, &object_name);
+    assert(kret == KERN_SUCCESS);
```
wrrobin, line 72: I am trusting your experience with this change as I have little idea about it. One question, if the `assert` fails here, does that tell the user exactly why the program aborted?

src/init.c
```diff
@@ -283,8 +306,16 @@ shmem_internal_init(int tl_requested, int *tl_provided)
...
+    size_t region_size;
+    static int in_bss;
+
+    /* Start of this memory region (at address 0) */
+    vm_address_t addr = get_base_address(mach_task_self(), &region_size, 0);
+    shmem_internal_data_base = (void*)addr;
+
+    /* End of the BSS memory region (using an unititialized static variable) */
+    addr = get_base_address(mach_task_self(), &region_size, &in_bss);
+    shmem_internal_data_length = ((void*)addr + region_size) - shmem_internal_data_base;
```
wrrobin, line 318: Confused here, may be misunderstanding. Shouldn't we subtract the `region_size` here?

src/init.c
kholland-intel: I agree with Wasi. I like the `assert`, it might be nice to include a message for the user if the program were to abort here.

kholland-intel (approved):
Overall, these changes look good to me.

src/init.c
davidozog: Here's what the assertion prints when things go wrong:
```
Assertion failed: (kret != KERN_SUCCESS), function get_base_address, file ../../src/init.c, line 72.
```
Do you think that's sufficient?

src/init.c
davidozog: The highest address of the BSS is `addr + region_size`, so to calculate the size of the whole data segment we subtract that highest address from the base address of the data segment (at address 0).  Does that make sense?

Is it confusing because I reuse the `addr` and `region_size` variables?  At this line, `addr` points to the start of the BSS and `region_size` is the size of the BSS.  I can use more variables for clarity if you'd like?

src/init.c
wrrobin: `shmem_internal_assertp` may be used as well.

src/init.c
wrrobin: Thanks @davidozog for the explanation and a bit of experiment helped me to understand here. Looks good.

wrrobin (approved):

src/init.c
davidozog: This is `shmem_internal_assertp`:
```
[0000] Assertion Failed: ../../src/init.c:72: get_base_address
[0000]                   kret != KERN_SUCCESS
```

Preference?  This option does have the PE number...

wrrobin: Looks to have a minimal impact. Leaving it to you. Either way, it is ok.

davidozog: I went with `shmem_internal_assertp` - does it look good?

If so, is the PR ready to merge?

wrrobin: Looks good. Please feel free to merge.

PR #1067, opened 2022-10-03
Adding ifdef to enable manual progress

tmh97:
Adding an #ifdef statement that will set the domain_attr.data_progress value to FI_PROGRESS_MANUAL when --enable-manual-progress is used as a configure option.

Without this #ifdef, domain_attr.data_progress is always set to FI_PROGRESS_AUTO even if --enable-manual-progress is specified.

Signed-off-by: tmh97 <thomas.huber@cornelis.networks.com>


tmh97:
Closing as this fix may only benefit opx!

PR #1069, opened 2022-10-31
Disabling bounce buffering when struct fi_context2 is required

tmh97:
When the struct fi_context2 is required by provider/application, disable bounce buffering.

Signed-off-by: Thomas Huber <thomas.huber@cornelisnetworks.com>

### DIFF
--- a/src/transport_ofi.c
+++ b/src/transport_ofi.c
@@ -1472,7 +1472,7 @@ int shmem_transport_init(void)
 
     /* The current bounce buffering implementation is only compatible with
      * providers that don't require FI_CONTEXT */
-    if (shmem_transport_ofi_info.p_info->mode & FI_CONTEXT) {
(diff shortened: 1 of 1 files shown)


davidozog (approved):
I'm good with this change.  Any issues with merging it, @wrrobin?

wrrobin (commented):
Ok with the change. Thanks @tmh97 for adding that. One minor comment, should we update the debug message (line 1477) printed out and comment before the if block to reflect the change?

tmh97:
> Ok with the change. Thanks @tmh97 for adding that. One minor comment, should we update the debug message (line 1477) printed out and comment before the if block to reflect the change?

Agreed! Just added a new commit that hopefully addresses this. Let me know what you think.

wrrobin (approved):

PR #1078, opened 2022-12-12
Teams: support negative strides, adds a unit test

### LINKED ISSUE #1050: Team split ops do not support negative strides
The v1.5 OpenSHMEM specification does not preclude negative strides in team split operations, but SOS currently does - [right here](https://github.com/Sandia-OpenSHMEM/SOS/blob/v1.5.1/src/shmem_team.c#L259).

This came about while considering "top-down" scan operations here: https://github.com/openshmem-org/specification/pull/488

davidozog:
Fixes #1050.

### DIFF
--- a/src/shmem_internal.h
+++ b/src/shmem_internal.h
@@ -598,7 +598,8 @@ static inline
 int shmem_internal_pe_in_active_set(int global_pe, int PE_start, int PE_stride, int PE_size)
 {
     int n = (global_pe - PE_start) / PE_stride;
-    if (global_pe < PE_start || (global_pe - PE_start) % PE_stride || n >= PE_size)
+    if ((global_pe < PE_start && PE_stride > 0) || (global_pe > PE_start && PE_stride < 0) ||
+        (global_pe - PE_start) % PE_stride || n >= PE_size)
         return -1;
     else {
         return n;
--- a/src/shmem_internal.h
+++ b/src/shmem_internal.h
@@ -597,6 +597,7 @@ void shmem_internal_bit_to_string(char *str, size_t str_size,
 static inline
 int shmem_internal_pe_in_active_set(int global_pe, int PE_start, int PE_stride, int PE_size)
 {
+    shmem_internal_assert(PE_stride != 0);
     int n = (global_pe - PE_start) / PE_stride;
     if ((global_pe < PE_start && PE_stride > 0) || (global_pe > PE_start && PE_stride < 0) ||
         (global_pe - PE_start) % PE_stride || n >= PE_size)
--- a/test/unit/shmem_team_negative_stride.c
+++ b/test/unit/shmem_team_negative_stride.c
@@ -69,14 +69,14 @@ int main(void)
         } else {
             errors++;
         }
-    }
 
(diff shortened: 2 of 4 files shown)


test/unit/shmem_team_negative_stride.c
```diff
@@ -0,0 +1,120 @@
...
+
+        if (new_team_id != -1 && new_team_id == abs(me - npes + 1)) {
+            printf("world team ID = %d, new team ID = %d\n", me, new_team_id);
+        } else {
+            errors++;
+        }
+    }
+
+    shmem_int_sum_reduce(new_team, dst, src, npes);
+
```
wrrobin, line 75: Should we do this reduce and the following check if `new_team == SHMEM_TEAM_INVALID`?

src/shmem_internal.h
```diff
@@ -598,7 +598,8 @@ static inline
 int shmem_internal_pe_in_active_set(int global_pe, int PE_start, int PE_stride, int PE_size)
 {
     int n = (global_pe - PE_start) / PE_stride;
```
wrrobin, line 600: I see two instances of this function being called in SOS. In one instance (call from `team_translate_pe`) this function is called, where PE_stride is the taken from the destination team. Just want to know, if it is guaranteed that any team should not have a PE_stride = 0;

src/shmem_internal.h
```diff
@@ -598,7 +598,8 @@ static inline
...
-    if (global_pe < PE_start || (global_pe - PE_start) % PE_stride || n >= PE_size)
+    if ((global_pe < PE_start && PE_stride > 0) || (global_pe > PE_start && PE_stride < 0) ||
+        (global_pe - PE_start) % PE_stride || n >= PE_size)
```
wrrobin, line 602: If `PE_stride` is negative, then `n` can become negative. Then, the last comparison `n >= PE_size` will never be true, because PE_size cannot be negative. I am trying to come with an example where this causes a problem, but could not. May be this is correct.

test/unit/shmem_team_negative_stride.c
davidozog: Oh jeez, I think so.  Good catch.

src/shmem_internal.h
davidozog: If `PE_stride` is negative, then `n` is only negative if `global_pe` > `PE_start`.  But in that case, `global_pe` would _not_ be in the active set (because `PE_start` has the largest PE value in the set) so `n >= PE_size` is never checked (because the 2nd or'd condition is true).  So I think this is ok.

src/shmem_internal.h
davidozog: Actually, it doesn't look possible for `PE_stride` to be zero here (because in `team_split_strided`, the check already happened and in `team_translate`, the team was already successfully created).  But it doesn't hurt to check with a non-persistent `shmem_internal_assert`.  I will go with that unless you suggest otherwise, @wrrobin.

wrrobin: Thanks @davidozog. Yes, this assert is fine.

wrrobin (approved):

PR #1083, opened 2023-03-01
Update version to v1.5.2rc1 in configure.ac, NEWS file, and simple build script

davidozog:
This updates the version number according to the [Release Checklist](https://github.com/Sandia-OpenSHMEM/SOS/wiki/Release-Checklist) for v1.5.2 release candidate 1.

After it's merged, I'll tag the repository with v1.5.2rc1.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -2,6 +2,25 @@
 
 Sandia OpenSHMEM NEWS -- history of user-visible changes.
 
+v1.5.2rc1
+---------
+
+- Added support for the CXI libfabric provider, enabling SOS on the Slingshot
+  interconnect.  Setup instructions are on SOS's Github wiki.
+- Added support for shmem_team_ptr routine.
+- Added support for negatives strides in the OpenSHMEM teams APIs.
+- Added checks for incorrect buffer overlap when error-checking is enabled.
+- Added a configure option to enable deprecated tests,
+  --enable-deprecated-tests, which is disabled by default.
+- Added a configure option to enable libfabric manual progress,
+  --enable-ofi-manual-progress, which is disabled by default.
+- Added experimental support for a hint to shmem_malloc_with_hints,
+  SHMEMX_MALLOC_NO_BARRIER, which removes the barrier at exit from the routine.
+  Users must synchronize appropriately after such an operation.
+- Initialized the OFI tx/rx capabilities appropriately to limit what is enabled
+  by providers. This resolves an issue the Omni-Path Express (opx) provider.
+- Patched the symmetric data segment initialization to better support MacOS.
+
 v1.5.1
 ------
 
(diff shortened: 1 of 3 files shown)


scripts/simple-build-ofi.sh
```diff
@@ -13,10 +13,10 @@
 set -e
 
 if [ -z "$SOS_VERSION" ] ; then
-    SOS_VERSION="v1.5.0"
+    SOS_VERSION="v1.5.1"
```
wrrobin, line 16: Not sure, but should this be 1.5.2?

NEWS
```diff
@@ -2,6 +2,25 @@
...
+- Added a configure option to enable libfabric manual progress,
+  --enable-ofi-manual-progress, which is disabled by default.
+- Added experimental support for a hint to shmem_malloc_with_hints,
+  SHMEMX_MALLOC_NO_BARRIER, which removes the barrier at exit from the routine.
+  Users must synchronize appropriately after such an operation.
+- Initialized the OFI tx/rx capabilities appropriately to limit what is enabled
+  by providers. This resolves an issue the Omni-Path Express (opx) provider.
+- Patched the symmetric data segment initialization to better support MacOS.
+
 v1.5.1
```
wrrobin, line 24: Some missed bullets, but not sure if they should be added anyway.
- Changes in the LICENSE file
- Unit test for shmem_ctx_get_team
- Disable bounce buffering for fi_context2
- Disabling PMIx tests in CI
- Bug fix and a unit test for in-place broadcast

I am ok with not keeping these in the NEWS. Just thought in case it slipped.

scripts/simple-build-ofi.sh
davidozog: Looking back at our last few releases, I think we keep this 1 minor release number behind the latest.  But the Release Checklist in the wiki could be more clear about this...  Are you ok with that protocol?  I think it's slightly easier to do it this way, then we don't have to worry about updating _after_ we drop the release.

NEWS
davidozog: I tracked everything except the bug fix for in-place broadcast.  Is that #1008, and wasn't that in `v1.5.1`?

The others I'm ok with excluding, are there any you think should be announced?  Seems like most users wouldn't know/care so much about these; but some developers might..

wrrobin: Yes, I am ok. Just thought if you missed them. 

I am actually not seeing in-place broadcast mentioned on the previous release note. If this change went in `1.5.1`, then its ok not to include it in the release notes now.

davidozog: Agreed - we must have missed it, but it might be an important thing to call out.  Do you think we should add it to the NEWS file retroactively?  I'm inclined to.. If so, is that ok on this PR?

wrrobin: I am ok. It might be better to make sure that v1.5.1 has those changes.

davidozog: Looks like it's already there in v1.5.1rc1 - is this it?
```
- Fixed issues with the OpenSHMEM teams API (e.g. team-based broadcasts now                                                                                                                  
  update the destination object on all PEs including the root).
  ```

wrrobin (approved):

kholland-intel (approved):
LGTM.

PR #1086, opened 2023-03-11
Wip/mmap xpmem heap

davidozog:
This new on-node transport mmap's the symmetric data and heap segments of all other PE's, much like XPMEM does.  The intent is to provide an alternative to XPMEM for shared memory transport when XPMEM is not available.  The RMA performance looks very similar to XPMEM.

To create a file-backed mmap of the data segment, I copy the whole segment to the file during initialization.  This may be risky if another thread/entity can somehow update the data segment before the mmap succeeds.  But I haven't seen any problems in practice.  (MacOS does not support this because writing to the file fails for some reason.)

Thanks to @halitdogan and Pardo for the help on this!

### DIFF
--- a/src/transport_mmap.c
+++ b/src/transport_mmap.c
@@ -1,0 +1,274 @@
+/* -*- C -*-
+ *
+ * Copyright 2011 Sandia Corporation. Under the terms of Contract
+ * DE-AC04-94AL85000 with Sandia Corporation, the U.S.  Government
+ * retains certain rights in this software.
+ *
+ * Copyright (c) 2017 Intel Corporation. All rights reserved.
+ * This software is available to you under the BSD license.
+ *
+ * This file is part of the Sandia OpenSHMEM software package. For license
+ * information, see the LICENSE file in the top level directory of the
+ * distribution.
+ *
+ */
+
+#include "config.h"
+
+#include <errno.h>
+#include <string.h>
+#include <sys/types.h>
+#include <unistd.h>
+#include <sys/mman.h>
+#include <fcntl.h>
+
+#define SHMEM_INTERNAL_INCLUDE
+#include "shmem.h"
+#include "shmem_internal.h"
+#include "shmem_comm.h"
+#include "runtime.h"
+#include "transport_mmap.h"
+
+#define MPIDI_OFI_SHMGR_NAME_MAXLEN (128)
+#define MPIDI_OFI_SHMGR_NAME_PREFIX "/sos_shm_colls_area"
+
+static void shm_create_key(char *key, size_t max_size, unsigned pe, size_t num) {
+    struct timeval tv;
+    gettimeofday(&tv, NULL);
+    long long ticks = num;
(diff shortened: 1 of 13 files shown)


src/transport_mmap.c
```diff
@@ -0,0 +1,295 @@
...
+    shm_unlink(key);
+    int fd = shm_open(key, O_RDWR | O_CREAT | O_TRUNC, 0666);
+    if (fd == -1) {
+        fprintf(stderr, "mmap_init data_seg error shm_open with errno(%s)\n", strerror(errno));
+        exit(1);
+    }
+
+    /* Write all current contents of the data segment to the file */
+    FILE *fp = fdopen(fd, "wb");
+    size_t ret = fwrite(base, shm_size, 1, fp);
```
jdinan, line 78: Can you share a reference to docs for this? As far as I've been able to track down, `fdopen` on a descriptor returned by `shm_open` is described as having undefined behavior.

davidozog: I can't find anything referencing this either, probably because it's a complete hack ;).

BTW, there's a good chance we won't merge this PR, at least as it is, for this very reason.  Maybe it should just be a topic branch until the symmetric data segment is deprecated or something.  The reason this WIP PR exists is because I'd like to evaluate some internal systems where there's no chance XPMEM will ever be installed, so we have to think out of the box.

Yes, manpages say `fdopen` on a shared memory object is undefined, but I think that's because there are many operations one may not be able to do on shared memory "files" (like seek, lock via fcntl, chmod, etc.) and possibly because there could be other processes writing to the shared region while the file is being opened.  So it's usually a very bad idea to do this.

But I'm wondering if it may be relatively harmless to do this within SOS init on Linux, because we're not doing any fancy ops on the file (just writing to it and then truncating it) and no other PE or thread will write to the region until after finalize (but I think there's a risk of say some crazy interrupt or something leading to changes to the data segment before it's mmap'd).

I'm surprised this approach always seems to work on my end (based on the unit tests) - but I've only tried a few flavors of Linux.  MacOS definitely doesn't like it at all, I get a failure on `fwrite`.

src/transport_mmap.c
```diff
@@ -0,0 +1,295 @@
...
+static void shm_create_key(char *key, size_t max_size, unsigned pe, size_t num) {
+    snprintf(key, max_size, "%s-%u-%zu", MPIDI_OFI_SHMGR_NAME_PREFIX, pe, num);
+}
+
+
+static void *shm_create_region(char* base, const char *key, int shm_size) {
+  if (shm_size == 0) return NULL;
+
+  shm_unlink(key);
+  int fd = shm_open(key, O_RDWR | O_CREAT | O_TRUNC, 0666);
```
davidozog, line 45: Can permissions be tighter, say 0600?

src/transport_mmap.c
jdinan: Yeah, this is the only mention I could find: https://stackoverflow.com/questions/23347877/map-char-pointer-to-in-memory-file-with-file-descriptor

davidozog:
@markbrown314 - do you think it's possible to use `tmpfs` instead of the `shm_` APIs to avoid some of the tricky issues mentioned above?

PR #1087, opened 2023-03-22
Configury: Add patches to support oneAPI compilers

davidozog:
This fixes a build issue when using newer versions of the oneAPI C compiler (icx), to build SOS.  Support for C11 generics is not correctly detected. 

(@wrrobin @kholland-intel - I think I was wrong when I mentioned there was a problem with fixed-width types and oneAPI compilers. Something like what's in this PR should be all that's needed.)

I didn't work too hard to generally support fortran cause it's deprecated, but I left a note 😄.  It seems to work with `ifort` and `ifx` seems to work with autoconf >= v2.70.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -298,7 +298,15 @@ AC_SUBST([WRAPPER_COMPILER_CC])
 AC_MSG_CHECKING([if -std=gnu11 works])
 
 AC_LANG_PUSH([C])
-CFLAGS="-pedantic-errors -std=gnu11"
+C_COMPILER_VERSION=$("$CC" --version 2> /dev/null)
+IS_ONEAPI_COMPILER=$(echo $C_COMPILER_VERSION | $GREP -c -E 'Intel\(R\) oneAPI')
+
+if test "$IS_ONEAPI_COMPILER" -ne "0" ; then
+  CFLAGS="-std=gnu11"
+else
+  CFLAGS="-pedantic-errors -std=gnu11"
+fi
+
 AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[
                    void g(int i){}
                    #define f(X) _Generic((X), default: g)(X)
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -298,7 +298,15 @@ AC_SUBST([WRAPPER_COMPILER_CC])
 AC_MSG_CHECKING([if -std=gnu11 works])
 
 AC_LANG_PUSH([C])
-CFLAGS="-pedantic-errors -std=gnu11"
+C_COMPILER_VERSION=$("$CC" --version 2> /dev/null)
+IS_ONEAPI_COMPILER=$(echo $C_COMPILER_VERSION | $GREP -c -E 'Intel\(R\) oneAPI')
```
davidozog, line 302: GNU docs suggest using `AC_PROG_EGREP` before $GREP -E, which doesn't seem to be checked before this in configure.ac.

configure.ac
```diff
@@ -516,11 +526,13 @@ fi
 if test "$bt_execinfo" = "yes"; then
   use_bt_option="$use_bt_option execinfo"
   AC_DEFINE([USE_BT_EXECINFO], [1], [Execinfo library calls for backtrace])
+  if test "$IS_ONEAPI_COMPILER" -eq "0" ; then
```
wrrobin, line 529: This is only for `execinfo` because of the `-rdynamic`?

configure.ac
wrrobin: I might not have got the comment. Do we need to add this check then? I am not seeing it added.

davidozog: Oh right, this was a reminder that I did not add the check, but it seems to work without it.

I looked into this briefly and I see that somewhere earlier there is already a check:
`checking for grep that handles long lines and -e... /usr/bin/grep`

I don't know exactly where that check occurs and if we can count on it...

wrrobin: Yes, I remember this also.. and it happens very early in the configure. I think we can go with it. If someone reports a bug, we can add it then.

wrrobin (approved):

Issue #1089, opened 2023-04-04
Receive capabilities may not be a subset of the endpoint capabilities.


tmh97:
As of now, Cornelis's OPX abides by the libfabric manpage which states "The capabilities (rx or tx) must be a subset of those requested of the associated endpoint." 

In the past we address this https://github.com/Sandia-OpenSHMEM/SOS/pull/1080, however I just realized that on line 1393 of src/transport_ofi.c https://github.com/Sandia-OpenSHMEM/SOS/blob/main/src/transport_ofi.c#L1393, we are still violating the lifabric manpages, which forces failure in OPX.

I'm hoping we can resolve this by:
- setting rx_attr->caps to info->p_info->caps
- setting rx_attr->caps to 0 or null (potentially not o.k. for a few other providers)
- any other suggestions that result in rx_attr->caps becoming a subset of the p_info->caps

PR #1091, opened 2023-04-17
Adding FI_RECV flag to ctx init caps

tmh97:
Closes issue #1089


davidozog (approved):
Thanks @tmh97.

@wrrobin - I'm thinking we can merge this to make testing it easier.  If something comes up, we can always revert.  Any issues/thoughts on that?

wrrobin (approved):

PR #1095, opened 2023-08-25
Missing reduction types for C++ templates

wrrobin:


bindings/shmem_bind_cxx.m4
```diff
@@ -91,10 +95,13 @@ $1(double,     double,             `SHM_INTERNAL_DOUBLE')$2
 $1(longdouble, long double,        `SHM_INTERNAL_LONG_DOUBLE')')dnl
 dnl
 define(`SHMEM_BIND_CXX_COLL_SUM_PROD', dnl args: macro_name, op_name, op_const
+SHMEM_BIND_CXX_COLL_SUM_PROD_EXTRAS($1,$2)dnl
 `$1(short,     short,              `SHM_INTERNAL_SHORT')$2
 $1(ushort,     unsigned short,     `SHM_INTERNAL_USHORT')$2
+$1(uchar,     unsigned char,     `SHM_INTERNAL_UCHAR')$2
 $1(int,        int,                `SHM_INTERNAL_INT')$2
 $1(uint,       unsigned int,       `SHM_INTERNAL_UINT')$2
+$1(int8,       int8_t,             `SHM_INTERNAL_INT8')$2
```
davidozog, line 104: Same question about `int8_t` here.

test/unit/cxx_test_shmem_bitwise_reduce.cpp
```diff
@@ -0,0 +1,183 @@
+/*
+ *  This test program is derived from a unit test created by Nick Park.
+ *  The original unit test is a work of the U.S. Government and is not subject
+ *  to copyright protection in the United States.  Foreign copyrights may
+ *  apply.
+ *
+ *  Copyright (c) 2021 Intel Corporation. All rights reserved.
```
davidozog, line 7: 2023?

test/unit/cxx_test_shmem_bitwise_reduce.cpp
```diff
@@ -0,0 +1,183 @@
...
+ *        provided with the distribution.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+ * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS
+ * BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN
+ * ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
```
davidozog, line 30: This doesn't look exactly like BSD-3C.  Shouldn't it mirror the license in the main LICENSE file?

davidozog: Same question/comment for all 3 tests.

bindings/generate_bind_cxx_extras.cpp
```diff
@@ -81,8 +81,51 @@
...
+           COMPARE_TYPES(double, CTYPE)             \
+           COMPARE_TYPES(long double, CTYPE)
 
+/* Note: Mirrors SHMEM_BIND_CXX_COLL_SUM_PROD */
+#define NEED_COLL_SUM_PROD_ASSOC(CTYPE)             \
+           COMPARE_TYPES_FIRST(short, CTYPE)        \
+           COMPARE_TYPES(unsigned short, CTYPE)     \
+           COMPARE_TYPES(unsigned char, CTYPE)      \
+           COMPARE_TYPES(int, CTYPE)                \
+           COMPARE_TYPES(int8_t, CTYPE)             \
```
davidozog, line 120: Is `int8_t` a special case?

bindings/generate_bind_cxx_extras.cpp
```diff
@@ -81,8 +81,51 @@
            COMPARE_TYPES(unsigned short, CTYPE)     \
            COMPARE_TYPES(unsigned int, CTYPE)       \
            COMPARE_TYPES(unsigned long, CTYPE)      \
+           COMPARE_TYPES(unsigned long long, CTYPE)
```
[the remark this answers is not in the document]
wrrobin: For `SYNC`, I have not done anything other than addition of `unsigned long long`. I see why you are asking though.. and I think removing `short` makes sense per spec.

bindings/generate_bind_cxx_extras.cpp
```diff
@@ -81,8 +81,51 @@
...
+#define NEED_COLL_AND_OR_XOR_ASSOC(CTYPE)           \
+           COMPARE_TYPES_FIRST(unsigned char, CTYPE)\
+           COMPARE_TYPES(short, CTYPE)              \
+           COMPARE_TYPES(unsigned short, CTYPE)     \
+           COMPARE_TYPES(int, CTYPE)                \
+           COMPARE_TYPES(unsigned int, CTYPE)       \
+           COMPARE_TYPES(long, CTYPE)               \
+           COMPARE_TYPES(unsigned long, CTYPE)      \
+           COMPARE_TYPES(long long, CTYPE)          \
+           COMPARE_TYPES(unsigned long long, CTYPE)
```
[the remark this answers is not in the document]
wrrobin: Will check.

bindings/generate_bind_cxx_extras.cpp
```diff
@@ -81,8 +81,51 @@
...
+           COMPARE_TYPES(int, CTYPE)                \
+           COMPARE_TYPES(int8_t, CTYPE)             \
+           COMPARE_TYPES(unsigned int, CTYPE)       \
+           COMPARE_TYPES(long, CTYPE)               \
+           COMPARE_TYPES(unsigned long, CTYPE)      \
+           COMPARE_TYPES(long long, CTYPE)          \
            COMPARE_TYPES(unsigned long long, CTYPE) \
+           COMPARE_TYPES(float, CTYPE)              \
+           COMPARE_TYPES(double, CTYPE)             \
+           COMPARE_TYPES(long double, CTYPE)
```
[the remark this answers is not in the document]
wrrobin: Yeah.. it is in weird order throughout. I can unify them with the spec. `size_t` and `signed char` are added as extras below.

bindings/generate_bind_cxx_extras.cpp
wrrobin: Yeah.. I was getting a build error without it, a conflict with signed char I believe..

bindings/shmem_bind_cxx.m4
```diff
@@ -78,10 +79,13 @@ $1(longlong,  long long,          `SHM_INTERNAL_LONG_LONG')$2
 $1(ulonglong, unsigned long long, `SHM_INTERNAL_ULONG_LONG')')dnl
 dnl
 define(`SHMEM_BIND_CXX_COLL_MIN_MAX', dnl args: macro_name, op_name, op_const
+SHMEM_BIND_CXX_COLL_MIN_MAX_EXTRAS($1,$2)dnl
 `$1(short,     short,              `SHM_INTERNAL_SHORT')$2
 $1(ushort,     unsigned short,     `SHM_INTERNAL_USHORT')$2
+$1(uchar,     unsigned char,     `SHM_INTERNAL_UCHAR')$2
 $1(int,        int,                `SHM_INTERNAL_INT')$2
 $1(uint,       unsigned int,       `SHM_INTERNAL_UINT')$2
+$1(int8,       int8_t,             `SHM_INTERNAL_INT8')$2
```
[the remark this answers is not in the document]
wrrobin: Will correct.

bindings/generate_bind_cxx_extras.cpp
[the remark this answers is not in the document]
wrrobin: Thanks @davidozog. I am removing short for sync.. but, for others, I am leaving them. We can fix them in a separate PR from you.

wrrobin:
@davidozog - Thanks for your reviews and catching some mistakes I had. The issue for `int8_t` and `signed char` was that I added them in wrong places. I reversed them. Please check. 
Regarding the copyright, I updated them. But, it seems many test files might have the other one. We might need to do a refresh of the copyright separately.

test/unit/cxx_test_shmem_bitwise_reduce.cpp
davidozog: Ah, I'm not sure what exactly is going on here, but it looks like it's more than just `shmem_team_reduce.c` that is slightly inconsistent with the LICENSE file.  This text looks more like BSD-2C, but even that is not exactly it...

I think we should keep these tests and the previous tests as they are for now (because we're copying most of the original tests), but I will look into (i.e., ask John) whether we can/should change them to the BSD-3C we have in the main license file for SOS v1.5.3 (there's also the small matter that our main LICENSE file is actually variant SPDX:BSD-3C-Open-MPI).

davidozog:
> @davidozog - Thanks for your reviews and catching some mistakes I had. The issue for `int8_t` and `signed char` was that I added them in wrong places. I reversed them. Please check.

Will do!

> Regarding the copyright, I updated them. But, it seems many test files might have the other one. We might need to do a refresh of the copyright in all the test files separately.

I'm thinking we should revert that, and keep the licenses as they were and fix later (there's always some risk in changing a license as you copy/paste the code...).  I'm pretty sure we can fix it later, but I'll double check with John.

test/unit/cxx_test_shmem_bitwise_reduce.cpp
wrrobin: Actually, most unit tests are having a 2-clause statement.

wrrobin:
Ok. I can revert the copyright. 

The previous test that I took the copyright from is introduced here: https://github.com/Sandia-OpenSHMEM/SOS/commit/d1506299bbcde0e19c0917ff5323ecf5be63ecbd. And, it looks like this exact copyright statement exists in other test files as well.. which might have happened because it had been copied from a previous unit test. I am not sure if the first sentence on the copyright is also true for all these tests.

davidozog (approved):
This looks correct to me, thanks Wasi!

`SHMEM_BIND_CXX_COLL_MIN_MAX ` and `SHMEM_BIND_CXX_COLL_SUM_PROD` could maybe use an ordering w.r.t the spec, and the test copyright could be 2023, but that's pretty minor.

test/unit/cxx_test_shmem_bitwise_reduce.cpp
davidozog: Right, but it's not the exact same BSD 2-clause as the SPDX version: https://spdx.org/licenses/BSD-2-Clause.html

wrrobin:
> This looks correct to me, thanks Wasi!
> 
> `SHMEM_BIND_CXX_COLL_MIN_MAX ` and `SHMEM_BIND_CXX_COLL_SUM_PROD` could maybe use an ordering w.r.t the spec, and the test copyrights could be 2023, but that's pretty minor.

Thanks Dave. I am incorporating those two, and going to merge after the tests pass.

PR #1096, opened 2023-08-28
tests: Replace shmemx_wtime() with tests_sos_wtime

davidozog:
This is intended to help users of the https://github.com/openshmem-org/tests-sos repository to build the test suite even when the OpenSHMEM library does not define `shmemx_wtime()` (e.g., OSHMPI).

### DIFF
--- a/test/unit/fadd_nbi.c
+++ b/test/unit/fadd_nbi.c
@@ -29,6 +29,7 @@
 #include <stdio.h>
 #include <shmem.h>
 #include <shmemx.h>
+#include "tests_sos/wtime.h"
 
 long ctr = 0;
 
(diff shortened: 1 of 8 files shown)


test/unit/fadd_nbi.c
```diff
@@ -29,6 +29,7 @@
 #include <stdio.h>
 #include <shmem.h>
 #include <shmemx.h>
```
wrrobin, line 31: We can remove `shmemx.h`? 

Also, is this the only test in the repo that uses `shmemx_wtime()`?

davidozog: > We can remove `shmemx.h`?

Yes, good catch.

> Also, is this the only test in the repo that uses `shmemx_wtime()`?

Oops - I just noticed there are ~4 other tests that do something like this:
```
#ifndef HAVE_SHMEMX_WTIME
static double shmemx_wtime(void) {
    struct timeval tv;
    gettimeofday(&tv, NULL);
    return (double) tv.tv_sec + (double) tv.tv_usec / 1000000.0;
}
#endif /* HAVE_SHMEMX_WTIME */
```

So it seems we can have all these 5 tests use the new header or I can just add this to the `fadd_nbi.c` test.  Any preference @wrrobin?

davidozog:
This is ready for review @kholland-intel and @wrrobin - thanks!

wrrobin (approved):

kholland-intel (approved):
LGTM

Issue #1097, opened 2023-08-28
Deprecated signed integer types (non fixed-width) for bitwise reductions are being tested


davidozog:
We are testing deprecated types for bitwise reductions (`short`, `int`, `long`, and `long long`) in the SOS test suite, but these are deprecated and not supported by other libraries (e.g. OSHMPI).

I think we should remove them from the tests and consider marking the routines as deprecated, if possible.  Removal might be appropriate too, because the operation is possibly ambiguous (given that these types might have either 1's or 2's complement representation).

avincigu:
Closing as it seem to have been addressed in https://github.com/Sandia-OpenSHMEM/SOS/pull/1098

PR #1098, opened 2023-08-28
Remove unsupported signed reductions tests for v1.5

davidozog:
This addresses issue #1097.

Signed integer types that are not fixed-width are deprecated in OpenSHMEM v1.5, so we should remove the corresponding tests for the sake of other libraries using the SOS test suite.

This PR just removes the instances from the tests.  The next step might be to mark the routines as deprecated, or possibly remove them.


kholland-intel (approved):
LGTM

PR #1099, opened 2023-08-29
Bitwise amo cpp bindings: removed fixed width from check

wrrobin:
@davidozog - The change you pointed out made sense. This PR handles it. Please check.


davidozog (approved):
Thanks @wrrobin

PR #1102, opened 2023-10-02
Teams: add SHMEMX_TEAM_NODE

davidozog:
This should be handy for supporting multi-NIC and/or topologically-aware collectives.

@wrrobin - perhaps this could be named `SHMEMX_TEAM_NODE` because it's relying on PMI's notion of a "node" (via `shmem_runtime_get_node_rank()`).  Here is a short summary of where this info comes from across the SOS "`runtime-pmi`" options:

- PMI1: `gethostname`
- PMI2: `gethostname`
- PMIx: `PMIX_LOCAL_PEERS` query
- MPI:  Split of type `MPI_COMM_TYPE_SHARED`

But `SHMEMX_TEAM_HOST` is fine with me if that's preferred/simpler.

### DIFF
--- a/src/shmem_team.c
+++ b/src/shmem_team.c
@@ -20,7 +20,7 @@
 
 #define SHMEM_TEAM_WORLD_INDEX   0
 #define SHMEM_TEAM_SHARED_INDEX  1
-#define SHMEM_TEAM_HOST_INDEX    2
+#define SHMEM_TEAM_NODE_INDEX    2
 #define SHMEM_TEAMS_MIN          3
 
 #define N_PSYNC_BYTES             8
@@ -33,8 +33,8 @@ shmem_team_t SHMEM_TEAM_WORLD = (shmem_team_t) &shmem_internal_team_world;
 shmem_internal_team_t shmem_internal_team_shared;
 shmem_team_t SHMEM_TEAM_SHARED = (shmem_team_t) &shmem_internal_team_shared;
 
-shmem_internal_team_t shmem_internal_team_host;
-shmem_team_t SHMEMX_TEAM_HOST = (shmem_team_t) &shmem_internal_team_host;
+shmem_internal_team_t shmem_internal_team_node;
+shmem_team_t SHMEMX_TEAM_NODE = (shmem_team_t) &shmem_internal_team_node;
 
 shmem_internal_team_t **shmem_internal_team_pool;
 long *shmem_internal_psync_pool;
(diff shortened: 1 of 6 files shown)


philipmarshall21 (approved):
LGTM! Should things like **SHMEM_TEAM_HOST_INDEX** and **shmem_internal_team_host** be **SHMEMX_TEAM_HOST_INDEX**/**shmemx_internal_team_host** though? From a readability perspective I definitely prefer things as they are, just not sure about the technicalities of using shmem vs shmemx. Are we good to use shmem internally for variable name prefixes to shmemx related objects so long as what is exposed to the user abides by the shmem/shmemx rules?

wrrobin:
Thanks @davidozog. `SHMEMX_TEAM_NODE` is better.

davidozog:
> LGTM! Should things like **SHMEM_TEAM_HOST_INDEX** and **shmem_internal_team_host** be **SHMEMX_TEAM_HOST_INDEX**/**shmemx_internal_team_host** though? From a readability perspective I definitely prefer things as they are, just not sure about the technicalities of using shmem vs shmemx. Are we good to use shmem internally for variable name prefixes to shmemx related objects so long as what is exposed to the user abides by the shmem/shmemx rules?

Thanks @philipmarshall21 - only the user-facing API needs the "X" to indicate it's an extension.  The runtime can handle internal symbols however, and as you say, it kinda makes sense to withhold the "X/x" from symbols that aren't visible to users anyway.

davidozog:
> Thanks @davidozog. `SHMEMX_TEAM_NODE` is better.

I think so too - it's renamed now.

src/shmem_team.c
```diff
@@ -140,14 +185,15 @@ int shmem_internal_team_init(void)
...
+    shmem_internal_team_pool[SHMEM_TEAM_NODE_INDEX] = &shmem_internal_team_node;
 
     /* Allocate pSync pool, each with the maximum possible size requirement */
     /* Create two pSyncs per team for back-to-back collectives and one for barriers.
      * Array organization:
      *
-     * [ (world) (shared) (team 1) (team 2) ...  (world) (shared) (team 1) (team 2) ... ]
-     *  <----------- groups 1 & 2-------------->|<------------- group 3 ---------------->
-     *  <--- (bcast, collect, reduce, etc.) --->|<------ (barriers and syncs) ---------->
+     * [ (world) (shared) (host) (team 1) (team 2) ...  (world) (shared) (host) (team 1) (team 2) ... ]
```
wrrobin, line 194: `node`?

test/shmemx/shmemx_team_node.c
```diff
@@ -0,0 +1,54 @@
...
+    printf("[PE: %d] SHMEMX_TEAM_NODE peers: { ", me);
+    for (int i = 0; i < npes; i++) {
+        if (shmem_team_translate_pe(SHMEM_TEAM_WORLD, i,
+                                    SHMEMX_TEAM_NODE) != -1) {
+            peers[num_peers++] = i;
+            printf("%d ", i);
+        }
+    }
+
+    printf("} (num_peers: %d)\n", num_peers);
```
wrrobin, line 41: Just a thought, should we use `fprintf` and direct them to `stdout`? Very minor question..

wrrobin (approved):
LGTM

test/shmemx/shmemx_team_node.c
davidozog: The `fprintf` man page suggests `printf` always directs to `stdout`, so I think it's probably ok.  Also glancing at the SOS unit tests overall it seems we use `fprintf` only when directing to `stderr` and `printf` for everything else, so at least this is consistent.  My opinion is it's effectively the same, so it's fine as long as we're consistent.

src/shmem_team.c
davidozog: done in [c5d1bfe](https://github.com/Sandia-OpenSHMEM/SOS/pull/1102/commits/c5d1bfe72cc5b42404928954010748169e4fe68f)

PR #1103, opened 2023-10-03
Enable GPU RDMA and external heap creation

wrrobin:


wrrobin:
@davidozog - Sorry for my late reply on this. The `shmem_init` is divided in three parts to allow heap creation from the user. The first part is `shmemx_runtime_init`. After the heap creation, the `shmemx_transport_init` takes care of the rest of the initialization stuffs. Let me know if you prefer, `pre_heap_init` and `post_heap_init` instead.

davidozog (commented):
Looks good, thanks Wasi - just a few questions, mostly about nomenclature.

A test or two would be nice, but we can do that in a later PR perhaps.

src/init_c.c
```diff
@@ -72,6 +73,28 @@ shmem_init(void)
...
+
+    ret = shmem_internal_runtime_init(SHMEM_THREAD_SINGLE, &tl_provided);
+    if (ret) abort();
+}
+
+
+void SHMEM_FUNCTION_ATTRIBUTES
+shmemx_transport_init(void)
+{
+    int ret = shmem_internal_heap_postinit();
```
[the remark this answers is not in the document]
wrrobin: Agree.. was also thinking similar signatures. Will update.

src/transport_ofi.c
```diff
@@ -1831,6 +2013,13 @@ int shmem_transport_fini(void)
     OFI_CHECK_ERROR_MSG(ret, "Target data MR close failed (%s)\n", fi_strerror(errno));
 #endif
 
+#ifdef USE_FI_HMEM
+    if (shmem_external_heap_pre_initialized) {
+        ret = fi_close(&shmem_transport_ofi_external_heap_mrfd->fid);
+        OFI_CHECK_ERROR_MSG(ret, "External heap MR close failed (%s)\n", fi_strerror(errno));
```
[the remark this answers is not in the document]
wrrobin: Good catch. Turns out, we are not also freeing `target_heap_keys` and `target_heap_addrs`. Am I missing it?

src/transport_ofi.c
```diff
@@ -622,40 +629,58 @@ int bind_enable_ep_resources(shmem_transport_ctx_t *ctx)
...
 
-        ret = fi_cntr_open(shmem_transport_ofi_domainfd, &cntr_attr,
-                           &shmem_transport_ofi_target_cntrfd, NULL);
-        OFI_CHECK_RETURN_STR(ret, "target CNTR open failed");
+    if (shmem_transport_ofi_info.p_info->domain_attr->mr_mode & FI_MR_ENDPOINT) {
+        ret = fi_ep_bind(shmem_transport_ofi_target_ep,
+                         &shmem_transport_ofi_target_cntrfd->fid, FI_REMOTE_WRITE);
+        OFI_CHECK_RETURN_STR(ret, "target CNTR binding to target EP failed");
+        ret = fi_mr_bind(shmem_transport_ofi_external_heap_mrfd,
+                         &shmem_transport_ofi_target_ep->fid, FI_REMOTE_WRITE);        OFI_CHECK_RETURN_STR(ret, "target EP binding to heap MR failed");
```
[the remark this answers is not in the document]
wrrobin: Huh.. not seeing in my branch. weird.

src/transport_ofi.c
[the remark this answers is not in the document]
davidozog: Ah, right - I'm not seeing that either.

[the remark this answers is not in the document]
davidozog: Just a minor note - I believe the behavior of `free` is to do nothing if the passed pointer is NULL, so the check for non-zero is probably not needed.  I guess it doesn't hurt though.

src/transport_ofi.c
```diff
@@ -622,40 +629,60 @@ int bind_enable_ep_resources(shmem_transport_ctx_t *ctx)
...
-        ret = fi_cntr_open(shmem_transport_ofi_domainfd, &cntr_attr,
-                           &shmem_transport_ofi_target_cntrfd, NULL);
-        OFI_CHECK_RETURN_STR(ret, "target CNTR open failed");
+    if (shmem_transport_ofi_info.p_info->domain_attr->mr_mode & FI_MR_ENDPOINT) {
+        ret = fi_ep_bind(shmem_transport_ofi_target_ep,
+                         &shmem_transport_ofi_target_cntrfd->fid, FI_REMOTE_WRITE);
+        OFI_CHECK_RETURN_STR(ret, "target CNTR binding to target EP failed");
+        ret = fi_mr_bind(shmem_transport_ofi_external_heap_mrfd,
+                         &shmem_transport_ofi_target_ep->fid, FI_REMOTE_WRITE);
+	OFI_CHECK_RETURN_STR(ret, "target EP binding to heap MR failed");
```
davidozog, line 670:
```suggestion
        ret = fi_mr_bind(shmem_transport_ofi_external_heap_mrfd,
                         &shmem_transport_ofi_target_ep->fid, FI_REMOTE_WRITE);
        OFI_CHECK_RETURN_STR(ret, "target EP binding to heap MR failed");
```

davidozog: Looks like it's just a tab thing - mind replacing with spaces @wrrobin?

src/transport_ofi.h
```diff
@@ -177,7 +183,16 @@ void shmem_transport_ofi_get_mr(const void *addr, int dest_pe,
...
+#ifdef USE_FI_HMEM
+    else if (shmem_external_heap_pre_initialized) {
+        if ((void*) addr >= shmem_external_heap_base &&
+             (uint8_t*) addr < (uint8_t*) shmem_external_heap_base + shmem_external_heap_length) {
+            *key = shmem_transport_ofi_external_heap_keys[dest_pe];
+            *mr_addr = shmem_transport_ofi_external_heap_addrs[dest_pe] +
+                ((uint8_t *) addr - (uint8_t *) shmem_external_heap_base);
+        }
+    }
+#endif /* USE_FI_HMEM */
```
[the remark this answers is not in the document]
davidozog: It works?  I'll resolve this thread, I'm just curious if you tested it in the wild.

davidozog (approved):

src/transport_ofi.c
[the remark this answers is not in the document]
wrrobin: Oh.. right. Removing them.

davidozog (approved):

src/transport_ofi.c
```diff
@@ -84,6 +84,13 @@ uint8_t**                       shmem_transport_ofi_target_heap_addrs;
 uint8_t**                       shmem_transport_ofi_target_data_addrs;
 #endif /* ENABLE_REMOTE_VIRTUAL_ADDRESSING */
 #endif /* ENABLE_MR_SCALABLE */
+
+#ifdef USE_FI_HMEM
+struct fid_mr*                  shmem_transport_ofi_external_heap_mrfd;
+uint64_t*                       shmem_transport_ofi_external_heap_keys;
```
philipmarshall21, line 90: Can you explain the concept of keys and addresses? Is a key essentially a password of sorts that is required to perform operations (reads/writes) on a registered memory region?

src/transport_ofi.c
```diff
@@ -622,40 +629,60 @@ int bind_enable_ep_resources(shmem_transport_ctx_t *ctx)
...
+    OFI_CHECK_RETURN_STR(ret, "target CNTR binding to external heap MR failed");
 
-        ret = fi_cntr_open(shmem_transport_ofi_domainfd, &cntr_attr,
-                           &shmem_transport_ofi_target_cntrfd, NULL);
-        OFI_CHECK_RETURN_STR(ret, "target CNTR open failed");
+    if (shmem_transport_ofi_info.p_info->domain_attr->mr_mode & FI_MR_ENDPOINT) {
+        ret = fi_ep_bind(shmem_transport_ofi_target_ep,
+                         &shmem_transport_ofi_target_cntrfd->fid, FI_REMOTE_WRITE);
+        OFI_CHECK_RETURN_STR(ret, "target CNTR binding to target EP failed");
+        ret = fi_mr_bind(shmem_transport_ofi_external_heap_mrfd,
```
philipmarshall21, line 668: Can you explain what the difference between this **mr_bind** call and the one on line 659 is? I understand that the **fid\*** argument is different for each. Is it that, by default, we associate **shmem_transport_ofi_external_heap_mrfd** with **shmem_transport_ofi_target_cntrfd** but overwrite that association in the case that the provider associates memory regions with endpoints (rather than domains)?

What does it mean for a provider to associate a memory region with an endpoint rather than a domain?

src/transport_ofi.c
```diff
@@ -744,6 +771,88 @@ int allocate_recv_cntr_mr(void)
...
+#endif
+    ret = ofi_mr_reg_bind(flags);
+    OFI_CHECK_RETURN_STR(ret, "OFI MR registration failed");
+
+    return ret;
+}
+
+#ifdef USE_FI_HMEM
+static
+int publish_external_mr_info(void)
```
philipmarshall21, line 821: For my understanding, are **publish_external_mr_info** and **publish_mr_info** required to enable PEs to know the heap base (either external or internal) of other PEs?

src/transport_ofi.c
philipmarshall21: Or I guess is it that a memory region can be associated with more than just a single completion counter or endpoint?

src/transport_ofi.c
wrrobin: Yes, `fi_mr_key` is the key that is used to access the memory region that is registered. After the memory registration the keys are exchanged to make sure that everyone has the remote key to access the heap address, which is also exchanged.

src/transport_ofi.c
wrrobin: You can bind endpoint or counter to ensure that any update to the memory region triggers an event on the bound object. Line number 659 is to bind the registered memory region with the target counter to ensure that any update on the memory region using `FI_REMOTE_WRITE` will generate an event to update the target counter. Line number 668 ensures binding with the endpoint when `FI_MR_ENDPOINT` is enabled. This binding is required for providers that requires this flag. From the manpages, 
```
Counter bindings are needed by providers that support the generation of completions based on fabric operations. Endpoint bindings are needed if the provider associates memory regions with endpoints (see FI_MR_ENDPOINT).
```
I am not sure about your last question... if a MR can be associated with multiple counters and/or endpoints. Need to check.

src/transport_ofi.c
wrrobin: Yes.. we publish these through the PMI key value store, IIRC.

philipmarshall21 (approved):

PR #1107, opened 2023-11-29
feature: Add multirail support and optional usage of hwloc for optimal NIC assignment

philipmarshall21:
This PR adds support for utilizing multiple NICs to increase peak RMA bandwidth. Additionally, it adds support for use of hwloc to optimally assign processes to a NIC based on each process' CPU affinity.


src/transport_ofi.c
```diff
@@ -1329,6 +1333,98 @@ int allocate_fabric_resources(struct fabric_info *info)
...
+    size_t num_close_nics = 0;
+    for (size_t i = 0; i < num_nics; i++) {
+        struct fi_info *cur_prov = provs[i];
+        if (cur_prov->nic->bus_attr->bus_type != FI_BUS_PCI) continue;
+
+        struct fi_pci_attr pci = cur_prov->nic->bus_attr->attr.pci;
+        hwloc_obj_t io_device = hwloc_get_pcidev_by_busid(shmem_topology, pci.domain_id, pci.bus_id, pci.device_id, pci.function_id);
+        if (!io_device) RAISE_ERROR_MSG("hwloc_get_pcidev_by_busid failed\n");
+        hwloc_obj_t first_non_io = hwloc_get_non_io_ancestor_obj(shmem_topology, io_device);
+        if (!first_non_io) RAISE_ERROR_MSG("hwloc_get_non_io_ancestor_obj failed\n");
```
davidozog, line 1358: Is it possible to fallback to round-robin if one of these routines return a NULL object?  (If so, I'd think a warning message may be appropriate).

src/transport_ofi.c
```diff
@@ -1329,6 +1333,98 @@ int allocate_fabric_resources(struct fabric_info *info)
...
+            num_close_nics++;
+        }
+    }
+    DEBUG_MSG("Num. NICs w/ affinity to process: %zu\n", num_close_nics);
+
+    if (!close_provs) {
+        RAISE_WARN_MSG("Could not detect any NICs with affinity to the process\n");
+
+        /* If no 'close' NICs, select from list of all NICs using round-robin assignment */
+        //return provs[shmem_team_my_pe(SHMEMX_TEAM_NODE) % num_nics];
```
davidozog, line 1375: I would think the node PE is required here... can it be uncommented?

If so, best to use `shmem_internal_team_node.my_pe` to preserve layering.

philipmarshall21: I was looking into this more recently and I don't think **SHMEMX_TEAM_NODE** can be used - at least not here. The **SHMEMX_TEAM_NODE** team is initialized in **shmem_internal_team_init()**, which is not called until after **shmem_transport_init()** (which calls **query_for_fabric())** in **shmem_internal_heap_postinit()**. This explains why I was seeing that each process was indicating that it was PE 0 in SHMEMX_TEAM_NODE.

For what it's worth, a recent run of the **shmem_bw_put_perf** benchmark seemed to indicate that the round-robin indexing using SHMEM_TEAM_WORLD still worked effectively such that there was no duplicate assignment of PEs to NICs (tested from 1-8 ppn)

I'll go ahead and remove the commented out lines though.

davidozog: Oh yeah, of course...  I completely missed that, thanks.

Do you think it's possible to use this (`my_world_pe % num_nics`) form of round robin for initialization of collectives and teams, then to reassign PEs to NICs based on team-node after that?

There may be some other solution too...  With the code as written, NIC bindings will vary with the total number of PEs... so it could be difficult to understand multi-NIC performance, and seems like it will break if there are differing numbers of NICs per node.

src/shmem_internal.h
```diff
@@ -51,6 +51,11 @@ extern int shmem_external_heap_device;
 
 extern unsigned int shmem_internal_rand_seed;
 
+#ifdef USE_HWLOC
+#include <hwloc.h>
+extern hwloc_topology_t shmem_topology;
```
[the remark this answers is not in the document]
philipmarshall21: Fixed!

src/transport_ofi.c
```diff
@@ -1417,24 +1513,67 @@ int query_for_fabric(struct fabric_info *info)
 
     /* If the user supplied a fabric or domain name, use it to select the
      * fabric.  Otherwise, select the first fabric in the list. */
```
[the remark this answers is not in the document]
philipmarshall21: Fixed!

src/transport_ofi.c
```diff
@@ -1417,24 +1513,67 @@ int query_for_fabric(struct fabric_info *info)
...
+        filtered_fabrics_list_head = info->fabrics;
+    }
+
+    struct fi_info *cur_fabric;
+
+    info->p_info = NULL;
+
+    for (cur_fabric = filtered_fabrics_list_head; cur_fabric; cur_fabric = cur_fabric->next) {
+        if (!fallback) fallback = cur_fabric;
+        if (cur_fabric->nic && !nic_already_used(cur_fabric->nic, multirail_fabric_list_head, num_nics)) {
```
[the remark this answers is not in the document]
philipmarshall21: Added comment and simplified assignment of **info->p_info** in case where **SHMEM_DISABLE_MULTIRAIL** is defined

parkerha1 (approved):
Looks good to me!

src/shmem_env_defs.h
```diff
@@ -107,6 +107,8 @@ SHMEM_INTERNAL_ENV_DEF(OFI_STX_ALLOCATOR, string, "round-robin", SHMEM_INTERNAL_
                        "Algorithm for allocating STX resources to contexts")
 SHMEM_INTERNAL_ENV_DEF(OFI_STX_DISABLE_PRIVATE, bool, false, SHMEM_INTERNAL_ENV_CAT_TRANSPORT,
                        "Disallow private contexts from having exclusive STX access")
+SHMEM_INTERNAL_ENV_DEF(DISABLE_MULTIRAIL, bool, false, SHMEM_INTERNAL_ENV_CAT_TRANSPORT,
+                       "Disable usage of multirail functionality")
```
davidozog, line 111: Let's add this to the README file, noting that it only affects the OFI transport.

In fact, for now we should probably call it `OFI_DISABLE_MULTIRAIL`.

src/transport_ofi.c
```diff
@@ -1420,34 +1517,82 @@ int query_for_fabric(struct fabric_info *info)
...
+        }
+    }
     if (NULL == info->p_info) {
         RAISE_WARN_MSG("OFI transport, no valid fabric (prov=%s, fabric=%s, domain=%s)\n",
                        info->prov_name != NULL ? info->prov_name : "<auto>",
                        info->fabric_name != NULL ? info->fabric_name : "<auto>",
                        info->domain_name != NULL ? info->domain_name : "<auto>");
         return ret;
     }
+    DEBUG_MSG("provider: %s\n", info->p_info->domain_attr->name);
```
davidozog, line 1595: I believe this is included in the `DEBUG_MSG` below, can we remove it here?

src/transport_ofi.c
```diff
@@ -1420,34 +1517,82 @@ int query_for_fabric(struct fabric_info *info)
...
+            if (!fallback) fallback = cur_fabric;
+            if (cur_fabric->nic && !nic_already_used(cur_fabric->nic, multirail_fabric_list_head, num_nics)) {
+                num_nics += 1;
+                if (!multirail_fabric_list_head) multirail_fabric_list_head = cur_fabric;
+                if (multirail_fabric_last_added) multirail_fabric_last_added->next = cur_fabric;
+                multirail_fabric_last_added = cur_fabric;
+            }
+        }
 
+        DEBUG_MSG("Total num. NICs detected: %d\n", num_nics);
```
davidozog, line 1568: Can this move to the aggregated "OFI provider:" output below?  [here I think](https://github.com/Sandia-OpenSHMEM/SOS/pull/1107/files#diff-3cac951d73790a8a9ce2e3647758bc66cd9fc5d4c3243173fee89846df3d2ce3L1487)

src/transport_ofi.c
```diff
@@ -1420,34 +1517,82 @@ int query_for_fabric(struct fabric_info *info)
...
+            struct fi_info **prov_list = (struct fi_info **) malloc(num_nics * sizeof(struct fi_info *));
+            for (cur_fabric = multirail_fabric_list_head; cur_fabric; cur_fabric = cur_fabric->next) {
+                prov_list[idx++] = cur_fabric;
+            }
+            qsort(prov_list, num_nics, sizeof(struct fi_info *), compare_nic_names);
+#ifdef USE_HWLOC
+            info->p_info = assign_nic_with_hwloc(info->p_info, prov_list, num_nics);
+#else
+            /* Round-robin assignment of NICs to PEs */
+            info->p_info = prov_list[shmem_internal_my_pe % num_nics];
```
davidozog, line 1583: This might cause over-subscription in some cases, right?  Perhaps a FIXME or a github issue is needed to remind us that this would be better as `SHMEM_TEAM_NODE->my_pe % num_nics` if possible.  But it's too early in initialization to do that here...

src/shmem_env_defs.h
philipmarshall21: Ok! If we change DISABLE_MULTIRAIL to OFI_DISABLE_MULTIRAIL is it necessary to explicitly mention in the README that it only affects the OFI transport?

src/transport_ofi.c
philipmarshall21: Added FIXME comment

davidozog (approved):

PR #1108, opened 2024-01-22
Add initial hwloc support

philipmarshall21:
This PR is a subset of the changes PR #1107 and is intended to isolate the integration of hwloc as an optional dependency to SOS from the addition of multi-NIC functionality

The changes to the **oac_\*** files reflect the current status of the **main** branch (at the time of this PR, that is commit `c1cfc910d92af43f8c27807a9a84c9c13f4fbc65`) as [the upstream repo](https://github.com/open-mpi/oac) does not have any tags.

The changes to the **opal_\*** files align with the **v5.0.1** tag of [the upstream repo](https://github.com/open-mpi/ompi).


config/opal_setup_wrappers.m4
```diff
@@ -41,8 +42,12 @@ AC_DEFUN([OPAL_WRAPPER_FLAGS_ADD], [
           [$1], [CXXFLAGS], [OPAL_FLAGS_APPEND_UNIQ([wrapper_extra_cxxflags], [$2])],
           [$1], [FCFLAGS], [OPAL_FLAGS_APPEND_UNIQ([wrapper_extra_fcflags], [$2])],
           [$1], [LDFLAGS], [OPAL_FLAGS_APPEND_UNIQ([wrapper_extra_ldflags], [$2])],
-          [$1], [LIBS], [OPAL_FLAGS_APPEND_UNIQ([wrapper_extra_libs], [$2])],
+          [$1], [STATIC_LDFLAGS], [OPAL_FLAGS_APPEND_UNIQ([wrapper_extra_static_ldflags], [$2])],
+          [$1], [LIBS], [OPAL_FLAGS_APPEND_MOVE([wrapper_extra_libs], [$2])],
+          [$1], [STATIC_LIBS], [OPAL_FLAGS_APPEND_UNIQ([wrapper_extra_static_libs], [$2])],
+          [$1], [PC_MODULES], [OPAL_APPEND_UNIQ([wrapper_extra_pkgconfig_modules], [$2])],
```
wrrobin, line 48: Are these changes because of the update or hwloc specific? Just want to make sure that we are ok with the upstream.

philipmarshall21: The changes to this file are to update it to match the current state of the upstream file (so no additional changes beyond pulling in contents of upstream).

philipmarshall21:
> Are the upstream OPAL changes taken from a release or a particular commit of OMPI? Either way, perhaps we should leave a note on this PR in case it ends up being helpful later.

The changes to the **oac_\*** files reflect the current status of the **main** branch (currently, that is commit `c1cfc910d92af43f8c27807a9a84c9c13f4fbc65`) as [the upstream repo](https://github.com/open-mpi/oac) does not have any tags.

The changes to the **opal_\*** files align with the **v5.0.1** tag of [the upstream repo](https://github.com/open-mpi/ompi).

configure.ac
```diff
@@ -121,6 +133,35 @@ AS_IF([test "$enable_threads" != "no"], [
        ])
 AM_CONDITIONAL([HAVE_PTHREADS], [test "$HAVE_POSIX_THREADS" = "1"])
 
+AS_IF([test "$with_hwloc" != "no"], [
```
wrrobin, line 136: Does it also work with a custom path given to `--with-hwloc`?

wrrobin (approved):
LGTM. We can explore other use-cases for hwloc later.

src/init.c
```diff
@@ -378,7 +383,52 @@ shmem_internal_heap_postinit(void)
...
+    hwloc_bitmap_and(bindset_covering_obj, bindset_all, covering_obj->cpuset);
+    hwloc_set_proc_cpubind(shmem_internal_topology, getpid(), bindset_covering_obj, HWLOC_CPUBIND_PROCESS); //Include HWLOC_CPUBIND_STRICT in flags?
+
+    hwloc_bitmap_free(bindset);
+    hwloc_bitmap_free(bindset_all);
+    hwloc_bitmap_free(bindset_covering_obj);
+#endif // HWLOC_ENFORCE_SINGLE_SOCKET || HWLOC_ENFORCE_SINGLE_NUMA_NODE
+#endif // USE_HWLOC
+
+if (shmem_internal_params.DEBUG) {
```
wrrobin, line 431: Minor: the indentation is wrong here.

philipmarshall21: Fixed. Thanks for catching that!

src/init.c
```diff
@@ -378,6 +383,51 @@ shmem_internal_heap_postinit(void)
...
 
 #ifdef HAVE_SCHED_GETAFFINITY
+#ifdef USE_HWLOC
+    ret = hwloc_topology_init(&shmem_internal_topology);
+    if (ret < 0) {
+        RETURN_ERROR_MSG("hwloc_topology_init failed (%s)\n", strerror(errno));
+    }
+
+    ret = hwloc_topology_set_io_types_filter(shmem_internal_topology, HWLOC_TYPE_FILTER_KEEP_ALL);
+	if (ret < 0) {
```
davidozog, line 393: small indent mistake here

src/init.c
```diff
@@ -378,6 +383,51 @@ shmem_internal_heap_postinit(void)
               shmem_internal_data_base, shmem_internal_data_length);
 
 #ifdef HAVE_SCHED_GETAFFINITY
+#ifdef USE_HWLOC
+    ret = hwloc_topology_init(&shmem_internal_topology);
+    if (ret < 0) {
```
davidozog, line 388: Should we also guard hwloc calls against `ret > 0`?

src/init.c
```diff
@@ -378,6 +383,51 @@ shmem_internal_heap_postinit(void)
...
+    }
+#ifdef HWLOC_ENFORCE_SINGLE_SOCKET
+    hwloc_obj_t covering_obj = hwloc_get_next_obj_covering_cpuset_by_type(shmem_internal_topology, bindset, HWLOC_OBJ_PACKAGE, NULL);
+    if (!covering_obj) RETURN_ERROR_MSG("hwloc_get_next_obj_covering_cpuset_by_type failed (could not detect object of type 'HWLOC_OBJ_PACKAGE' in provided cpuset)\n");
+#else //HWLOC_ENFORCE_SINGLE_NUMA_NODE
+    hwloc_obj_t covering_obj = hwloc_get_next_obj_covering_cpuset_by_type(shmem_internal_topology, bindset, HWLOC_OBJ_NUMANODE, NULL);
+    if (!covering_obj) RETURN_ERROR_MSG("hwloc_get_next_obj_covering_cpuset_by_type failed (could not detect object of type 'HWLOC_OBJ_NUMANODE' in provided cpuset)\n");
+#endif
+    hwloc_bitmap_and(bindset_covering_obj, bindset_all, covering_obj->cpuset);
+    hwloc_set_proc_cpubind(shmem_internal_topology, getpid(), bindset_covering_obj, HWLOC_CPUBIND_PROCESS); //Include HWLOC_CPUBIND_STRICT in flags?
```
davidozog, line 423: Check return value?

davidozog: Good question about `HWLOC_CPUBIND_STRICT`.  I'm ok leaving it out with the comment for now, though my guess is that for most HPC use-cases the flag would be appropriate (i.e., do the assignment regardless if the CPU(s) is busy at init time).

configure.ac
```diff
@@ -74,6 +76,16 @@ AC_ARG_ENABLE([completion-polling],
     [AC_HELP_STRING([--enable-completion-polling],
                     [Enable polling in quiet, fence, and local completion operations (default:disabled)])])
 
+AC_ARG_ENABLE([hwloc-enforce-single-socket],
+    [AC_HELP_STRING([--enable-hwloc-enforce-single-socket],
+                    [Repin process if assigned affinity to cores across multiple sockets (default:disabled)])])
+AS_IF([test "$enable_hwloc_enforce_single_socket" = "yes"], [AC_DEFINE([HWLOC_ENFORCE_SINGLE_SOCKET], [1], [Re-assign CPU affinity on processes with affinity to cores on multiple sockets])])
```
davidozog, line 82: Seems like hwloc moved from the term "socket" to "package" a while back.  Are we sure we want "socket"?

(I'm good with it...)

configure.ac
```diff
@@ -74,6 +76,16 @@ AC_ARG_ENABLE([completion-polling],
     [AC_HELP_STRING([--enable-completion-polling],
                     [Enable polling in quiet, fence, and local completion operations (default:disabled)])])
 
+AC_ARG_ENABLE([hwloc-enforce-single-socket],
+    [AC_HELP_STRING([--enable-hwloc-enforce-single-socket],
+                    [Repin process if assigned affinity to cores across multiple sockets (default:disabled)])])
```
davidozog, line 81: Slightly confusing wording... perhaps something like this?:
```suggestion
                    [Repin process to ensure it has affinity to cores within a single socket (default:disabled)])])
```
That might be more correct too... If this option is enabled, do we check for funky pinning first?  My understanding is that the code "repins" regardless of the initial pinning.

configure.ac
```diff
@@ -74,6 +76,16 @@ AC_ARG_ENABLE([completion-polling],
...
+AS_IF([test "$enable_hwloc_enforce_single_socket" = "yes"], [AC_DEFINE([HWLOC_ENFORCE_SINGLE_SOCKET], [1], [Re-assign CPU affinity on processes with affinity to cores on multiple sockets])])
+
+AC_ARG_ENABLE([hwloc-enforce-single-numa-node],
+    [AC_HELP_STRING([--enable-hwloc-enforce-single-numa-node],
+                    [Repin process if assigned affinity to cores across multiple NUMA nodes (default:disabled)])])
```
davidozog, line 86: Similar issue as above

src/init.c
```diff
@@ -378,6 +383,51 @@ shmem_internal_heap_postinit(void)
...
+    ret = hwloc_topology_set_io_types_filter(shmem_internal_topology, HWLOC_TYPE_FILTER_KEEP_ALL);
+	if (ret < 0) {
+        RETURN_ERROR_MSG("hwloc_topology_set_io_types_filter failed (%s)\n", strerror(errno));
+    }
+
+    ret = hwloc_topology_load(shmem_internal_topology);
+    if (ret < 0) {
+        RETURN_ERROR_MSG("hwloc_topology_load failed (%s)\n", strerror(errno));
+    }
+#if defined HWLOC_ENFORCE_SINGLE_SOCKET || defined HWLOC_ENFORCE_SINGLE_NUMA_NODE
```
davidozog, line 401: Style thing: SOS always puts parentheses around args to `defined`. (But I don't think it matters in this case...?)
```suggestion
#if defined(HWLOC_ENFORCE_SINGLE_SOCKET) || defined(HWLOC_ENFORCE_SINGLE_NUMA_NODE)
```

davidozog:
Side note: The UCX row with pmi-mpi looks slow in testing, but I think we can pretty safely ignore that...

src/init.c
philipmarshall21: I'm not sure that it's necessary - according to the hwloc documentation, the return value of the hwloc API calls being made will either be 0 (success) or -1 (error)

configure.ac
philipmarshall21: That's a good question. Personally, I think "socket" might be more intuitive and I may have a slight preference toward it, but I understand it's not exactly consistent with the terminology being used by hwloc

configure.ac
philipmarshall21: That is correct - repinning will occur regardless of initial pinning if SOS has been configured with single socket or single NUMA node options

src/init.c
davidozog: Yeah I saw that... I suppose I prefer the tighter check because any ret > 0 is undefined hwloc behavior (which would pass these checks).  Maybe I'm being pedantic - in practice it should be fine.

configure.ac
philipmarshall21: It does. I've checked on Borealis that passing a custom path (in this case `/home/philipma/multirail/hwloc/install`) to `--with-hwloc` works. Along with the expected text being output during the configure phase indicating that hwloc is being used from the provided path, a call to `ldd ./test/unit/hello` shows that **libhwloc** is being sourced from the provided path

image

src/init.c
davidozog: @philipmarshall21 - looks like this PR is all set except maybe this item.  Should we check the return value of `hwloc_set_proc_cpubind`?  It looks like the docs are a little rough here overall, but this suggests it's 0 or -1 like many of the other routines:
https://hwloc.readthedocs.io/en/stable/group__hwlocality__cpubinding.html

philipmarshall21:
Two quick questions @wrrobin @davidozog:

1. With the addition of hwloc support, should we expect to see the rpath for hwloc when doing something like `oshcc -show` (rpath for libfabric and libsma are being set correctly)? If so, this is currently not the case, and I'll need to look into why this is not happening.
2. In the changes to **init.c** that add hwloc API calls, should the application exit if the calls fail? Right now an ERROR message is printed but the application continues running.

philipmarshall21:
> > Two quick questions @wrrobin @davidozog:
> > 
> > 1. With the addition of hwloc support, should we expect to see the rpath for hwloc when doing something like `oshcc -show` (rpath for libfabric and libsma are being set correctly)? If so, this is currently not the case, and I'll need to look into why this is not happening.
> > 2. In the changes to **init.c** that add hwloc API calls, should the application exit if the calls fail? Right now an ERROR message is printed but the application continues running.
> 
> For 1. I would guess we should see the rpath, considering your changes in the PR. Are you using any other flags that might skip the configury code? For 2. my thought is we should not terminate. If one of the hwloc APIs fail or return error, it should print a warning that hwloc is not enabled for some reason. But, we should let the application proceed. Do other runtimes behave the same?

Ok, I've made the suggested changes (ensure rpath for hwloc is set in **configure.ac**, failed hwloc API calls now send warning message rather than error).

davidozog (commented):
@philipmarshall21 - I ran into a couple minor things but pending those, I think this PR is ready to merge.  All the RPATH stuff is working on my end, thanks for patching that.

davidozog (approved):

PR #1109, opened 2024-01-31
Signal set add APIs

wrrobin:
This PR adds the APIs proposed in https://github.com/openshmem-org/specification/pull/489


davidozog (commented):
Looks great, but maybe the tests should go into `test/shmemx`?  We've recently organized the tests in that manner to better support other libraries that might not have newer APIs like this.

Also, I wonder if it would be easier to handle this change after our upcoming "`tests-sos` as a submodule" change in #1106... perhaps we can discuss offline.

test/shmemx/signal_set.c
```diff
@@ -0,0 +1,91 @@
...
+        shmem_wait_until(&sig_addr, SHMEM_CMP_NE, 0);
+#else
+        shmem_uint64_wait_until(&sig_addr, SHMEM_CMP_NE, 0);
+#endif
+        sig_value = shmem_signal_fetch(&sig_addr);
+    }
+
+    for (i = 0; i < MSG_SZ; i++) {
+        if (target[i] != source[i]) {
+            fprintf(stderr, "%10d: target[%d] = %ld not matching %ld with SHMEM_SIGNAL_ADD\n",
```
rithwiktom, line 81: signal set here.

src/data_c.c4
```diff
@@ -700,6 +708,52 @@ shmem_signal_fetch(const uint64_t* sig_addr)
     return val;
 }
 
+void SHMEM_FUNCTION_ATTRIBUTES
+shmemx_signal_add(uint64_t *sig_addr, uint64_t signal, int pe)
```
philipmarshall21, line 712: Should sig_addr be **const**? I know [in the PR where this was added in the spec](https://github.com/openshmem-org/specification/pull/489/files) the change was suggested to remove the **const** modifier from **sig_addr**, but it looks like [in the main branch of the OpenSHMEM spec repo](https://github.com/openshmem-org/specification/blob/master/content/shmem_signal_add.tex) it is still a **const** parameter

src/data_c.c4
```diff
@@ -700,6 +708,52 @@ shmem_signal_fetch(const uint64_t* sig_addr)
...
+    SHMEM_ERR_CHECK_PE(pe);
+    SHMEM_ERR_CHECK_CTX(ctx);
+    SHMEM_ERR_CHECK_SYMMETRIC(sig_addr, sizeof(uint64_t));
+
+    shmem_internal_atomic(ctx, sig_addr, &signal, sizeof(uint64_t),
+                          pe, SHM_INTERNAL_SUM, SHM_INTERNAL_UINT64);
+}
+
+void SHMEM_FUNCTION_ATTRIBUTES
+shmemx_signal_set(uint64_t *sig_addr, uint64_t signal, int pe)
```
philipmarshall21, line 735: Same question about **const**-ness of **sig_addr**

wrrobin:
Stale. Closing and opening a new one.

PR #1110, opened 2024-02-13
Disabling put_scalar when disable-ofi-inject is used

wrrobin:


davidozog:
@wrrobin - We can ignore the UCX over pmi-mpi failure, I'm removing it in #1106.  Do you think we should add at least one test with `--disable-ofi-inject` here?

wrrobin:
Thanks @davidozog. You are right about the diffs. I will add a test in CI. We can merge this after #1106 is merged.

philipmarshall21 (approved):

davidozog:
This looks good to merge when ready @wrrobin

wrrobin:
Yeah.. I thought about it too. And, initially, I made changes only on the transport layer for a particular function, but it requires more changes. 
If it is ok, lets move forward with this. I am keeping a issue created for this and will handle separately.

PR #1113, opened 2024-02-26
src: Fix bug in SHMEM_DEF_TEST_ALL

philipmarshall21:
Currently the **shmem_test_all** API does not account for elements that are suppose to be ignored (as set by the **status** argument) when determining the return value.


davidozog (commented):
Very good catch @philipmarshall21!  This is my bone-headed mistake I think.

Can you please check `shmem_test_all_vector` I think this bug is there as well.

davidozog (approved):
Thanks @philipmarshall21

wrrobin (approved):
@philipmarshall21 - This one looks good. Please feel free to merge. 

Regarding changes to the other APIs, I am not quite sure if that is applicable. Lets defer that to a later PR any way.

PR #1114, opened 2024-03-04
Add ibput and ibget as extension APIs

philipmarshall21:
Adding implementations for IBPUT/IBGET APIs that will be part of a future OpenSHMEM specification. See specification details regarding these functions [here](https://github.com/openshmem-org/specification/pull/448).

Initially, these functions will be added as extension APIs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -156,6 +156,7 @@ jobs:
           echo "OS_NAME=$(lsb_release -si)-$(ls_release -sr)" >> $GITHUB_ENV
       - uses: actions/checkout@v4
         with:
+          fetch-depth: 0
           submodules: 'true'
       - name: Install dependencies
         run: |
@@ -217,6 +218,7 @@ jobs:
       # SOS
       - name: Build SOS (${{ matrix.sos_config }})
         run: |
+          git submodule update --remote
           ./autogen.sh
           mkdir build; cd build
           ../configure --prefix=${SOS_INSTALL_DIR} --with-ofi=${LIBFABRIC_INSTALL_DIR} ${{ matrix.sos_config }}
@@ -489,6 +491,7 @@ jobs:
           echo "OS_NAME=$(lsb_release -si)-$(ls_release -sr)" >> $GITHUB_ENV
       - uses: actions/checkout@v4
         with:
+          fetch-depth: 0
           submodules: 'true'
       - name: Install dependencies
         run: |
@@ -541,6 +544,7 @@ jobs:
       # SOS
       - name: Build SOS (${{ matrix.sos_config }})
         run: |
+          git submodule update --remote
           ./autogen.sh
           mkdir build; cd build
           ../configure --prefix=${SOS_INSTALL_DIR} --with-ucx=${UCX_INSTALL_DIR} ${{ matrix.sos_config }}
@@ -581,6 +585,7 @@ jobs:
           echo "OS_NAME=$(lsb_release -si)-$(ls_release -sr)" >> $GITHUB_ENV
       - uses: actions/checkout@v4
         with:
+          fetch-depth: 0
           submodules: 'true'
       - name: Install dependencies
         run: |
@@ -633,6 +638,7 @@ jobs:
       # SOS
       - name: Build SOS (${{ matrix.name }})
         run: |
+          git submodule update --remote
           ./autogen.sh
           mkdir build; cd build
           ../configure --prefix=${SOS_INSTALL_DIR} --with-portals4=${PORTALS4_INSTALL_DIR} ${{ matrix.sos_config }}
@@ -660,6 +666,7 @@ jobs:
           echo "OS_NAME=$(lsb_release -si)-$(ls_release -sr)" >> $GITHUB_ENV
       - uses: actions/checkout@v4
         with:
+          fetch-depth: 0
           submodules: 'true'
       - name: Install dependencies
         run: |
@@ -688,6 +695,7 @@ jobs:
       # SOS
       - name: Build SOS (${{ matrix.name }})
         run: |
+          git submodule update --remote
           ./autogen.sh
           mkdir build; cd build
           ../configure --prefix=${SOS_INSTALL_DIR} ${{ matrix.sos_config }}
--- a/src/data_c.c4
+++ b/src/data_c.c4
@@ -481,7 +481,7 @@ SHMEM_PROF_DEF_CTX_PUT_N_SIGNAL_NBI(`mem')
 
 #define SHMEM_DEF_IBPUT(STYPE,TYPE)                           \
   void SHMEM_FUNCTION_ATTRIBUTES                              \
-  SHMEM_FUNC_PROTOTYPE(STYPE##_ibput, TYPE *target,            \
+  SHMEMX_FUNC_PROTOTYPE(STYPE##_ibput, TYPE *target,            \
                        const TYPE *source, ptrdiff_t tst,     \
                        ptrdiff_t sst, size_t bsize, size_t nblocks, int pe)  \
     SHMEM_ERR_CHECK_INITIALIZED();                            \
@@ -531,7 +531,7 @@ SHMEM_PROF_DEF_CTX_PUT_N_SIGNAL_NBI(`mem')
 
 #define SHMEM_DEF_IBPUT_N(NAME,SIZE)                          \
   void SHMEM_FUNCTION_ATTRIBUTES                             \
-  SHMEM_FUNC_PROTOTYPE(ibput##NAME, void *target,             \
+  SHMEMX_FUNC_PROTOTYPE(ibput##NAME, void *target,             \
                        const void *source, ptrdiff_t tst,    \
                        ptrdiff_t sst, size_t bsize, size_t nblocks, int pe) \
     SHMEM_ERR_CHECK_INITIALIZED();                           \
@@ -585,7 +585,7 @@ SHMEM_PROF_DEF_CTX_PUT_N_SIGNAL_NBI(`mem')
 
 #define SHMEM_DEF_IBGET(STYPE,TYPE)                            \
   void SHMEM_FUNCTION_ATTRIBUTES                              \
-  SHMEM_FUNC_PROTOTYPE(STYPE##_ibget, TYPE *target,            \
+  SHMEMX_FUNC_PROTOTYPE(STYPE##_ibget, TYPE *target,            \
                        const TYPE *source,                    \
                        ptrdiff_t tst, ptrdiff_t sst,          \
                        size_t bsize, size_t nblocks, int pe)                 \
@@ -638,7 +638,7 @@ SHMEM_PROF_DEF_CTX_PUT_N_SIGNAL_NBI(`mem')
 
 #define SHMEM_DEF_IBGET_N(NAME,SIZE)                      \
   void SHMEM_FUNCTION_ATTRIBUTES                          \
-  SHMEM_FUNC_PROTOTYPE(ibget##NAME, void *target,          \
+  SHMEMX_FUNC_PROTOTYPE(ibget##NAME, void *target,          \
                        const void *source, ptrdiff_t tst, \
                        ptrdiff_t sst, size_t bsize,       \
                        size_t nblocks, int pe)            \
--- a/src/data_c.c4
+++ b/src/data_c.c4
@@ -485,6 +485,7 @@ SHMEM_PROF_DEF_CTX_PUT_N_SIGNAL_NBI(`mem')
                        const TYPE *source, ptrdiff_t tst,     \
                        ptrdiff_t sst, size_t bsize, size_t nblocks, int pe)  \
     if (bsize == 0) return;                                   \
+    long completion = 0;                                      \
     SHMEM_ERR_CHECK_INITIALIZED();                            \
     SHMEM_ERR_CHECK_PE(pe);                                   \
     SHMEM_ERR_CHECK_CTX(ctx);                                 \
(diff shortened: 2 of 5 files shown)


src/data_c.c4
```diff
@@ -439,6 +479,33 @@ SHMEM_PROF_DEF_CTX_PUT_N_SIGNAL_NBI(`mem')
...
+    SHMEM_ERR_CHECK_STRIDE_GTE_BSIZE(tst, bsize);             \
+    SHMEM_ERR_CHECK_STRIDE_GTE_BSIZE(sst, bsize);             \
+    SHMEM_ERR_CHECK_OVERLAP(target, source,                   \
+                   sizeof(TYPE) * ((nblocks-1) * tst + 1),     \
+                   sizeof(TYPE) * ((nblocks-1) * sst + 1), 0); \
+    size_t i = 0;                                             \
+    for ( ; nblocks > 0 ; --nblocks) {                        \
+      for (i = 0; i < bsize; i++) {                           \
+        shmem_internal_put_scalar(ctx, target++, source++,    \
+                                  sizeof(TYPE), pe);          \
```
davidozog, line 503: @philipmarshall21 - I understood this API to chunk puts into `bsize` segments, is that wrong?

philipmarshall21: My understanding is that the API puts `nblocks` segments, each of size `bsize`. I think the **shmem_internal_put_scalar** can be replaced with something similar to what is **SHMEM_DEF_PUT** (except, `nelems` would be `bsize`; not sure if that is what you are hinting at):
image


Although, this would require calling **shmem_internal_put_wait** `nblocks` times, which may be unavoidable.

src/data_c.c4
```diff
@@ -439,6 +479,33 @@ SHMEM_PROF_DEF_CTX_PUT_N_SIGNAL_NBI(`mem')
...
+                       ptrdiff_t sst, size_t bsize, size_t nblocks, int pe)  \
+    SHMEM_ERR_CHECK_INITIALIZED();                            \
+    SHMEM_ERR_CHECK_PE(pe);                                   \
+    SHMEM_ERR_CHECK_CTX(ctx);                                 \
+    SHMEM_ERR_CHECK_POSITIVE(tst);                            \
+    SHMEM_ERR_CHECK_POSITIVE(sst);                            \
+    SHMEM_ERR_CHECK_SYMMETRIC(target, sizeof(TYPE) * ((nblocks-1) * tst + 1)); \
+    SHMEM_ERR_CHECK_NULL(source, nblocks);                    \
+    SHMEM_ERR_CHECK_STRIDE_GTE_BSIZE(tst, bsize);             \
+    SHMEM_ERR_CHECK_STRIDE_GTE_BSIZE(sst, bsize);             \
```
wrrobin, line 495: Should we also check for `bsize >= 0`? It needs to be, because otherwise the `nblocks` loop below will run needlessly. May be, we should add that in the spec as well. 
If we add that check then line num 490 and 491 becomes redundant.

src/data_c.c4
wrrobin, line 503: It took me a while to get it.. but, curious to know, why didn't you write it in the following way? What am I missing?
```
    for ( ; nblocks > 0 ; --nblocks) {                        \
      for (i = 0; i < bsize; i++) {                           \
        shmem_internal_put_scalar(ctx, target[i], source[i],    \
                                  sizeof(TYPE), pe);          \
      }                                                       \
      target += tst;                                  \
      source += sst;                                  \
    }                                                         \
  }
```

philipmarshall21: From what I can tell, the two implementations are functionally the same, although, I agree that the approach you've provided is easier to understand, so I'll go ahead and make the change here and in the other applicable functions

src/data_c.c4
philipmarshall21: That's a good idea. I'll add that check for bsize and remove the checks that then become redundant. bsize should be strictly greater than 0, right?

philipmarshall21: I guess you can provide a value of 0 to bsize, but then isn't the call essentially a no-op?

wrrobin: For error checking, you can add the check if `bsize > 0`. But, not sure if we should generate an error for `bsize = 0`. May be we can have that check at the beginning and skip all.

src/data_c.c4
philipmarshall21: Went ahead and changed the implementation so the inner for-loop that runs for **bsize** iterations is no longer needed. Instead, for each iteration in the **nblocks** loop, the implementation now does a single put or get of size `bsize * sizeof(TYPE)`

src/data_c.c4
davidozog: I think it might be ok to accept passing `bsize == 0`...  SOS never returns from the other RMA routines when `nelems` is zero, right?  I think our strategy has been to pass zero-byte messages all the way through libfabric, where it's documented as a valid transfer (see [fi_rma](https://ofiwg.github.io/libfabric/main/man/fi_rma.3.html)).  But yeah, if `nblocks` is big then that's a pretty weird use-case ;)

src/data_c.c4
```diff
@@ -490,6 +581,34 @@ SHMEM_PROF_DEF_CTX_PUT_N_SIGNAL_NBI(`mem')
...
+  void SHMEM_FUNCTION_ATTRIBUTES                              \
+  SHMEMX_FUNC_PROTOTYPE(STYPE##_ibget, TYPE *target,          \
+                       const TYPE *source,                    \
+                       ptrdiff_t tst, ptrdiff_t sst,          \
+                       size_t bsize, size_t nblocks, int pe)  \
+    if (bsize == 0) return;                                   \
+    SHMEM_ERR_CHECK_INITIALIZED();                            \
+    SHMEM_ERR_CHECK_PE(pe);                                   \
+    SHMEM_ERR_CHECK_CTX(ctx);                                 \
+    SHMEM_ERR_CHECK_NON_NEGATIVE(bsize);                      \
```
davidozog, line 594: We might not need this given the datatype is a `size_t` (unsigned long).

src/data_c.c4
```diff
@@ -439,6 +479,31 @@ SHMEM_PROF_DEF_CTX_PUT_N_SIGNAL_NBI(`mem')
...
+  void SHMEM_FUNCTION_ATTRIBUTES                              \
+  SHMEMX_FUNC_PROTOTYPE(STYPE##_ibput, TYPE *target,          \
+                       const TYPE *source, ptrdiff_t tst,     \
+                       ptrdiff_t sst, size_t bsize, size_t nblocks, int pe)  \
+    if (bsize == 0) return;                                   \
+    SHMEM_ERR_CHECK_INITIALIZED();                            \
+    SHMEM_ERR_CHECK_PE(pe);                                   \
+    SHMEM_ERR_CHECK_CTX(ctx);                                 \
+    SHMEM_ERR_CHECK_NON_NEGATIVE(bsize);                      \
+    SHMEM_ERR_CHECK_SYMMETRIC(target, sizeof(TYPE) * ((nblocks-1) * tst + 1)); \
```
davidozog, line 492: I'm thinking the size of the region is `sizeof(TYPE) * ((nblocks-1) * tst + bsize))`..  is that right?

src/data_c.c4
```diff
@@ -439,6 +479,31 @@ SHMEM_PROF_DEF_CTX_PUT_N_SIGNAL_NBI(`mem')
...
+    SHMEM_ERR_CHECK_NULL(source, nblocks);                    \
+    SHMEM_ERR_CHECK_STRIDE_GTE_BSIZE(tst, bsize);             \
+    SHMEM_ERR_CHECK_STRIDE_GTE_BSIZE(sst, bsize);             \
+    SHMEM_ERR_CHECK_OVERLAP(target, source,                   \
+                   sizeof(TYPE) * ((nblocks-1) * tst + 1),    \
+                   sizeof(TYPE) * ((nblocks-1) * sst + 1), 0); \
```
davidozog, line 498:
```suggestion
                   sizeof(TYPE) * ((nblocks-1) * tst + bsize),    \
                   sizeof(TYPE) * ((nblocks-1) * sst + bsize), 0); \
```
I think?

src/data_c.c4
```diff
@@ -515,6 +634,35 @@ SHMEM_PROF_DEF_CTX_PUT_N_SIGNAL_NBI(`mem')
...
+  void SHMEM_FUNCTION_ATTRIBUTES                          \
+  SHMEMX_FUNC_PROTOTYPE(ibget##NAME, void *target,        \
+                       const void *source, ptrdiff_t tst, \
+                       ptrdiff_t sst, size_t bsize,       \
+                       size_t nblocks, int pe)            \
+    if (bsize == 0) return;                               \
+    SHMEM_ERR_CHECK_INITIALIZED();                        \
+    SHMEM_ERR_CHECK_PE(pe);                               \
+    SHMEM_ERR_CHECK_CTX(ctx);                             \
+    SHMEM_ERR_CHECK_NON_NEGATIVE(bsize);                  \
```
davidozog, line 647: We might not need this given the datatype is a `size_t` (unsigned long).

src/data_c.c4
```diff
@@ -464,6 +529,32 @@ SHMEM_PROF_DEF_CTX_PUT_N_SIGNAL_NBI(`mem')
...
+  SHMEMX_FUNC_PROTOTYPE(ibput##NAME, void *target,           \
+                       const void *source, ptrdiff_t tst,    \
+                       ptrdiff_t sst, size_t bsize, size_t nblocks, int pe) \
+    if (bsize == 0) return;                                  \
+    SHMEM_ERR_CHECK_INITIALIZED();                           \
+    SHMEM_ERR_CHECK_PE(pe);                                  \
+    SHMEM_ERR_CHECK_CTX(ctx);                                \
+    SHMEM_ERR_CHECK_NON_NEGATIVE(bsize);                     \
+    SHMEM_ERR_CHECK_SYMMETRIC(target,                        \
+                           (SIZE) * ((nblocks-1) * tst + 1)); \
```
davidozog, line 543: possibly:
```suggestion
                           (SIZE) * ((nblocks-1) * tst + bsize)); \
```

src/data_c.c4
```diff
@@ -464,6 +529,32 @@ SHMEM_PROF_DEF_CTX_PUT_N_SIGNAL_NBI(`mem')
...
+    SHMEM_ERR_CHECK_NULL(source, nblocks);                   \
+    SHMEM_ERR_CHECK_STRIDE_GTE_BSIZE(tst, bsize);            \
+    SHMEM_ERR_CHECK_STRIDE_GTE_BSIZE(sst, bsize);            \
+    SHMEM_ERR_CHECK_OVERLAP(target, source,                  \
+                        (SIZE) * ((nblocks-1) * tst + 1),    \
+                        (SIZE) * ((nblocks-1) * sst + 1), 0); \
```
davidozog, line 549: possibly:
```suggestion
                        (SIZE) * ((nblocks-1) * tst + bsize),    \
                        (SIZE) * ((nblocks-1) * sst + bsize), 0); \
```

davidozog: similar question below in ibget's.

src/data_c.c4
davidozog: Ha!  I had to draw a colorful picture myself - yes indeed I think it checks out.

.github/workflows/ci.yml
```diff
@@ -217,6 +217,7 @@ jobs:
       # SOS
       - name: Build SOS (${{ matrix.sos_config }})
         run: |
+          git submodule update --remote
```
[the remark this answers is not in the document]
philipmarshall21: Sounds good - we can keep them separate. Figured I might as well test it out with this PR and then revert the CI changes after finding out what works, then push the CI changes in a different PR. Even though I don't expect any issues with the ibput/ibget tests with different test configurations, it would be great to validate this PR with the new tests incorporated in the CI

src/data_c.c4
```diff
@@ -464,6 +529,32 @@ SHMEM_PROF_DEF_CTX_PUT_N_SIGNAL_NBI(`mem')
...
+    SHMEM_ERR_CHECK_STRIDE_GTE_BSIZE(sst, bsize);            \
+    SHMEM_ERR_CHECK_OVERLAP(target, source,                  \
+                        (SIZE) * ((nblocks-1) * tst + bsize), \
+                        (SIZE) * ((nblocks-1) * sst + bsize), 0); \
+    for ( ; nblocks > 0 ; --nblocks) {                       \
+      shmem_internal_put_nb(ctx, target, source,             \
+                            bsize * (SIZE), pe,              \
+                            &completion);                    \
+      target = (uint8_t*)target + tst*(SIZE);                \
+      source = (uint8_t*)source + sst*(SIZE);                \
```
wrrobin, line 554: Just nitpick: could you use similar spacing for `*` as in other lines?

src/shmem_internal.h
```diff
@@ -361,7 +369,8 @@ extern hwloc_topology_t shmem_internal_topology;
 #define SHMEM_ERR_CHECK_OVERLAP(ptr1, ptr2, size1, size2, complete_overlap_allowed)
 #define SHMEM_ERR_CHECK_NULL(ptr, nelems)
 #define SHMEM_ERR_CHECK_CMP_OP(op)
-#define SHMEM_ERR_CHECK_SIG_OP(op)                                                      \
+#define SHMEM_ERR_CHECK_SIG_OP(op)
```
wrrobin, line 372: 😮

wrrobin (approved):

src/data_c.c4
philipmarshall21: Sure thing! I had tried to keep the styling consistent with the existing implementations of iput/iget. Should I make the spacing for `*` consistent in those functions as well?

wrrobin: Oh, I see. Yeah, that would make sense. Thanks.

parkerha1 (approved):
lgtm

davidozog (approved):
Sorry for the delay, looks great!

src/data_c.c4
```diff
@@ -439,6 +479,31 @@ SHMEM_PROF_DEF_CTX_PUT_N_SIGNAL_NBI(`mem')
...
+    SHMEM_ERR_CHECK_NULL(source, nblocks);                    \
+    SHMEM_ERR_CHECK_STRIDE_GTE_BSIZE(tst, bsize);             \
+    SHMEM_ERR_CHECK_STRIDE_GTE_BSIZE(sst, bsize);             \
+    SHMEM_ERR_CHECK_OVERLAP(target, source,                   \
+                   sizeof(TYPE) * ((nblocks-1) * tst + 1),    \
+                   sizeof(TYPE) * ((nblocks-1) * sst + 1), 0); \
+    size_t i = 0;                                             \
+    for ( ; nblocks > 0 ; --nblocks) {                        \
+      shmem_internal_put_scalar(ctx, target, source,          \
+                                bsize * sizeof(TYPE), pe);    \
```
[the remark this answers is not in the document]
davidozog: Note to past Dave: I don't see any concern w/ large_put source buffer re-use.  There's always a `put_wait` in `data_c.c4` RMA definitions, so we're good.

PR #1116, opened 2024-03-14
Add error checking for objects on external heap

philipmarshall21:
Now checking if an object resides on the external heap when determining if said object is symmetric. This PR also revises the overlap check to allow overlap in target and source buffers for RMA operations if `target_pe != source_pe`

### DIFF
--- a/src/data_c.c4
+++ b/src/data_c.c4
@@ -287,7 +287,8 @@ SHMEM_PROF_DEF_CTX_PUT_N_SIGNAL_NBI(`mem')
     SHMEM_ERR_CHECK_SYMMETRIC(target, sizeof(TYPE) * nelems);    \
     SHMEM_ERR_CHECK_NULL(source, nelems);                        \
     SHMEM_ERR_CHECK_OVERLAP(target, source, sizeof(TYPE) *       \
-                            nelems, sizeof(TYPE) * nelems, 0);   \
+                            nelems, sizeof(TYPE) * nelems, 0,    \
+                            (shmem_internal_my_pe == pe));       \
     shmem_internal_put_nb(ctx, target, source,                   \
                           sizeof(TYPE) * nelems, pe,             \
                           &completion);                          \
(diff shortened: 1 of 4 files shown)


philipmarshall21:
Not convinced the change to SHMEM_ERR_CHECK_SYMMETRIC_HEAP is quite correct yet

philipmarshall21:
Seems like the overlap checks are missing entirely from **data_f.c**, **synchronization_f.c**, and **collectives_f.c**. Are those files still being maintained / does the **SHMEM_ERR_CHECK_OVERLAP** check need to be added to them?

davidozog:
> Seems like the overlap checks are missing entirely from **data_f.c**, **synchronization_f.c**, and **collectives_f.c**. Are those files still being maintained / does the **SHMEM_ERR_CHECK_OVERLAP** check need to be added to them?

That’s probably not a concern, those are for Fortran interfaces, which are all deprecated.

src/data_c.c4
```diff
@@ -287,7 +287,8 @@ SHMEM_PROF_DEF_CTX_PUT_N_SIGNAL_NBI(`mem')
     SHMEM_ERR_CHECK_SYMMETRIC(target, sizeof(TYPE) * nelems);    \
     SHMEM_ERR_CHECK_NULL(source, nelems);                        \
     SHMEM_ERR_CHECK_OVERLAP(target, source, sizeof(TYPE) *       \
-                            nelems, sizeof(TYPE) * nelems, 0);   \
+                            nelems, sizeof(TYPE) * nelems, 0,    \
+                            (shmem_internal_my_pe == pe));       \
```
parkerha1, line 291: Why do the checks in collectives_c get passed a 1 but these are passing the conditional? Do we know the collectives_c macros are already on the same PE?

src/shmem_internal.h
```diff
@@ -247,46 +247,66 @@ extern hwloc_topology_t shmem_internal_topology;
...
+            }                                                                                 \
+        }                                                                                     \
+        else if (ptr_base >= shmem_internal_heap_base && ptr_base < heap_ext) {               \
+            if (ptr_ext > heap_ext) {                                                         \
+                RAISE_ERROR_MSG("Argument \"%s\" [%p..%p) exceeds "                           \
+                                "sym. heap region [%p..%p)\n", #ptr_in, ptr_base,             \
+                                ptr_ext, shmem_internal_heap_base, heap_ext);                 \
+            }                                                                                 \
+        }                                                                                     \
+        else if (shmem_external_heap_base) {                                                  \
```
parkerha1, line 275: what kind of devices could provide an external heap? GPU's would be my first thought, but I was under the impression that SOS doesn't support using their memory.

parkerha1 (approved):
LGTM

src/data_c.c4
philipmarshall21: Good question. The "1" value is indicating the source/destination overlap check should always occur for APIs such as the collectives. However, for RMA operations, the source and destination buffers can overlap so long as the destination PE and source PE are different (i.e. `shmem_internal_my_pe == pe` evaluates to 0, ignoring the overlap check).

src/shmem_internal.h
philipmarshall21: In the case of Intel® SHMEM, we are using the external heap to allow SOS to interact with data in device (GPU) memory. SOS supports this usage when configured with **--enable-ofi-hmem**.

davidozog (approved):
Looks great 💯, thanks @philipmarshall21!

PR #1117, opened 2024-04-05
MR descriptor list

wrrobin:


src/transport_ofi.h
```diff
@@ -1353,8 +1354,8 @@ void shmem_transport_atomic_fetch(shmem_transport_ctx_t* ctx, void *target,
     shmem_transport_fetch_atomic_nbi(ctx, (void *) source, (const void *) &dummy,
                                      target, len, pe, FI_SUM, datatype);
 #else
-    shmem_transport_fetch_atomic(ctx, (void *) source, (const void *) NULL,
-                                 target, len, pe, FI_ATOMIC_READ, datatype);
+    shmem_transport_fetch_atomic_nbi(ctx, (void *) source, (const void *) NULL,
+                                     target, len, pe, FI_ATOMIC_READ, datatype);
```
davidozog, line 1358: Is this change related to the MR descriptors?  I'm not sure I understand the reason for the change, can you please explain?  It seems relatively harmless... but wouldn't `shmem_transport_atomic_swap` need this change also (for consistency)?

davidozog (commented):
@wrrobin, do you happen to know how the MR descriptors are traversed by the provider?  For example, if there is a list of 3 descriptors and the target is in the external heap, are all 3 MR descriptors read to lookup which applies?  Could there be a performance cost associated with that?

src/transport_ofi.h
wrrobin: No, this is not related. But, the reason of this change is after this call, there is always a `get_wait` call which ensures completion in the internal layer. I might have missed `swap`, let me take a look.

wrrobin:
> @wrrobin, do you happen to know how the MR descriptors are traversed by the provider? For example, if there is a list of 3 descriptors and the target is in the external heap, are all 3 MR descriptors read to lookup which applies? Could there be a performance cost associated with that?

Great question. And, I do not know what order it is traversed. Let me ask this question and add you to that email thread.

wrrobin:
@rithwiktom - Could you get a chance to test this out on your end?

rithwiktom (approved):
Tested and works on ORPRE.

davidozog (approved):

PR #1119, opened 2024-04-15
Multi-NIC Bug Fix: Set 'next' attribute of list tail to NULL

philipmarshall21:
This PR makes a minor bug fix to the code that is responsible for generating a list of providers that have NIC information. This should resolve the issues we were seeing on some of our development systems that have a single NIC when SOS is being used with multi-NIC functionality enabled.

### DIFF
--- a/src/transport_ofi.c
+++ b/src/transport_ofi.c
@@ -1562,7 +1562,7 @@ int query_for_fabric(struct fabric_info *info)
                 multirail_fabric_list_tail = cur_fabric;
             }
         }
-        multirail_fabric_list_tail->next = NULL;
+        if (multirail_fabric_list_tail) multirail_fabric_list_tail->next = NULL;
 
         if (num_nics == 0) {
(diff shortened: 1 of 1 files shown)


wrrobin (approved):

davidozog (approved):

src/transport_ofi.c
```diff
@@ -1562,6 +1562,7 @@ int query_for_fabric(struct fabric_info *info)
                 multirail_fabric_list_tail = cur_fabric;
             }
         }
+        multirail_fabric_list_tail->next = NULL;
```
wrrobin, line 1565: This can generate a null pointer assignment if it is not assigned within the for loop above.

davidozog (approved):

src/transport_ofi.c
philipmarshall21: Looks like it was the cause of the test failure. Thanks for catching it!

Issue #1019, opened 2021-10-05
num_contexts not considered in shmem_team_config_t


davidozog:
SOS currently doesn't consider the `num_contexts` hint in `shmem_team_config_t`.  It grows on-demand by multiples of 128.  Maybe it would be advantageous to pre-allocate `config->num_contexts` contexts on teams?

PR #1122, opened 2024-04-30
num_contexts patch

davidozog:
This resolves the issue discovered by @ronawho regarding unit test `shmem_team_get_config` here:
https://github.com/openshmem-org/tests-sos/pull/37

~~This is just a temporary patch to better align with the API spec.  SOS still effectively ignores the `num_contexts` hint, so #1019 is not quite addressed.~~

This also addresses #1019

### DIFF
--- a/src/shmem_team.c
+++ b/src/shmem_team.c
@@ -334,10 +334,22 @@ int shmem_internal_team_split_strided(shmem_internal_team_t *parent_team, int PE
         myteam->start       = global_PE_start;
         myteam->stride      = PE_stride;
         myteam->size        = PE_size;
-        if (config) {
-            myteam->config      = *config;
-            myteam->config_mask = config_mask;
+
+        if (config_mask == 0) {
+            shmem_team_config_t default_config;
+            default_config.num_contexts = 0;
+            myteam->config_mask = 0;
+            memcpy(&myteam->config, &default_config, sizeof(shmem_team_config_t));
+        } else {
+            if (config_mask != SHMEM_TEAM_NUM_CONTEXTS) {
+                RAISE_WARN_MSG("Invalid team_split_strided config_mask (%ld)\n", config_mask);
+                return -1;
+            } else {
+                myteam->config      = *config;
+                myteam->config_mask = config_mask;
+            }
         }
+
         myteam->contexts_len = 0;
         myteam->psync_idx = -1;
 
--- a/src/shmem_team.c
+++ b/src/shmem_team.c
@@ -336,21 +336,27 @@ int shmem_internal_team_split_strided(shmem_internal_team_t *parent_team, int PE
         myteam->size        = PE_size;
 
         if (config_mask == 0) {
-            shmem_team_config_t default_config;
-            default_config.num_contexts = 0;
-            myteam->config_mask = 0;
-            memcpy(&myteam->config, &default_config, sizeof(shmem_team_config_t));
+            shmem_team_config_t defaults;
+            myteam->config_mask   = 0;
+            myteam->contexts_len  = 0;
+            defaults.num_contexts = 0;
+            memcpy(&myteam->config, &defaults, sizeof(shmem_team_config_t));
         } else {
             if (config_mask != SHMEM_TEAM_NUM_CONTEXTS) {
                 RAISE_WARN_MSG("Invalid team_split_strided config_mask (%ld)\n", config_mask);
                 return -1;
             } else {
-                myteam->config      = *config;
-                myteam->config_mask = config_mask;
+                shmem_internal_assertp(config->num_contexts >= 0);
+                myteam->config       = *config;
+                myteam->config_mask  = config_mask;
+                myteam->contexts_len = config->num_contexts;
+                myteam->contexts     = malloc(config->num_contexts * sizeof(shmem_transport_ctx_t*));
+                for (int i = 0; i < config->num_contexts; i++) {
+                    myteam->contexts[i] = NULL;
+                }
             }
         }
 
-        myteam->contexts_len = 0;
         myteam->psync_idx = -1;
 
         shmem_internal_op_to_all(psync_pool_avail_reduced,
--- a/src/shmem_team.c
+++ b/src/shmem_team.c
@@ -336,6 +336,10 @@ int shmem_internal_team_split_strided(shmem_internal_team_t *parent_team, int PE
         myteam->size        = PE_size;
 
         if (config_mask == 0) {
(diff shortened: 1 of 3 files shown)


philipmarshall21 (approved):

src/shmem_team.c
```diff
@@ -334,11 +334,29 @@ int shmem_internal_team_split_strided(shmem_internal_team_t *parent_team, int PE
...
-        if (config) {
-            myteam->config      = *config;
-            myteam->config_mask = config_mask;
+
+        if (config_mask == 0) {
+            shmem_team_config_t defaults;
+            myteam->config_mask   = 0;
+            myteam->contexts_len  = 0;
+            defaults.num_contexts = 0;
+            memcpy(&myteam->config, &defaults, sizeof(shmem_team_config_t));
```
wrrobin, line 343: Do we have to assert that `config` argument is NULL/empty in this case?

bcmIntc: If we use assert(), my understanding is that's only built for debug.  Seems we need a NULL check for release?

bcmIntc (approved):
Assuming Wasi's question is addressed, looks good!

src/shmem_team.c
davidozog: The spec says a `config_mask` value of 0 indicates that all `config` parameters should be set to the default values, so I don't think we need to check `config`.  However, it's likely beneficial to users to know the config is being ignored in case they accidentally pass a 0 mask.  Perhaps we can add a check if `ENABLE_ERROR_CHECKING` is defined?

wrrobin: Yes, that sounds good. Keeping it under `ENABLE_ERROR_CHECKING` also makes sense.

davidozog: @wrrobin/@bcmIntc - I implemented the error check, but now I'm reconsidering, because this case probably shouldn't generate an error (e.g., the test in https://github.com/openshmem-org/tests-sos/pull/37 is valid usage, but would fail with error-checking).  How about printing a warning instead?

wrrobin: @davidozog - Oh.. sorry, I missed that you meant as an error. I was actually thinking only a warning to the developer that the supplied config cannot be used and ignored.

davidozog: @wrrobin, currently `ENABLE_ERROR_CHECKING` only reports errors (resulting in aborts), and the checks are mostly on the critical path (so best to enable by default).  I think a simple warning is better in this case... and it's not on the critical path so we can just keep the warning around.  Do you agree?

wrrobin: Yes. Keeping it as debug message is another option, but I think warning is better.

parkerha1 (approved):
LGTM

wrrobin (approved):

src/shmem_team.c
```diff
@@ -334,11 +334,33 @@ int shmem_internal_team_split_strided(shmem_internal_team_t *parent_team, int PE
...
         myteam->size        = PE_size;
-        if (config) {
-            myteam->config      = *config;
-            myteam->config_mask = config_mask;
+
+        if (config_mask == 0) {
+            if (config != NULL) {
+                RAISE_WARN_MSG("%s %s\n", "team_split_strided operation encountered an unexpected",
+                               "non-NULL config structure passed with a config_mask of 0.");
+            }
```
mrogowski, line 342: @davidozog - @dalcinl and I disagree. 🙂 We started getting warnings in [shmem4py](https://github.com/mpi4py/shmem4py). We checked the specification and we do not see it saying `config` should be `NULL` if `config_mask == 0`. In fact, we are not even sure if a NULL pointer is allowed.

dalcinl: > In fact, we are not even sure if a NULL pointer is allowed.

In here, Marcin is talking about the standard: there is no explicit wording saying that if the `config_mask` is 0, then config may be `NULL`. Or did we miss something? What if some implementation decides to add a check `assert(config != NULL)` irrespective of the value of `config_mask`? Would such behavior be in contradiction of the 1.5 standard?

davidozog: @mrogowski @dalcinl I agree there is a "blind spot" in this section of the standard right now (v1.5), and I have a note to try to fix it for the upcoming new version (v1.6).  I would prefer to add the following statement to the standard (do you think it's sufficient?):
```
If \VAR{config} is a null pointer, then \VAR{config_mask} must be 0, otherwise
the behavior is undefined.
```
We also might want something like this for improved clarity:
```
If \VAR{config_mask} is 0, then `shmem_team_get_config` performs no action.
```

So if `config_mask` is 0 and `config` is non-null, then that's perfectly fine.  Given the state of OpenSHMEM v1.5, we opted to include an SOS warning in this special case, but I think we could also move it to "DEBUG" output or simply remove it altogether - I'd prefer to remove it myself, especially if the 1st new statement above is added to OpenSHMEM v1.6.

Another option is to simply prohibit a null `config` pointer, but my hunch is that's a bit more restrictive than what was intended for v1.5, but I'm open to it!

Any preferences from @wrrobin and @stewartl318?

dalcinl: I agree with your previous suggestions with a small addition for maximum clarity:

```
 If \VAR{config_mask} is 0, then `shmem_team_get_config` performs no action and \VAR{config} may be `NULL`.
```

> but I think we could also move it to "DEBUG" output or simply remove it altogether

I guess that would be good enough. Extra points if you guys ever allow for these warnings as an opt-in via some environment variable.

> Another option is to simply prohibit a null `config` pointer, but my hunch is that's a bit more restrictive than what was intended for v1.5, but I'm open to it!

Indeed, there is little point in such restriction. Moreover, it is kind of a backward incompatible change.

stewartl318: I think the internet robustness principle states it well: be conservative in what you do, be liberal in what you accept from others

I think there is no real value in any sort of debug or error message when mask is 0 but config is non-null.

davidozog: @dalcinl @stewartl318 - Thanks for the input!

>  Extra points if you guys ever allow for these warnings as an opt-in via some environment variable.

This sounds like what I meant by "DEBUG" message above.  The `DEBUG_MSG` macro in SOS will only print to `stderr` if the `SHMEM_DEBUG` environment variable is set.

> I think there is no real value in any sort of debug or error message when mask is 0 but config is non-null.

I tend to agree... but maybe the debug message is a good compromise since the spec is pretty under-defined for this special case.  Let's move the discussion to PR #1138, which proposes changing this to a debug message.

PR #1123, opened 2024-05-01
Signal set add

wrrobin:


philipmarshall21 (approved):
Looks good! 👍

davidozog (commented):
Looks like this is testing against tests-sos commit https://github.com/openshmem-org/tests-sos/commit/26de6baf701ee66c04db4d63d58206ea67436c3b, is that rigth?

Perhaps we can update it to tests-sos `main`, rerun the tests and call it good?

davidozog:
@wrrobin - hope you don't mind I manually updated the submodule to point to `main` (only updating .gitmodules still seems not to be enough).  And interestingly, there is a failure with `mt_membar` with XPMEM+shr_atomics.

wrrobin:
> @wrrobin - hope you don't mind I manually updated the submodule to point to `main` (only updating .gitmodules still seems not to be enough). And interestingly, there is a failure with `mt_membar` with XPMEM+shr_atomics.

@davidozog - Did it pass in a rerun?

davidozog:
> > @wrrobin - hope you don't mind I manually updated the submodule to point to `main` (only updating .gitmodules still seems not to be enough). And interestingly, there is a failure with `mt_membar` with XPMEM+shr_atomics.
> 
> @davidozog - Did it pass in a rerun?

Yes it did... I was unable to reproduce it on neptune, trying Ubuntu now...

davidozog (approved):
That xpmem test passed about 9 or 10 times in a row, so merging.  Let's keep an eye on it..

wrrobin:
Sounds good. Thanks @davidozog

PR #1130, opened 2024-05-20
configury: update much configury to OMPI main

davidozog:
This PR aligns SOS's configury with OMPI's main branch ref [cebad71e](https://github.com/open-mpi/ompi/commit/cebad71e4f50d49e486ccd54668cf58698556a7c)

It removes several configure-time warnings about obsolete autotools features and adds relevant updates for OFI (esp. for HMEM), UCX, and Portals4.

All changes from `OMPI/config/*` are verbatim, and `SOS/configure.ac` was run through `autoupdate`.

### DIFF
--- a/config/getdate.sh
+++ b/config/getdate.sh
@@ -1,0 +1,14 @@
+#!/bin/sh
+#
+# Copyright (c) 2017      Research Organization for Information Science
+#                         and Technology (RIST). All rights reserved.
+# Copyright (c) 2020      Cisco Systems, Inc.  All rights reserved.
+
+# Provide a way to override build date for reproducible build results
+# See https://reproducible-builds.org/ for why this is good.
+
+# There are several different flavors of date(1) out there.
+# Try a few different CLI options for date(1) to see which one works.
+
+SOURCE_DATE_EPOCH="${SOURCE_DATE_EPOCH:-$(date +%s)}"
+date -u -d "@$SOURCE_DATE_EPOCH" "$@" 2>/dev/null || date -u -r "$SOURCE_DATE_EPOCH" "$@" 2>/dev/null || date -u "$@"
--- a/config/ompi_check_ucx.m4
+++ b/config/ompi_check_ucx.m4
@@ -7,6 +7,7 @@
 # Copyright (c) 2016      Los Alamos National Security, LLC. All rights
 #                         reserved.
 # Copyright (c) 2016 Cisco Systems, Inc.  All rights reserved.
+# Copyright (c) 2022      Amazon.com, Inc. or its affiliates.  All Rights reserved.
 # $COPYRIGHT$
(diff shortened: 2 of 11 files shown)


markbrown314 (approved):

bcmIntc:
> I took a cursory look at the code generated and it looks fine to me. Generated code is notoriously difficult to audit for correctness. I will +1 it, but I think we should add some tests (perhaps in CI) to validate that the parameters provided by configure work as advertised.

I agree, I feel we should strive to have test cases either through direct (code) testing or via CI coverage in general.

davidozog:
> Generated code is notoriously difficult to audit for correctness.

You might be surprised to learn that this code is not generated for the most part - I'm pretty sure many humans have contributed to OMPI's configury over the years.  For example, this is the commit history on the UCX config file:
https://github.com/open-mpi/ompi/commits/main/config/ompi_check_ucx.m4

>  I think we should add some tests (perhaps in CI) to validate that the parameters provided by configure work as advertised.

The existing CI has pretty good coverage on the changes in this PR.  Here are some pointers to where that happens in the [SOS ci.yml ](https://github.com/Sandia-OpenSHMEM/SOS/blob/v1.5.3/.github/workflows/ci.yml) file:

- `--with-ofi`: https://github.com/Sandia-OpenSHMEM/SOS/blob/v1.5.3/.github/workflows/ci.yml#L227
- `--with-ucx`: https://github.com/Sandia-OpenSHMEM/SOS/blob/v1.5.3/.github/workflows/ci.yml#L553
- `--wiht-portals4` https://github.com/Sandia-OpenSHMEM/SOS/blob/v1.5.3/.github/workflows/ci.yml#L647

That said, it's not perfect coverage.  We're not testing scenarios where the path is not provided explicitly, for example.

`opal_case_sensitive_fs_setup.m4` I've checked on my Mac and it looks like a good improvement in my opinion.

I double checked `attributes` and `visibility` manually.  These are not user-facing so a little tougher to integrate in CI...

Anyway, does that address the concerns or am I missing something?

One issue with this PR is that I integrated with a particularly [random upstream commit](https://github.com/open-mpi/ompi/commit/cebad71e4f50d49e486ccd54668cf58698556a7c)... I believe at the time I did this work, the latest OMPI release had some configury issue, but I can't recall exactly...  Maybe it's better to try to align with a newer release, but I don't see a strong need as long as we take note of where exactly we're cribbing this from.

The license coverage should be good unless we missed something in the past.  Only the `getdate.sh` file is new, and it should be covered by our citation of the https://spdx.org/licenses/BSD-3-Clause-Open-MPI license.

abrooks98 (approved):

markbrown314:
@davidozog thanks for the change set and the insights. Please merge in when ready.

PR #1133, opened 2024-06-05
fix pedantic compiler warnings

davidozog:
These are small fixes for compiler warnings found with `--enable-pedantic`.

### DIFF
--- a/src/transport_ofi.c
+++ b/src/transport_ofi.c
@@ -1348,6 +1348,7 @@ int allocate_fabric_resources(struct fabric_info *info)
 }
 
 #ifdef USE_HWLOC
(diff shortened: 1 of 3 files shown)


src/transport_ofi.c
```diff
@@ -1348,6 +1348,7 @@ int allocate_fabric_resources(struct fabric_info *info)
 }
 
 #ifdef USE_HWLOC
+static inline
```
philipmarshall21, line 1351: Just curious, why is **static inline** required?

philipmarshall21 (approved):

wrrobin (approved):

PR #1134, opened 2024-06-10
Add DISABLE_NONFETCH_AMO support to op_to_all implementations

philipmarshall21:
This PR extends the support of the **--disable-nonfetch-amo** configuration option to avoid the use of non-fetching atomics in **shmem_internal_op_to_all_linear** and **shmem_internal_op_to_all_tree**

### DIFF
--- a/src/collectives.c
+++ b/src/collectives.c
@@ -613,19 +613,8 @@ shmem_internal_op_to_all_linear(void *target, const void *source, size_t count, 
         SHMEM_WAIT_UNTIL(pSync, SHMEM_CMP_EQ, 0);
 
         /* send data, ack, and wait for completion */
-#ifdef DISABLE_NONFETCH_AMO
-        /* FIXME: This is a temporary workaround to resolve a known issue with non-fetching AMOs when using
-           the CXI provider */
-        unsigned long long tmp_fetch = 0;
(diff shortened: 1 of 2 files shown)


wrrobin (approved):

src/collectives.c
```diff
@@ -613,8 +613,19 @@ shmem_internal_op_to_all_linear(void *target, const void *source, size_t count,
         SHMEM_WAIT_UNTIL(pSync, SHMEM_CMP_EQ, 0);
 
         /* send data, ack, and wait for completion */
+#ifdef DISABLE_NONFETCH_AMO
```
davidozog, line 616: Is it possible these patches could go into `shmem_internal_atomicv`?  If so, it seems like that might be better "layering".

philipmarshall21: Sure! I've made the recommended change, though it does require reacquiring the **type_size** and **count** variables from the **len** parameter passed into **shmem_internal_atomicv**, which does require a division operation.

davidozog: Thanks @philipmarshall21.  I'd say the extra division probably isn't too bad considering the ~1/2 round-trip-time added by this workaround (which is hopefully temporary anyway).

davidozog (approved):

PR #1137, opened 2024-06-21
OFI/CXI: add intermittent quiet for CXI workaround

davidozog:
This is another temporary workaround related to PR #1134.

### DIFF
--- a/src/transport_ofi.h
+++ b/src/transport_ofi.h
@@ -40,6 +40,10 @@
 #define ENABLE_TARGET_CNTR 0
 #endif
 
+#if DISABLE_NONFETCH_AMO
+#define SOS_OFI_CXI_QUIET_RATE 200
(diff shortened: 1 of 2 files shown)


src/transport_ofi.h
```diff
@@ -40,6 +40,10 @@
 #define ENABLE_TARGET_CNTR 0
 #endif
 
+#if DISABLE_NONFETCH_AMO
+#define SOS_OFI_CXI_QUIET_RATE 200
+#endif
```
wrrobin, line 45: Should we keep a comment here regarding the reason of tis?

wrrobin (approved):

src/transport_ofi.h
davidozog: Sure - added.

PR #1147, opened 2024-09-18
Fix for atomicv on CXI

wrrobin:


src/shmem_comm.h
```diff
@@ -336,6 +336,7 @@ shmem_internal_atomicv(shmem_ctx_t ctx, void *target, const void *source,
         shmem_internal_fetch_atomic(ctx, ((uint8_t *) target) + (i * type_size),
                                     ((uint8_t *) source) + (i * type_size), &tmp_fetch, type_size,
                                     pe, op, datatype);
+        shmem_transport_get_wait((shmem_transport_ctx_t *)ctx);
```
davidozog, line 339: Can we move this outside the loop over `count`?  I don't quite see why we need to wait after every atomic?

src/shmem_comm.h
```diff
@@ -336,6 +336,7 @@ shmem_internal_atomicv(shmem_ctx_t ctx, void *target, const void *source,
...
 #ifdef SOS_OFI_CXI_QUIET_RATE
```
davidozog, line 340: Is there any chance we can remove this after the above fix?

src/shmem_comm.h
wrrobin: Nope, we cant as the loop itself can go beyond the number of atomics that causes the issue on CXI.

src/shmem_comm.h
wrrobin: I think based on the atomic OP, we cant get rid of this completely. But, will check.

davidozog (approved):
As discussed offline, perhaps `SHMEM_REDUCE_ALGORITHM=recdbl` (maybe `ring`?) is a better workaround (when disabling non-fetching AMO) given this could dramatically impair the performance of `linear` and `tree` algorithms.

Issue #1151, opened 2024-10-09
SHMEM_TEAM_SPLIT_STRIDED expects global stride instead of relative stride?


zhongchen530:
The function SHMEM_TEAM_SPLIT_STRIDED takes in a parent team, start, size, and stride argument to produce a new team. I would expect in this case the stride to be relative to the parent team. However, what I observed was that it expects the stride to be relative to the SHMEM_TEAM_WORLD instead.

For instance if number of PEs is 4, meaning SHMEM_TEAM_WORLD : {0,1,2,3,4,5}.

shmem_team_split_strided(SHMEM_TEAM_WORLD, 0, 2, 3, NULL, 0, &new_team);

would result in 

new_team : {0,2,4}

If I again call 

shmem_team_split_strided(new_team, 0, 2, 2, NULL, 0, &another_team);

this would result in 

another_team : {0, 2} 

The stride is still relative to the global PE but not the parent team new_team in this case.

Is this an intended behavior?

davidozog:
@zhongchen530 - Oops!  I had to write all that to see that maybe SOS does _not_ behave how I described, and you found an issue.  Will investigate..

zhongchen530:
> @zhongchen530 - Oops! I had to write all that to see that maybe SOS does _not_ behave how I described, and you found an issue. Will investigate..

Yes, your explanation above is what I initially expected, but it doesn't behave that way. Instead, ```another_team``` is observed to be ```{0,2}``` instead.

davidozog:
Closed by #1153

PR #1153, opened 2024-10-10
teams: input split stride is relative to parent

davidozog:
This is a potential fix for #1151 

@zhongchen530 - If you are able, can you please check if this change results in the expected behavior for `shmem_team_split_strided`?  I'm also curious if you can please share any details about your use-case that identified this issue.  Or perhaps you are simply exploring and/or learning the API?

### DIFF
--- a/src/shmem_team.c
+++ b/src/shmem_team.c
@@ -299,7 +299,8 @@ int shmem_internal_team_split_strided(shmem_internal_team_t *parent_team, int PE
     }
 
     int global_PE_start = shmem_internal_team_pe(parent_team, PE_start);
-    int global_PE_end   = global_PE_start + PE_stride * (PE_size -1);
+    int global_PE_stride = parent_team->stride * PE_stride;
+    int global_PE_end = global_PE_start + global_PE_stride * (PE_size -1);
 
     if (PE_start < 0 || PE_start >= parent_team->size ||
         PE_size <= 0 || PE_size > parent_team->size   ||
@@ -318,7 +319,7 @@ int shmem_internal_team_split_strided(shmem_internal_team_t *parent_team, int PE
     }
 
     int my_pe = shmem_internal_pe_in_active_set(shmem_internal_my_pe,
-                                                global_PE_start, PE_stride, PE_size);
+                                                global_PE_start, global_PE_stride, PE_size);
 
     long *psync = shmem_internal_team_choose_psync(parent_team, REDUCE);
     shmem_internal_team_t *myteam = NULL;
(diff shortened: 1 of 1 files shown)


src/shmem_team.c
```diff
@@ -299,7 +299,8 @@ int shmem_internal_team_split_strided(shmem_internal_team_t *parent_team, int PE
     }
 
     int global_PE_start = shmem_internal_team_pe(parent_team, PE_start);
-    int global_PE_end   = global_PE_start + PE_stride * (PE_size -1);
+    int global_PE_stride = parent_team->stride * PE_stride;
```
wrrobin, line 302: Do we need to take the absolute value?

davidozog: I don't think so... it can help to think of an example.

Let's assume `parent_team` is a team that was split from team world with a negative stride, and `PE_stride` is positive.

Say, `world = {0,1,2,3}` and we split with (start,stride,size) of (3,-1,4) so `parent_team = {3,2,1,0}`.  Now we do another simple duplication split of `parent_team` with (0,2,2).  In this case we want `new_team = {3,1}`.

For the duplication split, since `parent_team->stride` is negative (it's the global stride, -1), and `PE_stride` is positive 2, we want the global stride to remain negative.  In this case -2.  With absolute value, the result above would lead to a "team wraparound" in the positive direction.

Does that makes sense?

wrrobin: Thanks @davidozog. This definitely helps. I was thinking the example differently, I guess. 
For example, in your case, if the 2nd split is chosen to be (3, -2, 2) on parent team {3, 2, 1, 0} to get a team {2,0}. Then because we are not taking the absolute value, the `global_pe_stride` would be positive, which would cause wrap-around, right? Did I miss something?

davidozog: I think if the 2nd split is chosen to be (3, -2, 2) on parent team {3, 2, 1, 0}, this would actually result in team {0,2}, which has a positive stride relative to the global team.

So indeed two negative strides indeed make a positive stride.  Let me think if I can prove this more formally...

wrrobin (approved):

bcmIntc (approved):

philipmarshall21 (approved):

zhongchen530:
> This is a potential fix for #1151
> 
> @zhongchen530 - If you are able, can you please check if this change results in the expected behavior for `shmem_team_split_strided`? I'm also curious if you can please share any details about your use-case that identified this issue. Or perhaps you are simply exploring and/or learning the API?

I apologize for the delayed reply. I am currently working on writing a more complex test suite with more functional tests of various API. I discovered this issue when I tried to write a test where I recursively split teams into even and odd teams, like even of even and odd of even etc. I was going to rebuild shmem using the pull request and was waiting for my virtual cluster  to be cloned before doing so. Unfortunately, there was some issue with the cloning of the VM on my side I wouldn't be able to change the shmem version until that is resolved.

zhongchen530:
Hi @davidozog, I finally got permission to rebuild shmem on my cluster with the updated main branch. The original stride issue related to ```shmem_team_split``` is now fixed. However, when I ran ```make test``` to run the test suite,  ```shmem_team_tsplit_2d``` is now failing to pass. I ran the test with 4 PES. I am afraid the change have broken ```shmem_team_split_2d call or that the test itself was wrong in the first place.

PR #1157, opened 2024-10-24
teams: force split stride to be nonzero + cleanup

davidozog:
Global stride calculations might be incorrect if user passes a stride of 0 (or a size of 1 and some non-unity stride).  This ensures nonzero internal stride values, and cleans up the sanity checks.

### DIFF
--- a/src/shmem_team.c
+++ b/src/shmem_team.c
@@ -298,23 +298,27 @@ int shmem_internal_team_split_strided(shmem_internal_team_t *parent_team, int PE
         return 1;
     }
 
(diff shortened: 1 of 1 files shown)


wrrobin (approved):
@davidozog - These changes look good to me. Is there any concern for any of these changes in Coverity?

davidozog:
> @davidozog - These changes look good to me. Is there any concern for any of these changes in Coverity?

Thanks @wrrobin!  I doubt it, but I'll check to be sure.

davidozog:
Coverity checks out, so I'll merge.

Issue #1159, opened 2024-12-05
Off-by-one error for gethostname() in runtime code


markbrown314:
src/runtime_util.c: shmem_runtime_util_put_hostname() calls gethostname() with a length of MAX_HOSTNAME_LEN (which resolves to 64 bytes in Linux). 

```C
    ret = gethostname(hostname, MAX_HOSTNAME_LEN);
    if (ret != 0) {
        RETURN_ERROR_MSG("gethostname failed (%d)", ret);
        return ret;
    }
```

The issue is MAX_HOSTNAME_LEN is the maximum total characters of the hostname not including the end of string character '\0'. If the hostname happens to be 64 characters long, gethostname() would return an error and shmem_runtime_util_put_hostname() would state error "gethostname failed (-1)"

The input length parameter needs to be set to MAX_HOSTNAME_LEN+1

markbrown314:
This has been resolved!

PR #1160, opened 2024-12-05
Fixed Off-by-one error in gethostname()

markbrown314:
MAX_HOSTNAME_LEN is the maximum total characters of the hostname not including the end of string character '\0'. If the hostname happens to be 64 characters long, gethostname() would return an error and shmem_runtime_util_put_hostname() would state error "gethostname failed (-1)"

Changed length to MAX_HOSTNAME_LEN+1

Issue #1159

### DIFF
--- a/src/runtime_util.c
+++ b/src/runtime_util.c
@@ -89,7 +89,7 @@ int shmem_runtime_util_put_hostname(void)
     char hostname[MAX_HOSTNAME_LEN+1];
     int ret;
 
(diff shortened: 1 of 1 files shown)


markbrown314:
Good point let me look at those.

markbrown314:
common.h::check_hostname_validation() will need to be fixed in the test-sos tree.

bcmIntc (approved):
looks good

PR #1167, opened 2025-01-31
tests: update tests-sos submodule to latest commit

davidozog:
I propose we update the `tests-sos` submodule so the new scan collectives are covered:

Should be https://github.com/openshmem-org/tests-sos/commit/ef46e42a85dadaa643e8ba44a74e37f79bc1ec73


markbrown314 (approved):
This change reflects previously approved updates in the test-sos for v.1.5.3. Approved!

PR #1168, opened 2025-02-12
[Bug Fix] Changed shmem_team_destroy to return void

avincigu:


src/shmem_team.c
```diff
@@ -504,11 +504,11 @@ int shmem_internal_team_split_2d(shmem_internal_team_t *parent_team, int xrange,
     return 0;
 }
 
-int shmem_internal_team_destroy(shmem_internal_team_t *team)
+void shmem_internal_team_destroy(shmem_internal_team_t *team)
 {
 
     if (team == SHMEM_TEAM_INVALID) {
-        return -1;
+        return;
```
avincigu, line 511: This is the only change I think would merit discussion. It used to return -1 but no error was really displayed. My question is if the team is invalid do we want to RAISE an error like we do below?

bcmIntc: The copy of the spec I have laying around says 'If team compares equal to SHMEM_TEAM_INVALID, then no operation is performed'. I think this means you might be able to just return silently.

markbrown314 (commented):
@avincigu in line src/teams_c.c4:147 the code is attempting to return an int even though the function does not return a value. Please remove this return.

```
teams_c.c: In function ‘shmem_team_destroy’:
teams_c.c:147:12: warning: ISO C forbids ‘return’ with expression, in function returning void [-Wpedantic]
147 |     return shmem_internal_team_destroy((shmem_internal_team_t *)team);
|            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
teams_c.c:139:1: note: declared here
139 | shmem_team_destroy(shmem_team_t team)
| ^~~~~~~~~~~~~~~~~~
```

bcmIntc (approved):

davidozog:
Looks good - let's just loop in @abrooks98 to see if Intel SHMEM could be affected?  (Maybe you checked already?)  At a glance, it actually looks like it's already good to go, but I could be wrong.

davidozog (approved):

markbrown314 (approved):
@avincigu looks good. Please merge (with select squash and merge)

PR #1169, opened 2025-02-19
Removing deprecated OFI_USE_PROVIDER env variable

avincigu:


davidozog (approved):
I'm good with this change.

I'm also dubious there are valid use-cases for `SHMEM_OFI_PROVIDER` (when FI_PROVIDER could/should be preferred), but I suppose we should keep it around in case there are more complicated scenarios alongside other runtimes...  See for example https://github.com/ofiwg/libfabric/issues/8079.  Takeaway: SOS itself does not set `FI_PROVIDER`, and it shouldn't going forward.  But users can set FI_PROVIDER if desired... it filters the available providers in libfabric at init.

It might be good to document somewhere that `SHMEM_OFI_PROVIDER` sets `prov_name` in the libfabric Fabric Attributes structure ([see fi_fabric.3](https://ofiwg.github.io/libfabric/main/man/fi_fabric.3.html)), nothing more.

Issue #1162, opened 2025-01-17
--disable-rpath has no effect on libsma.so


davidozog:
On 5fb1d6e05f5cd6e7d69d4464186d597605c33bdd the `--disable-rpath` configure flag seems to have no effect on the linkage of `libsma.so`.  However, it does seem to do the right thing (e.g., add ` -Wl,-rpath -Wl,/path/to/libfabric/lib -Wl,-rpath -Wl,/path/to/SOS/lib`) to the compiler wrappers (`oshcc` and friends). I'm unsure if this is a regression.

Thanks for the report @tmh97 and @bcernohous.

markbrown314:
I confirmed the bug --disable-rpath does not have any affect. There is no code in the configure script to handle this flag.

markbrown314:
This issue is caused by the generated libtool library setting the variable hardcode_into_libs to yes.

markbrown314:
@davidozog please validate whether PR #1188 resolves your issue and then I will merge.

davidozog:
This issue is fixed by #1188.

But it highlights a very similar issue #1194...

PR #1188, opened 2025-02-26
configury: Fixed disable-rpath

markbrown314:
Issue #1162

The build system relies on libtool to build libraries. libtool will insert an rpath into libraries unless you set hardcode_into_libs = no. Modified configury to set hardcode_into_libs if disable-rpath is set.


bcmIntc (approved):

davidozog (approved):
This pretty much does the right thing on my end (at least locally, Ubuntu 24.04)

It has the interesting effect that `pmi-simple` is not found by `ldd` unless explicitly setting `LD_LIBRARY_PATH` or `LD_PRELOAD`, but I think that's fine... `make check` still works in this case, so we could add a `--disable-rpath` row in CI if so desired.  (but maybe that's not very valuable).

Anyway, I'll keep an eye on it, but it certainly seems to fix #1162

Nice find on `hardcode_into_libs` @markbrown314!  And thanks for the fix!

Issue #1148, opened 2024-09-25
UCX + CMA shmem_ctx test fail: nonexistent PA


davidozog:
https://github.com/Sandia-OpenSHMEM/SOS/actions/runs/10924530684/job/30323799111?pr=1146

The occurrence seems intermittent and rare.

```
FAIL: shmem_ctx
===============

[0001] DEBUG: ../../src/init.c:376: shmem_internal_heap_postinit
[0001]        Thread level=MULTIPLE, Num. PEs=2
[0001]        Sym. heap=0x5555c0000000 len=537919488 -- data=0x555555558000 len=104
[0001] DEBUG: ../../src/init.c:457: shmem_internal_heap_postinit
[0001]        Affinity to 4 processor cores: { 0 1 2 3 }
Sandia OpenSHMEM 1.5.3rc1
  SHMEM_INFO                 1 (type: bool, default: 0)
	Print library information message at startup
  SHMEM_VERSION              0 (type: bool, default: 0)
	Print library version at startup
  SHMEM_DEBUG                1 (type: bool, default: 0)
	Enable debugging messages
  SHMEM_SYMMETRIC_SIZE       536870912 (type: size, default: 536870912)
	Symmetric heap size

Additional options:
make[6]: *** [Makefile:1180: test-suite.log] Error 1
make[5]: *** [Makefile:1288: check-TESTS] Error 2
make[4]: *** [Makefile:1495: check-am] Error 2
make[3]: *** [Makefile:469: check-recursive] Error 1
  SHMEM_SYMMETRIC_HEAP_USE_HUGE_PAGES 0 (type: bool, default: 0)
	Use Linux huge pages for symmetric heap
  SHMEM_SYMMETRIC_HEAP_PAGE_SIZE 2097152 (type: size, default: 2097152)
	Page size to use for huge pages
  SHMEM_SYMMETRIC_HEAP_USE_MALLOC 0 (type: bool, default: 0)
	Allocate the symmetric heap using malloc
  SHMEM_BOUNCE_SIZE          0 (type: size, default: 2048)
	Maximum message size to bounce buffer
  SHMEM_MAX_BOUNCE_BUFFERS   128 (type: long, default: 128)
	Maximum number of bounce buffers per context
  SHMEM_TRAP_ON_ABORT        0 (type: bool, default: 0)
	Generate trap if the program aborts or calls shmem_global_exit
  SHMEM_TEAMS_MAX            10 (type: long, default: 10)
[...]

Issue #1048, opened 2022-03-23
UCX w/ MPI deadlocking in CI


davidozog:
This row:
[UCX (--enable-pmi-mpi CC=mpicc --disable-fortran)](https://github.com/Sandia-OpenSHMEM/SOS/issues/new#logs)

Example here:
https://github.com/Sandia-OpenSHMEM/SOS/runs/5650454734?check_suite_focus=true

davidozog:
I haven't seen this failure in a long time - it may be a false positive.

davidozog:
This issue recurred at least twice in PR #1108, so reopening the issue for now.  In one case on the spec-example test `shmem_team_broadcast` deadlocked.

We may prefer to disable or swap out this test row until it's resolved because it can slow down PR procedures considerably.

davidozog:
Seeing another instance of probable deadlock on PR #1106, on unit test `cxx_test_shmem_put` I think.

markbrown314:
This might be a different manifestation of #1148 (hang rather than crash).

PR #1189, opened 2025-02-26
CI: Ignore failures during UCX testing

markbrown314:
There two issues with UCX in CI #1148 and #1048 that are under debug. Temporarily disable the test until these tickets are root caused and resolved inorder to reduce noise in CI results.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -555,6 +555,7 @@ jobs:
           make -j
           make install
       - name: Test SOS (${{ matrix.sos_config }})
+        continue-on-error: true
         run: |
           cd build
           make check TESTS= -j


markbrown314:
UCX testing in CI mostly fails which leads to noise in the CI results. I propose disabling the tests temporarily until the issue is root caused.

avincigu (commented):
I am a little bit against this. The UCX tests actually found bugs when I was developing the scan algorithms. We know the signature for the bugs, and though it is very annoying, we can usually just rerun the test so it passes.

markbrown314:
> I am a little bit against this. The UCX tests actually found bugs when I was developing the scan algorithms. We know the signature for the bugs, and though it is very annoying, we can usually just rerun the test so it passes.

Ok, that is fair. The failure occurs intermittently and always occurs in the UCX cma endpoint flush routine during shmem_transport_fini(). This indicates there is some form of memory corruption occurring, but it seems to only be reproducible in the GitHub CI test harness. 

How would you feel about this? I could define a configure flag to disable the flush on close in UCX transport (by disabling the UCP_EP_CLOSE_MODE_FLUSH flag from the call to ucp_ep_close_nb()). This disable flag would be enabled during the CI testing until we root cause the issue. This way the UCX testing still occurs but that specific bug is masked.

avincigu:
> > I am a little bit against this. The UCX tests actually found bugs when I was developing the scan algorithms. We know the signature for the bugs, and though it is very annoying, we can usually just rerun the test so it passes.
> 
> Ok, that is fair. The failure occurs intermittently and always occurs in the UCX cma endpoint flush routine during shmem_transport_fini(). This indicates there is some form of memory corruption occurring, but it seems to only be reproducible in the GitHub CI test harness.
> 
> How would you feel about this? I could define a configure flag to disable the flush on close in UCX transport (by disabling the UCP_EP_CLOSE_MODE_FLUSH flag from the call to ucp_ep_close_nb()). This disable flag would be enabled during the CI testing until we root cause the issue. This way the UCX testing still occurs but that specific bug is masked.

If it is easy enough to do, I would very much prefer this solution, however, only if you think it is a simple fix. If it involves a lot of work, forget about it and we can disable the whole thing.

abrooks98 (changes requested):
In my opinion, disabling an entire section in CI should be the last possible resort. Either adding a config to disable the failing portion or adding the ability to skip specific tests during `make check` should be explored first

bcmIntc:
Another option may be to add an exception framework to CI for 'known issues' - when detected re-test. That way we don't bury a bug.

markbrown314:
Github actions has job setting that will convert errors into warnings: [continue-on-error](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error).

This might be a good compromise. The test are run, but erroneous results for UCX's make check are converted into warnings and are annotated in the results output.

abrooks98 (approved):
Thanks for the compromise. I would suggest rewording the PR title, otherwise LGTM

markbrown314:
> Thanks for the compromise. I would suggest rewording the PR title, otherwise LGTM

Good catch!

bcmIntc (approved):

avincigu (approved):

Issue #1194, opened 2025-03-18
--disable-rpath has no effect on tests-sos submodule binaries


davidozog:
This is very similar to the problem in https://github.com/Sandia-OpenSHMEM/SOS/issues/1162, except that it affects the `tests-sos` submodule binaries.

In short, despite the fix in PR #1188, flags like ` -Wl,-rpath -Wl,/path/to/libfabric/lib -Wl,-rpath -Wl,/path/to/SOS/lib` are still passed to the link link of each individual test in `tests-sos`.

davidozog:
This problem does not arise on Ubuntu 24.04, but even there I still see the undesired `-Wl, -=rpath -Wl, ...` flags are passed to the compiler.  However, this issue does arise on RHEL 9 (with the default Autotool packages), which is a required platform at Cornelis Networks.

davidozog:
Closed by #1195

PR #1195, opened 2025-03-18
configury: -disable-rpath carried through to tests

davidozog:
This addresses issue #1194.

The solution is a little strange and derived from some poorly documented features of `libtool` that I learned about from stackoverflow - so I cite that source.

It essentially adds a dummy symbol to `hardcode_libdir_flag_spec`, which prevents flags like `-Wl,-rpath -Wl,/path/to/libfabric/lib -Wl,-rpath -Wl,/path/to/SOS/lib` from being added to the link line for the `tests-sos` binaries.


markbrown314:
Good catch! I did not look at the test-sos binaries. I only checked the libsma's DT_RUNPATH entry.

davidozog:
> Good catch! I did not look at the test-sos binaries. I only checked the libsma's DT_RUNPATH entry.

Indeed - I can't take credit, @bcernohous noticed the issue - thanks for catching it early Bob!

bcmIntc (approved):

PR #1213, opened 2025-07-21
Add fadd Routine for Internal Atomic Counter

markbrown314:


avincigu (approved):
Looks good, the only thing is that the workflow changes seem to be the same as in https://github.com/Sandia-OpenSHMEM/SOS/pull/1212/ I think you probably need to rebase

bcmIntc (approved):

Issue #1185, opened 2025-02-24
Symmetric Size Environment Variable Input Validation Lacking


markbrown314:
Support for multipliers in SHMEM_SYMMETRIC_SIZE environment variable lacks comprehensive input validation.

e.g. You can set SHMEM_SYMMETRIC_SIZE to "20Mabc123" and it truncates to 20M in the parsers, but it should really fail due to the remaining invalid characters.

markbrown314:
Resolved! Thank you.

PR #1216, opened 2025-08-19
Fixed SHMEM_SYMMETRIC_SIZE

kphuphanwoe:
i inspired by [OpenMPI](https://github.com/open-mpi/ompi/blob/6ec63e2b69d4ef5b0e085d38abfab35d24008318/oshmem/info/info.c#L240C10-L240C30)

#1185

### DIFF
--- a/src/shmem_env.c
+++ b/src/shmem_env.c
@@ -38,9 +38,12 @@ atol_scaled(char *str, shmem_internal_env_size *out)
     double p = -1.0;
     char f;
 
-    n = sscanf(str, "%lf%c", &p, &f);
+    if (2 == sscanf(str, "%lf%c%n", &p, &f, &n)) {
+        if (str[n] != '\0') {
+            fprintf(stderr, "Invalid size in environment variable\n");
+            return 1;
+        }
 
-    if (n == 2) {
         switch (f) {
             case 'k':
             case 'K':


kphuphanwoe:
Hi @markbrown314, apologies for the tag.
I opened this PR to fix issue. When you have some time, could you please review it? Thank you!

markbrown314:
Thanks! I will take a look at as soon as I can.

src/shmem_env.c
```diff
@@ -38,9 +38,9 @@ atol_scaled(char *str, shmem_internal_env_size *out)
     double p = -1.0;
     char f;
 
-    n = sscanf(str, "%lf%c", &p, &f);
+    if (2 == sscanf(str, "%lf%c%n", &p, &f, &n)) {
+        if (str[n] != '\0') { exit(1); }
```
markbrown314: You need to return 1 rather than exit(1)
When the `atol_scaled()` returns a non-zero value it prints "Invalid size in environment variable" to user. Exit will just shut it down abruptly.

markbrown314 (changes requested):
Thanks for the commit. Please clean up the commit message. Here is an example you can use:
```
Improved SHMEM_SYMMETRIC_SIZE Validation

Issue #1185 

If SHMEM_SYMMETRIC_SIZE contains invalid characters, return error.

Signed-off by: Your Real Name <email address>
```
It would be best if your signed off name be your real name rather than a pseudonym. As well as the Author Name (user.name git config).

kphuphanwoe:
The previous one was mistyped. I forgot to use `-s` in `git commit --amend`.

markbrown314 (approved):
This looks good. 🎉 Thank you for the PR.

Issue #1217, opened 2026-01-27
Improving OFI Transport Counter Polling Implementation


markbrown314:
The current libfabric counter polling implementation requires a series of separate lock acquisitions `fi_cntr()`, `fi_cntr_err()`, and `fi_cq_read()`, that can be simplified if we only use `fi_cnt_wait()`. We see improvements in put latency benchmarks when the `shmem_transport_ofi_put_quiet()` polling code is removed and no change in functionality.

Original:
```C
void shmem_transport_put_quiet(shmem_transport_ctx_t* ctx)
{
    ...
    while (poll_count < shmem_transport_ofi_put_poll_limit ||
           shmem_transport_ofi_put_poll_limit < 0) {
        success = fi_cntr_read(ctx->put_cntr);
        fail = fi_cntr_readerr(ctx->put_cntr);
        cnt = SHMEM_TRANSPORT_OFI_CNTR_READ(&ctx->pending_put_cntr);

        shmem_transport_probe();

        if (success < cnt && fail == 0) {
            SHMEM_TRANSPORT_OFI_CTX_UNLOCK(ctx);
            SPINLOCK_BODY();
            SHMEM_TRANSPORT_OFI_CTX_LOCK(ctx);
        } else if (fail) {
            RAISE_ERROR_MSG("Operations completed in error (%" PRIu64 ")\n", fail);
        } else {
            SHMEM_TRANSPORT_OFI_CTX_UNLOCK(ctx);
            return;
        }
        poll_count++;
    }
    cnt_new = SHMEM_TRANSPORT_OFI_CNTR_READ(&ctx->pending_put_cntr);
    do {
        cnt = cnt_new;
        ssize_t ret = fi_cntr_wait(ctx->put_cntr, cnt, -1);
        cnt_new = SHMEM_TRANSPORT_OFI_CNTR_READ(&ctx->pending_put_cntr);
        OFI_CTX_CHECK_ERROR(ctx, ret);
    } while (cnt < cnt_new);
    shmem_internal_assert(cnt == cnt_new);
    ...
```

Simplified:

```C
void shmem_transport_put_quiet(shmem_transport_ctx_t* ctx)
{
    ...
    uint64_t cnt, cnt_new;

    cnt_new = SHMEM_TRANSPORT_OFI_CNTR_READ(&ctx->pending_put_cntr);
    do {
        cnt = cnt_new;
[...]

PR #1218, opened 2026-01-27
Improving OFI Transport Counter Polling Implementation

markbrown314:
The current libfabric counter polling implementation requires a series of separate lock acquisitions fi_cntr(), fi_cntr_err(), and fi_cq_read(), that can be simplified if we only use fi_cnt_wait(). We see improvements in put latency benchmarks when the shmem_transport_ofi_put_quiet() polling code is removed and no change in functionality.

Addresses Issue #1217 
For evaluation purposes only.

### DIFF
--- a/src/transport_ofi.c
+++ b/src/transport_ofi.c
@@ -1731,6 +1731,7 @@ static int shmem_transport_ofi_ctx_init(shmem_transport_ctx_t *ctx, int id)
     cntr_put_attr.events   = FI_CNTR_EVENTS_COMP;
     cntr_get_attr.events   = FI_CNTR_EVENTS_COMP;
 
+#if 0
     /* Set FI_WAIT based on the put and get polling limits defined above */
     if (shmem_transport_ofi_put_poll_limit < 0) {
         cntr_put_attr.wait_obj = FI_WAIT_NONE;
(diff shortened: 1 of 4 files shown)


src/transport_ofi.c
```diff
@@ -1731,6 +1731,7 @@ static int shmem_transport_ofi_ctx_init(shmem_transport_ctx_t *ctx, int id)
     cntr_put_attr.events   = FI_CNTR_EVENTS_COMP;
     cntr_get_attr.events   = FI_CNTR_EVENTS_COMP;
 
+#if 0
```
bcmIntc, line 1734: Did you mean to leave this dead code?

src/transport_ofi.h
bcmIntc, line 501: You might want to update this text, as it seems no longer correct.

markbrown314: This part is incomplete. The original code had this complex counter wait mechanism, before it performs an OFI counter wait. I need to run some more analysis on why it was implemented this way in the first place (Chesterton's fence principle) before removing it completely.

src/transport_ofi.c
markbrown314: Yes, see comment below.

Issue #1207, opened 2025-06-24
Huge Page implementation uses incompatible mmap call


bcmIntc:
The function mmap_alloc() in symmetric_heap_c.c tries to find and open the hugetlbfs filesystem if huge page support is enabled via SHMEM_SYMMETRIC_HEAP_USE_HUGE_PAGES=1.
According to the man for mmap: "The mapping is not backed by any file; its contents are initialized to zero.  **The fd argument is ignored**; ..."
In practice I observe huge pages are not enabled unless I change the code to comply.

```
const char basename[] = "hugepagefile.SOS";
(...)
sprintf(file_name, "%s/%s.%d", directory, basename, getpid());
fd = open(file_name, O_CREAT | O_RDWR, 0755);
(...)

ret = mmap(requested_base, bytes, PROT_READ | PROT_WRITE, MAP_ANON | MAP_PRIVATE, **fd**, 0);
```

markbrown314:
That is a good catch.

PR #1227, opened 2026-04-15
symmetric_heap: fix huge pages mmap and silent fallback

bcmIntc:
Addresses two bugs in mmap_alloc() when SHMEM_SYMMETRIC_HEAP_USE_HUGE_PAGES=1:

1. (HIGH) MAP_ANON|MAP_PRIVATE was used even when a valid hugetlbfs fd was open. On Linux, MAP_ANONYMOUS causes the kernel to ignore the fd, so huge pages were never actually used. Fix: use MAP_SHARED when fd > 0 so the mapping is backed by the hugetlbfs file.

2. (LOW) When find_hugepage_dir() found no matching hugetlbfs mount, the code silently fell back to regular pages with no diagnostic. Fix: emit RAISE_WARN_MSG so the user knows their request was not honored.

Also moved the Linux-specific variables (file_name, fd, directory) and the mmap dispatch inside the '#ifdef __linux__' block so all huge page logic is cleanly scoped to Linux, with a plain anonymous mmap in the #else for all other platforms.

Addresses Issue #1207

### DIFF
--- a/src/symmetric_heap_c.c
+++ b/src/symmetric_heap_c.c
@@ -191,7 +191,8 @@ static void *mmap_alloc(size_t bytes)
                     sprintf(file_name, "%s/%s.%d", directory, basename, getpid());
                     fd = open(file_name, O_CREAT | O_RDWR, 0755);
                     if (fd < 0) {
-                        RAISE_WARN_STR("file open failed, cannot use huge pages");
+                        RAISE_WARN_MSG("file open failed (%s), cannot use huge pages",
+                                       strerror(errno));
                         fd = 0;
                     } else {
                         /* have to round up by the pagesize being used */
--- a/src/symmetric_heap_c.c
+++ b/src/symmetric_heap_c.c
@@ -164,15 +164,15 @@ shmem_internal_get_next(intptr_t incr)
 #define ONEGIG (1024UL*1024UL*1024UL)
 static void *mmap_alloc(size_t bytes)
 {
-    char *file_name = NULL;
-    int fd = 0;
-    char *directory = NULL;
     void *requested_base =
         (void*) (((unsigned long) shmem_internal_data_base +
                   shmem_internal_data_length + 2 * ONEGIG) & ~(ONEGIG - 1));
     void *ret;
 
 #ifdef __linux__
+    char *file_name = NULL;
+    int fd = 0;
+    char *directory = NULL;
     /* huge page support only on Linux for now, default is to use 2MB large pages */
     if (shmem_internal_params.SYMMETRIC_HEAP_USE_HUGE_PAGES) {
         const char basename[] = "hugepagefile.SOS";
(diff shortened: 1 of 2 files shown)


src/symmetric_heap_c.c
```diff
@@ -210,12 +210,24 @@ static void *mmap_alloc(size_t bytes)
...
          * would silently fall back to regular pages. */
-        ret = mmap(requested_base, bytes, PROT_READ | PROT_WRITE, MAP_SHARED, fd, 0);
-        if (file_name)
+        if (ftruncate(fd, bytes) == -1) {
+            RAISE_WARN_MSG("ftruncate on hugetlbfs file failed (%s), "
+                           "falling back to regular pages\n",
+                           strerror(errno));
+            unlink(file_name);
+            close(fd);
+            free(directory);
```
markbrown314, line 219: The directory and file_name can be leaked if fd is zero. This should be moved to a cleanup section.

There is also a bug in this code outside of this PR. If stdin is not open it is possible that you can get an fd of 0. fd should be initialized to a negative number.

markbrown314 (commented):
The cleanup code needs a little bit of work.

src/symmetric_heap_c.c
```diff
@@ -164,15 +164,15 @@ shmem_internal_get_next(intptr_t incr)
...
-    int fd = 0;
-    char *directory = NULL;
     void *requested_base =
         (void*) (((unsigned long) shmem_internal_data_base +
                   shmem_internal_data_length + 2 * ONEGIG) & ~(ONEGIG - 1));
     void *ret;
 
 #ifdef __linux__
+    char *file_name = NULL;
+    int fd = 0;
```
markbrown314, line 174: Don't initialize to zero. Initialize to -1. It is possible (although unlikely to get a fd of zero after open if stdin is not open).

src/symmetric_heap_c.c
```diff
@@ -199,34 +199,38 @@ static void *mmap_alloc(size_t bytes)
...
                 }
             }
+        } else {
+            RAISE_WARN_MSG("No hugetlbfs mount found for page size %zu, "
+                           "falling back to regular pages\n",
+                           shmem_internal_params.SYMMETRIC_HEAP_PAGE_SIZE);
         }
     }
+
+    if (fd) {
```
markbrown314, line 209: This should be a negative check. You would need to fix the code above in the fd < 0 error case that resets fd to 0 and just leave it negative.

src/symmetric_heap_c.c
```diff
@@ -199,34 +199,38 @@ static void *mmap_alloc(size_t bytes)
...
+        /* Map the hugetlbfs file directly; MAP_ANON must not be used here
+         * because MAP_ANONYMOUS causes the kernel to ignore the fd, which
+         * would silently fall back to regular pages. */
+        ret = mmap(requested_base, bytes, PROT_READ | PROT_WRITE, MAP_SHARED, fd, 0);
+        if (file_name)
+            unlink(file_name);
+        close(fd);
+        free(directory);
```
markbrown314, line 217: directory and file_name can leak.

src/symmetric_heap_c.c
```diff
@@ -199,34 +199,38 @@ static void *mmap_alloc(size_t bytes)
...
+        free(file_name);
+    } else {
+        ret = mmap(requested_base, bytes, PROT_READ | PROT_WRITE, MAP_ANON | MAP_PRIVATE, -1, 0);
+    }
+#else
+    ret = mmap(requested_base, bytes, PROT_READ | PROT_WRITE,
```
markbrown314, line 223: nitpick: there are a lot of mmaps in this code block. This could have been simplified by just using a flags variable. (set it to some defaults and fd set to -1 and if you want to use hugetlbfs set fd to reference hugepagefs file and set the flag to MAP_SHARED.

src/symmetric_heap_c.c
```diff
@@ -199,34 +199,38 @@ static void *mmap_alloc(size_t bytes)
...
-               fd,
-               0);
     if (ret == MAP_FAILED) {
         RAISE_WARN_MSG("Unable to allocate sym. heap, size %zuB: %s\n"
                        RAISE_PE_PREFIX
                        "Try reducing SHMEM_SYMMETRIC_SIZE or number of PEs per node\n",
                        bytes, strerror(errno), shmem_internal_my_pe);
         ret = NULL;
     }
-    if (fd) {
```
markbrown314, line 219: This part is fine other than the bogus fd != 0 check. It does not leak if you cleanup here.

markbrown314 (commented):
Some issues with cleanups and error checks.

Issue #1230, opened 2026-05-07
OFI transport: inject_size capped at 16 bytes due to fi_getinfo hint


bcmIntc:
**Description**

  `shmem_transport_ofi_max_buffered_send` was initialized to `sizeof(long double)`
  (16 bytes) before being passed as `tx_attr.inject_size` in the `fi_getinfo`
  hints. The CXI provider treats the hint value as a ceiling rather than a floor,
  returning exactly 16 — even though CXI's natural inject_size is 192 bytes.

  As a result, any put in the 17-192 byte range was sent via RDMA (with MR
  registration lookup) rather than the faster inline inject path.

  **Root cause**

  `src/transport_ofi.c`, function `query_for_fabric`:

  ```c
  shmem_transport_ofi_max_buffered_send = sizeof(long double);  // hint = 16
  ...
  tx_attr.inject_size = shmem_transport_ofi_max_buffered_send;
  ...
  ret = fi_getinfo(..., &hints, &info);
  ...
  shmem_transport_ofi_max_buffered_send = info->tx_attr->inject_size;  // adopts 16, not 192
  ```

  **Fix**

  Initialize shmem_transport_ofi_max_buffered_send = 0 (no minimum requirement)
  so the provider returns its natural inject_size, which is then adopted. This works for all providers except VERBS,
  which will provide a conservative injection size.

  See: https://github.com/Sandia-OpenSHMEM/SOS/pull/1231

  **Impact**

  Puts in the 17-192 byte range now use the inline inject path on CXI. Operations
  outside that range (AMOs ≤ 16 bytes, large puts > 192 bytes) are unaffected.

PR #1231, opened 2026-05-12
Issue #1230: transport/ofi: do not constrain inject_size hint to sizeof(long double)

bcmIntc:
## Summary

Addresses Issue #1230.

  - `query_for_fabric` previously initialized `shmem_transport_ofi_max_buffered_send = sizeof(long double)` before calling `fi_getinfo`, which passed that value as
  `tx_attr.inject_size` — telling the provider it *must* support at least 16-byte inject. This caused `fi_getinfo` to filter out providers that advertise a smaller
  natural inject size.
  - This change sets the hint to 0 (no minimum constraint) so the provider reports its natural `inject_size`. The returned value is immediately adopted after
  `fi_getinfo`, and the existing assert (`inject_size >= max_buffered_send`) trivially holds since both are 0 at that point.
  - On providers where the natural inject size is smaller than 16 bytes, the previous behavior could silently fail to select an otherwise valid endpoint, or cause
  `fi_getinfo` to return an inflated `inject_size` that doesn't reflect actual NIC capability.

  ## Test plan

  - [ ] Build and `make check` with CXI provider
  - [ ] Build and `make check` with verbs provider
  - [ ] Confirm `shmem_transport_ofi_max_buffered_send` value logged at startup reflects the provider's natural inject size
  - [ ] Verify no regression in small-message latency (inject path still engaged for messages <= provider's natural inject size)

### DIFF
--- a/src/transport_ofi.c
+++ b/src/transport_ofi.c
@@ -1470,7 +1470,9 @@ int query_for_fabric(struct fabric_info *info)
     struct fi_fabric_attr fabric_attr = {0};
     struct fi_ep_attr   ep_attr = {0};
 
-    shmem_transport_ofi_max_buffered_send = sizeof(long double);
+    /* Hint 0 = no minimum inject requirement; provider returns its natural
+     * inject_size, which is adopted below after fi_getinfo. */
+    shmem_transport_ofi_max_buffered_send = 0;
 
     fabric_attr.prov_name = info->prov_name;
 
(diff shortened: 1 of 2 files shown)


src/transport_ofi.c
```diff
@@ -1470,7 +1470,9 @@ int query_for_fabric(struct fabric_info *info)
     struct fi_fabric_attr fabric_attr = {0};
     struct fi_ep_attr   ep_attr = {0};
 
-    shmem_transport_ofi_max_buffered_send = sizeof(long double);
+    /* Hint 0 = no minimum inject requirement; provider returns its natural
+     * inject_size, which is adopted below after fi_getinfo. */
+    shmem_transport_ofi_max_buffered_send = 0;
```
markbrown314, line 1475: Looks good!

PR #1232, opened 2026-06-04
Add 2MB Huge Page Support for Symmetric Heap + CXI Fence Optimization

bcmIntc:
This PR includes three changes targeting CXI/Slingshot performance on systems like Perlmutter.

  **Changes**

  1. Skip put_quiet in Fence for CXI Provider (7880ab43)

  The CXI provider doesn't require explicit quiet operations before fences since fences are already ordering operations. Skipping the redundant put_quiet() reduces fence overhead.

  2. Anonymous MAP_HUGETLB for Symmetric Heap (0e985e7e)

  Enables allocation of the symmetric heap using 2MB huge pages via anonymous MAP_HUGETLB, which works with kernel nr_overcommit_hugepages without requiring mounted hugetlbfs. Falls back to transparent huge pages (THP) via madvise() if explicit huge pages are unavailable.

  Usage:
  export SHMEM_SYMMETRIC_HEAP_USE_HUGE_PAGES=1

  3. Explicitly Request 2MB Pages with MAP_HUGE_SHIFT (99b900ac)

  Adds (21 << MAP_HUGE_SHIFT) to the MAP_HUGETLB mmap call to explicitly request 2MB pages (2^21 = 2MB). This matches the allocation strategy used by the CXI libfabric provider for its internal buffers.

  **Performance Impact**

  When combined with scalable MR mode (--enable-ofi-mr=scalable), the huge page changes enable the CXI NIC's ATU (Address Translation Unit) to use 2MB page granularity instead of 4KB, dramatically reducing translation overhead:

  - ATU cache hit rate: 74-84% → 98.3%
  - ATU cache misses: ~10-15M per node → <300 per node
  - 2MB page translations: 0 → ~450K-490K per node

  Additional benefits:
  - Reduced CPU TLB misses for symmetric heap accesses
  - Eliminated credit stall cycles (260M → 0)
  - Improved latency distribution (96% in fastest bucket)

  **Configuration Requirements**

  The ATU benefits require scalable MR mode (recommended for CXI/Slingshot):
  ./configure --enable-ofi-mr=scalable --enable-mr-endpoint ...

  With scalable MR, fi_mr_reg() registers the entire address space, allowing the CXI provider to detect 2MB backing pages and configure the ATU accordingly. Also eliminates ASLR sensitivity (no setarch --addr-no-randomize needed).

  **Testing**

  Verified on NERSC Perlmutter with 2-node and 256-PE jobs using CXI telemetry counters to confirm ATU derivative (2MB) page activity.

  ---

### DIFF
--- a/src/transport_ofi.c
+++ b/src/transport_ofi.c
@@ -105,7 +105,7 @@ long                            shmem_transport_ofi_amo_pipeline_depth;
 size_t                          shmem_transport_ofi_max_buffered_send;
 size_t                          shmem_transport_ofi_max_msg_size;
 size_t                          shmem_transport_ofi_bounce_buffer_size;
-long                            shmem_transport_ofi_max_bounce_buffers;
+size_t                          shmem_transport_ofi_max_bounce_buffers;
 size_t                          shmem_transport_ofi_addrlen;
 #ifdef ENABLE_MR_RMA_EVENT
 int                             shmem_transport_ofi_mr_rma_event;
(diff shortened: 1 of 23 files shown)


src/transport_ofi.c
```diff
@@ -1723,6 +1726,12 @@ int query_for_fabric(struct fabric_info *info)
               shmem_transport_ofi_stx_max,
               num_nics);
 
+    /* Set CXI provider flag for fence optimization */
```
markbrown314: This is useful, but we should generalize this into a provider check function:
bool shmem_transport_check_provider(const char *name)

If name matches, provider it return true

PR #1237, opened 2026-06-25
Performance: hierarchical barrier, tiered huge pages, and OFI/CXI transport improvements

bcmIntc:
Adds a NIC-offloading hierarchical barrier and reintroduces two CXI/heap optimizations in isolation, each independently gated.

 # Hierarchical barrier (XPMEM + NIC dissemination) #

  - New --enable-hierarchical-barrier (98af1b76): three-phase barrier — intranode gather/fanout via CPU atomics over XPMEM (cache-line-padded per-PE slots), NIC puts restricted to the internode dissemination phase (node roots only). Keeps intranode traffic off the NIC.
  - Correctness + scratch fixes (dbe498c7, 71521f4c): per-team sense counters eliminate the overlapping-team signal-mismatch deadlock; persistent init-time scratch replaces the alloca/malloc-per-call; node-id-based root sets; dead-code cleanup.
  - Per-team topology caching (33cc4bec): active-set/root-set/tree topology is invariant per team but was rebuilt every call (incl. an O(PE_size²) scan). Cache it per team, build once; root-set build made O(PE_size). Allocation-failure-safe fallback to scratch.
  - AUTO-selection gating (d5b40ee4): only select hierarchical when the job spans multiple nodes and on a job-global min PPN, so every PE makes the same choice on heterogeneous-PPN jobs. Explicit BARRIER_ALGORITHM=hierarchical still honored.
  - --enable-hierarchical-barrier now requires XPMEM specifically (intranode phase needs shmem_shr_transport_ptr).

 # CXI hybrid MR descriptor (9eab54d4, 3ba9445e, 2cc39b22) #

  Enables FI_CXI_DOM_OPS_3 enable_hybrid_mr_desc after fi_domain() (as Cray SHMEM does) so the provider trusts a non-NULL desc and skips its internal MR-registration walk on large puts. Gated by SHMEM_OFI_CXI_HYBRID_MR_DESC and an exact-strcmp provider check. Compiled in only when a configure probe finds the real struct fi_cxi_dom_ops member (no hand-rolled struct mirror → no ABI-drift risk); the probe supplies a full prerequisite-include set so it's robust across  libfabric versions/clusters, with a configure-summary line surfacing the result.

 # Symmetric heap huge pages (52858899) #

  Reintroduces tiered huge-page allocation gated on SHMEM_SYMMETRIC_HEAP_USE_HUGE_PAGES (default off): hugetlbfs file → anonymous MAP_HUGETLB → regular pages. No THP/madvise tier. Fallback path is byte-for-byte the prior fixed-address mapping, preserving VA symmetry across PEs.

 # Misc #1236 

  - db8cf246: zero max_buffered_send before the fi_getinfo hint so the provider advertises its natural inject_size.

 # Notes: # 
CXI hybrid MR and huge pages are both off/neutral by default. All optimizations are individually gated for safe A/B testing.


bcmIntc:
# Test config: #
--with-ofi=/opt/cray/libfabric/1.22.0/ --with-xpmem=/usr/lib --enable-pmi-mpi --disable-fortran --enable-ofi-mr=scalable --enable-ofi-manual-progress --disable-libtool-wrapper --enable-mr-endpoint --enable-manual-progress --enable-hierarchical-barrier CC=mpicc CXX=mpicxx

# Envt: #
export SHMEM_SYMMETRIC_HEAP_USE_HUGE_PAGES=1
export SHMEM_SYMMETRIC_SIZE=3G
export SHMEM_BARRIER_ALGORITHM=auto
export SHMEM_OFI_PROVIDER=cxi
export FI_PROVIDER=cxi
export FI_MR_CACHE_MAX_COUNT=32768
export FI_MR_CACHE_MAX_SIZE=-1
export SHMEM_BOUNCE_SIZE=8192

